PR #21, opened 2016-07-11
Add Synchronous to MySqlConnectionStringBuilder.

ejball:
This initial implementation could have adverse affects on performance, so you might not want to merge, but it is an idea for enforcing synchronous access, even when using the async methods, to improve debugging.

### DIFF
--- a/src/MySqlConnector/MySqlClient/MySqlConnection.cs
+++ b/src/MySqlConnector/MySqlClient/MySqlConnection.cs
@@ -261,7 +261,7 @@ namespace MySql.Data.MySqlClient
 
 		internal ValueTask<T> AdaptTask<T>(ValueTask<T> task)
 		{
-			if (!Synchronous)
+			if (!Synchronous || task.IsCompletedSuccessfully)
 				return task;
 
 			return new ValueTask<T>(task.AsTask().GetAwaiter().GetResult());
--- a/src/MySqlConnector/MySqlClient/MySqlConnection.cs
+++ b/src/MySqlConnector/MySqlClient/MySqlConnection.cs
@@ -114,14 +114,14 @@ namespace MySql.Data.MySqlClient
 				if (m_session == null)
 				{
 					m_session = new MySqlSession(pool);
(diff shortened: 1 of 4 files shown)


src/MySqlConnector/MySqlClient/MySqlConnection.cs
```diff
@@ -240,6 +240,32 @@ internal MySqlSession Session
...
+
+			return Task.FromResult(task.GetAwaiter().GetResult());
+		}
+
+		internal ValueTask<T> AdaptTask<T>(ValueTask<T> task)
+		{
+			if (!Synchronous)
+				return task;
+
+			return new ValueTask<T>(task.AsTask().GetAwaiter().GetResult());
```
bgrainger: Should this have an optimised code path for the case where the `ValueTask` already has a value (which is expected to be common, hence the use of `ValueTask`)?

bgrainger: Maybe `.AsTask()` returns a completed task efficiently; but still there is unnecessary allocation.

bgrainger:
I'm not convinced I like this approach. I haven't attempted it, but it feels like one could make the socket operations synchronous, then use `ValueTask` to bubble the results as far up the call stack as possible. (Then, the remaining uses of `Task` would still be synchronous.)

This ties into an improvement I want to make to use all new non-public methods that return `ValueTask` in ExecuteDbDataReaderAsync and reimplement the public API in terms of those new ValueTask methods.

src/MySqlConnector/MySqlClient/MySqlConnection.cs
ejball: Yes, that would be a great change.

ejball:
I don't like it either, but I wasn't ready to do the hard work of using synchronous socket operations. This is just a proof of concept that may encourage someone to eventually do it right. :-)

ejball:
Rebased.

bgrainger:
> wasn't ready to do the hard work of using synchronous socket operations

Turned out to not be that hard; code is referenced from #62. I'm going to close this PR in favour of (the eventual resolution of) that one.

PR #25, opened 2016-07-11
Fix GetName failure after reading past end of results. Fixes #24

### LINKED ISSUE #24: InvalidOperationException("There is no current result set.") - reader.GetName(0)
Calling reader.GetName(0) throws "InvalidOperationException". The source of the exception is in the function VerifyHasResult. The state of the reader is "State.AlreadyReadFirstRow".
 
```
		private void VerifyHasResult()
		{
			VerifyNotDisposed();
			if (m_state != State.ReadResultSetHeader && m_state != State.ReadingRows && m_state != )
				throw new InvalidOperationException("There is no current result set.");
		}
```

bgrainger:
Change exception type to match official connector.

### DIFF
--- a/src/MySqlConnector/MySqlClient/MySqlDataReader.cs
+++ b/src/MySqlConnector/MySqlClient/MySqlDataReader.cs
@@ -24,12 +24,14 @@ namespace MySql.Data.MySqlClient
 			while (m_state == State.ReadingRows || m_state == State.ReadResultSetHeader)
 				await ReadAsync(cancellationToken).ConfigureAwait(false);
 
-			if (m_state == State.NoMoreData)
-				return false;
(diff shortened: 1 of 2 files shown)


ejball:
Consider adding a C# comment to explain why you're violating the framework guidelines by throwing that exception.

bgrainger:
It's what the documentation specifies: https://msdn.microsoft.com/en-us/library/system.data.idatarecord.getname.aspx

ejball:
How about that! Never mind me.

bgrainger:
It's certainly not what I expected (see my original code). Also, the documentation doesn't specify what exception to throw after `NextResultAsync` returns `false`; I match the official connector but `IndexOutOfRangeException` definitely feels wrong.

PR #50, opened 2016-09-09
json type; bool casting; async reader waiting

caleblloyd:
We're getting closer!  Here's the 3 things that I have added:

**JSON Support**

Needed a way to cast byte[] from GetValue into a JsonObject[T] in Pomelo's library.  Since JsonObject[T] may be passed as JsonObject[string], JsonObject[Dictionary[string,string]], or other types, a simple cast did not work.  Solution was to use reflection to cast to these types.  Now any object that implements an implicit method for byte[] can deserialize JSON.  Here's Pomelo's proposed implementation:

    public static implicit operator JsonObject<T>(byte[] serialized)
    {
        return new JsonObject<T>
        {
            Json = Text.Encoding.UTF8.GetString(serialized)
        };
    }

**Boolean Casting**

Added logic to cast integer types stored in the DBMS to booleans if requested.

**Asynchronous Wait for Reader to Complete**

MySQL only supports one active reader at a time.  If multiple asynchronous queries are sent at once in the same Database Context, this can cause multiple statements to execute at once that try to open readers.

I needed a way to asynchronously wait on a reader to complete so that I could serialize these operations.  The solution I am proposing uses a semaphore that is locked until the reader is complete, then unlocked.  I'm open for suggestions on other ways to achieve this, but this was the only way I could come up with that didn't involve polling.


bgrainger:
> MySQL only supports one active reader at a time. If multiple asynchronous queries are sent at once in the same Database Context, this can cause multiple statements to execute at once that try to open readers.

I would consider this a bug in the calling code. I don't think it's the responsibility of this library to serialize operations. (There are many "wrong" things you could do with overlapping commands on the same connection; I think catching just one of them is the wrong approach.) But I need to investigate this more to make a final decision.

caleblloyd:
@bgrainger I agree that the ADO.NET library should not serialize operations, and the change made in this PR does not serialize operations.

It simply provides an asynchronous method that says "Hey, this reader is finished" so that higher level code can perform serialization if needed.

All of the actual serialization is handled in Pomelo.

bgrainger:
> Needed a way to cast byte[] from GetValue into a JsonObject[T] in Pomelo's library.

Can Pomelo do `(JsonObject<T>) reader.GetBytes(0)`? I'm not understanding why the reflection needs to be in this library (and why `GetValue<T>` needs to work for any arbitrary `T`).

caleblloyd:
My goal was to make this an extensible function for deserializing JSON across all libraries without introducing a JSON library as a dependency.

In the case of Entity Framework, the internal Entity Framework code is calling:

    GetFieldValue(model.Field)

This is called for every single field in a model.  This causes an arbitrary T to be passed to GetFieldValue regardless of the type because objects in models are defind like such:

    public JsonObject<Dictionary<String,String>> Field {get; set;}
    public JsonObject<String> Field2 {get; set;}
    public JsonObject<Dictionary<String,Decimal>> Field3 {get; set;}
    ...

The change in this PR attempts to use normal typecasting first, and catches a InvalidCastException.  Reflection is only used as a last resort, if the InvalidCastException is thrown so the performance impact of reflection will not happen unless an exception was going to be thrown anyways.

Other libraries that implement MySqlConnector and need JSON support should be able to use this as well.

bgrainger:
RE: Asynchronous Wait

Is the issue that `MySqlRelationalCommand.ExecuteAsync` can return a `MySqlDataReader`, but that the caller of that method can call `ExecuteAsync` again before they've finished reading (and disposed) the reader? So we want `ExecuteAsync` to block in that scenario?

caleblloyd:
> Is the issue that MySqlRelationalCommand.ExecuteAsync can return a MySqlDataReader, but that the caller of that method can call ExecuteAsync again before they've finished reading (and disposed) the reader? So we want ExecuteAsync to block in that scenario?

@bgrainger that is correct

bgrainger:
I need to read through the code (might not have time today) but my initial feeling is that async wait and JSON would be better handled through a custom `DbDataReader` implementation in Pomelo.EF that wraps `MySqlDataReader` and adds the EF-specific concerns.

(I've written that type of code before, so I can fork Pomelo.EF and test if this approach would work.)

caleblloyd:
@bgrainger I explored that as an option.

It would work just fine for the JSON conversion, although it's my opinion that all libraries implementing JSON support could benefit from casting to arbitrarily typed JSON objects given that they provide the proper `public static implicit operator JsonObject<T>(byte[] serialized)` implementation.  I think the implementation in this PR would sufficiently resolve JSON type support for #4 

I couldn't figure out a way to get a wrapping class to implement an Async Wait for the `MySqlDbDataReader` though.  Presently, all of the Reader State code is held in a private variable (`m_state`), and there is no way for the wrapping class to receive updates about the Reader State Change.  The proposed method implements an asynchronous method that always returns immediately if the reader is closed.  An alternative would be a synchronous public method to get the reader state, and an event in the reader that is fired on reader state change.

Thanks for your help thus far and enjoy the weekend!

Kagamine:
The conversion doesn't depend on ADO layer, it just read the json field as a string, then call JSON.NET to make conversions.

bgrainger:
General comments:

* Please put each logical change in its own commit (instead of fixing three things in one commit).
* Ideally, open a separate PR for each logical change, so each can be discussed separately.
* Please add tests to a) verify that the new code mirrors or is compatible with MySql.Data, b) verify that no new bugs are introduced, c) create regression tests that will catch future bugs.

Specific comments on individual changes will be forthcoming.

.gitignore
```diff
@@ -16,3 +16,4 @@ Thumbs.db
 *.suo
 .vs/
 *.xproj.user
+.idea
```
bgrainger: What is this related to?

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -97,7 +103,20 @@ private bool ReadAsyncRemainder(PayloadData payload)
...
+		        return (ushort) value != 0;
+		    if (value is int)
+		        return (int) value != 0;
+		    if (value is uint)
+		        return (uint) value != 0;
+		    if (value is long)
+		        return (long) value != 0;
+		    if (value is ulong)
+		        return (ulong) value != 0;
+		    return (bool) value;
```
bgrainger: I'd prefer to put an `is bool` test first, as I would expect that to be the most common  scenario.

Missing checks for `sbyte` and `byte`. Missing tests for the new behaviour.

I've pushed a commit with just this change (and new tests): 46e6bb58d565b2a392a8e7f610b6a736a457d3fa

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -398,6 +417,9 @@ public override string GetDataTypeName(int ordinal)
 			case ColumnType.NewDecimal:
 				return "DECIMAL";
 
+			case ColumnType.Json:
+			    return "JSON";
```
bgrainger: I want to support this (very soon) but haven't developed the full test suite yet.

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -555,7 +579,32 @@ public override object GetValue(int ordinal)
...
+	        }
+	        catch (InvalidCastException e)
+	        {
+	            try
+	            {
+	                // try casting using reflection; needed for json
+	                var dataParam = Expression.Parameter(typeof(byte[]), "data");
+	                var body = Expression.Block(Expression.Convert(dataParam, typeof(T)));
+	                var run = Expression.Lambda(body, dataParam).Compile();
+	                return (T)run.DynamicInvoke(GetValue(ordinal));
```
bgrainger: I'd prefer to move this up to Pomelo.EntityFrameworkCore.MySql, which has the need to create `JsonObject<T>` objects. 

See proposed implementation at https://github.com/bgrainger/Pomelo.EntityFrameworkCore.MySql/commit/0e2a88a0612f439a9414d6b997b795ec77bc5a09

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -781,12 +830,32 @@ private enum State
...
+	        m_done_reading.Release();
+	    }
+
 		static readonly Task<bool> s_falseTask = Task.FromResult(false);
 		static readonly Task<bool> s_trueTask = Task.FromResult(true);
 
 		MySqlCommand m_command;
 		MySqlSession m_session;
-		State m_state;
+	    SemaphoreSlim m_done_reading = new SemaphoreSlim(1);
```
bgrainger: Although `SemaphoreSlim` is lightweight, it seems wasteful to instantiate one of these for every single `MySqlDataReader` that's created (when most clients will not use it).

Prefer to initialise this in the constructor with every other field.

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -781,12 +830,32 @@ private enum State
...
+	    State _m_state;
+	    State m_state
```
bgrainger: Shouldn't have two fields with almost identical names. This could become a private property `State`, or a `private SetState(State state)` method could be added to encapsulate the setting logic.

src/MySqlConnector/project.json
```diff
@@ -37,6 +37,7 @@
 				"System.Globalization": "4.0.11-*",
 				"System.IO": "4.1.0-*",
 				"System.Linq": "4.1.0-*",
+				"System.Linq.Expressions": "4.1.0-*",
```
bgrainger: I'd prefer not to take an additional (and typically unnecessary) dependency in this library. As above, I think this should move up to Pomelo.EF.

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -781,12 +830,32 @@ private enum State
...
+	    {
+	        get { return _m_state; }
+	        set
+	        {
+	            if (value == State.None || value == State.NoMoreData)
+	            {
+	                m_done_reading.Release();
```
bgrainger: The logic is incorrect here. `NextResultAsync` calls `Reset` which sets `m_state = State.None`. In the case of multiple result sets, `NextResultAsync` might get called multiple times, which will repeatedly `Release` the semaphore, increasing its count. This will let a thread in early, before the reader has been fully read.

bgrainger: Overall, I think it will be more reliable to move all this logic to Pomelo.EF, e.g., as in the proposed implementation here: https://github.com/bgrainger/Pomelo.EntityFrameworkCore.MySql/commit/7354ecf6dbf211721104943f878c07ad53c52581

.gitignore
caleblloyd: This is Jetbrains Project Rider. It's a standalone C# editor with ReSharper. I develop on Linux so no visual studio for me

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -517,6 +540,7 @@ public override object GetValue(int ordinal)
 			case ColumnType.Blob:
 			case ColumnType.MediumBlob:
 			case ColumnType.LongBlob:
+			case ColumnType.Json:
```
caleblloyd: Without this, GetValue on a JSON field throws an Exception

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
caleblloyd: @bgrainger what logic would be correct for a close notifier?  Does MySql ever send a packet that says "all result sets have been read, no more data on the wire"?

bgrainger: Yes, it will eventually send an EOF packet.

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
bgrainger: If this semaphore starts with an initial count of 1, then the first call to `WaitAsync` will enter the semaphore immediately; thus this won't actually enforce mutual exclusion.

I'm also not sure that this would correctly handle the case of multiple threads calling `ExecuteAsync` and all waiting on the semaphore.

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
bgrainger: I think that this library could raise an event when that happens. That's zero overhead for clients who don't need it, but Pomelo.EF could subscribe to the event and release its semaphore when that happens. Then you wouldn't need clients to dispose the reader, but simply keep reading until `Read(Async)` returns false.

This still seems like a programming error that shouldn't be automatically masked by Pomelo.EF, but I'm not familiar enough with EF to know if that's a standard pattern or not, and if not implementing it would inconvenience clients.

bgrainger:
Split JSON data type out to #51.

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
bgrainger: See #51.

bgrainger:
Addressed the four parts of this PR as follows:
* fixed `GetBoolean` casting: 46e6bb58d565b2a392a8e7f610b6a736a457d3fa
* added `JSON` support: aa53c11bbb9040dca0090d77bcb1ce9c1cc8d16f
* moved casting with reflection to https://github.com/bgrainger/Pomelo.EntityFrameworkCore.MySql/commit/0e2a88a0612f439a9414d6b997b795ec77bc5a09
* moved data reader locking to https://github.com/bgrainger/Pomelo.EntityFrameworkCore.MySql/commit/7354ecf6dbf211721104943f878c07ad53c52581

.gitignore
bgrainger: Added in 1d4238f5f040ab9a065134e2317f81152c66d8fd.

Issue #38, opened 2016-09-01
Implement connection pooling.


OzieGamma:
Hi,

I tried your connector out. It worked in a small demo App ("select name from users").

When I launched a load test (I wanted to test .NET Core performance on Swisscom cloud), It ended up crashing quite often with "Not more connections available".

Maybe I am doing something wrong, here is my code:
```csharp
    public interface IDb
    {
        Task<IList<string>> GetUserNames();
    }

    public class Db : IDb
    {
        public string ConnectionString { get; private set; }

        public Db(string connectionString)
        {
            this.ConnectionString = connectionString;
        }

        public async Task<IList<string>> GetUserNames()
        {
            using (IDbConnection connection = new MySqlConnection(ConnectionString))
            {
                var results = await connection.QueryAsync<string>("select Name from Users");
                return results.AsList();
            }
        }
    }
```

bgrainger:
Although I have used connection pooling with this library, it hasn't been stress-tested yet, so there may be a bug.

What is the full connection string (without User ID and Password)? Approximately how many connections are you opening simultaneously?

OzieGamma:
```csharp
            dynamic vcap_services = JsonConvert.DeserializeObject(Configuration["VCAP_SERVICES"]);

            var server = vcap_services.mariadb[0].credentials.host;
            var port = vcap_services.mariadb[0].credentials.port;
            var database = vcap_services.mariadb[0].credentials.database;
            var username = vcap_services.mariadb[0].credentials.username;
            var password = vcap_services.mariadb[0].credentials.password;
            var connectionString =
                $"Server = {server}; Port = {port}; Database = {database}; Uid = {username};Pwd = {password};";
```

I guess it creates one per Http Request?

Isn't the library supposed to handle that? 

Or do I need to specify the max number of connections in the connection string?

bgrainger:
The library should handle that just fine. Can you paste in the complete exception message and full call stack you're getting when it fails?

OzieGamma:
Sure, I'll do that tomorrow!

caleblloyd:
There is definitely an issue with the connection pool, I'm hitting in in performance tests from #55 

I'm working up a separate PR with a fix right now

OzieGamma:
Sorry for never answering this. I moved to use Postgres (use a different provider).

bgrainger:
A connection pooling PR has been merged; closing this issue due to lack of steps to repro.

PR #56, opened 2016-09-15
Fix Connection Pool

caleblloyd:
I've taken a stab at fixing the connection pool.  I tested this implementation with MySqlConnector.Performance, but it seems to hang the SideBySide.New tests.  SideBySide.New does not print which test is currently running, so I'm not sure how to debug.

@bgrainger can you take a look at the SideBySide.New tests and see if you can tell why they are hanging?  If you can add logging to it to print the method it's in, I can resolve the bug.  I tried briefly but I'm not very familiar with XUnit.

Async targets are running stable at 1500 Requests Per Second in the Performance Tests.  Very nice!!!  I'm running on an Intel Core i5-6260U, load test, web app, and database all running on the same machine.  Should do better on a dedicated app server.

	Rate = 1500 RPS
	Duration = 10s
	Targets = targets-async.txt

	Requests      [total, rate]            15000, 1500.10
	Duration      [total, attack, wait]    10.000748548s, 9.999322811s, 1.425737ms
	Latencies     [mean, 50, 95, 99, max]  23.299198ms, 2.395148ms, 168.782969ms, 271.810346ms, 395.213902ms
	Bytes In      [total, mean]            173108268, 11540.55
	Bytes Out     [total, mean]            5561250, 370.75
	Success       [ratio]                  100.00%
	Status Codes  [code:count]             200:15000

### DIFF
--- a/src/MySqlConnector/MySqlClient/ConnectionPool.cs
+++ b/src/MySqlConnector/MySqlClient/ConnectionPool.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using MySql.Data.Serialization;
@@ -7,64 +8,211 @@ namespace MySql.Data.MySqlClient
 {
 	internal sealed class ConnectionPool
 	{
-		public async Task<MySqlSession> TryGetSessionAsync(CancellationToken cancellationToken)
+		public async Task<MySqlSession> GetSessionAsync(CancellationToken cancellationToken)
 		{
-			try
-			{
-				await m_semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+			AssertInvariants();
 
-				if (m_sessions.Count > 0)
-					return m_sessions.Dequeue();
-			}
-			finally
+			using (cancellationToken.Register(() =>
 			{
-				m_semaphore.Release();
+				// wake up all waiting threads if the cancellation token is cancelled
+				lock (m_lock)
+					Monitor.PulseAll(m_lock);
+			}))
+			{
+				MySqlSession temporarySession = null;
+
+				try
+				{
+					// keep looping until cancelled (timeout) or a session is retrieved
+					while (true)
+					{
+						cancellationToken.ThrowIfCancellationRequested();
+
+						bool isNew;
+						AssertInvariants();
+
+						// grab an idle session, or take an available session, or block
+						lock (m_lock)
+						{
+							if (m_idleSessions.Count > 0)
+							{
+								temporarySession = m_idleSessions.Dequeue();
+								isNew = false;
+								m_temporary++;
+							}
+							else if (m_available > 0)
+							{
+								temporarySession = new MySqlSession(this);
+								isNew = true;
+								m_temporary++;
+								m_available--;
+							}
+							else
+							{
+								Monitor.Wait(m_lock);
+								continue;
+							}
+						}
+						AssertInvariants();
+
+						// asynchronously connect the session or verify that an idle session is valid
+						bool isValid;
+						if (isNew)
+						{
+							await temporarySession.ConnectAsync(m_server.Split(','), m_port, m_userId, m_password, m_database, cancellationToken).ConfigureAwait(false);
+							isValid = true;
+						}
+						else
+						{
+							isValid = await TryActivateIdleSessionAsync(temporarySession, cancellationToken).ConfigureAwait(false);
+						}
+
+						// if valid, return it
+						if (isValid)
+						{
+							lock (m_lock)
+							{
+								m_activeSessions.Add(temporarySession);
+								m_temporary--;
+
+								var session = temporarySession;
+								temporarySession = null;
+								return session;
+							}
+						}
+
+						// destroy this invalid temporary session and try again
+						lock (m_lock)
+						{
+							m_temporary--;
+							m_available++;
+							temporarySession = null;
+
+							Monitor.Pulse(m_lock);
+						}
+					}
+				}
+				catch
+				{
+					// clean up any local temporary session upon exception
+					lock (m_lock)
+					{
+						if (temporarySession != null)
+						{
+							m_temporary--;
+							m_available++;
+							Monitor.Pulse(m_lock);
+						}
+					}
+					throw;
+				}
+				finally
+				{
+					AssertInvariants();
+				}
 			}
-			return null;
 		}
 
-		public bool Return(MySqlSession session)
+		private async Task<bool> TryActivateIdleSessionAsync(MySqlSession session, CancellationToken cancellationToken)
 		{
-			try
+			// test that session is still valid and (optionally) reset it
+			if (!await session.TryPingAsync(cancellationToken).ConfigureAwait(false))
 			{
-				m_semaphore.Wait();
+				await session.DisposeAsync(cancellationToken).ConfigureAwait(false);
+				return false;
+			}
 
-				// TODO: Dispose oldest connections in the pool first?
-				if (m_sessions.Count >= m_maximumSize)
-					return false;
-				m_sessions.Enqueue(session);
-				return true;
-			}
-			finally
+			if (m_resetConnections)
+				await session.ResetConnectionAsync(m_userId, m_password, m_database, cancellationToken).ConfigureAwait(false);
+			return true;
+		}
+		
+		public void Return(MySqlSession session)
+		{
+			AssertInvariants();
+
+			lock (m_lock)
 			{
-				m_semaphore.Release();
+				if (!m_activeSessions.Remove(session))
+					throw new InvalidOperationException("Returned session wasn't active.");
+				m_temporary++;
 			}
+
+			AssertInvariants();
+			
+			if (m_closeOnReturn)
+			{
+				session.DisposeAsync(CancellationToken.None).GetAwaiter().GetResult();
+				lock (m_lock)
+				{
+					m_temporary--;
+					m_available++;
+					Monitor.Pulse(m_lock);
+				}
+			}
+			else
+			{
+				lock (m_lock)
+				{
+					m_temporary--;
+					m_idleSessions.Enqueue(session);
+					Monitor.Pulse(m_lock);
+				}
+			}
+			
+			AssertInvariants();
 		}
 
 		public async Task ClearAsync(CancellationToken cancellationToken)
 		{
-			try
+			AssertInvariants();
+
+			while (!cancellationToken.IsCancellationRequested)
 			{
-				await m_semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+				MySqlSession session = null;
 
-				while (m_sessions.Count > 0)
+				// get one idle session at a time and clean it up
+				AssertInvariants();
+				lock (m_lock)
 				{
-					cancellationToken.ThrowIfCancellationRequested();
+					if (m_idleSessions.Count > 0)
+					{
+						session = m_idleSessions.Dequeue();
+						m_temporary++;
+					}
+				}
+				AssertInvariants();
 
-					var session = m_sessions.Dequeue();
+				if (session == null)
+					break;
+
+				try
+				{
 					await session.DisposeAsync(cancellationToken).ConfigureAwait(false);
 				}
+				catch (OperationCanceledException)
+				{
+				}
+				finally
+				{
+					AssertInvariants();
+					lock (m_lock)
+					{
+						m_temporary--;
+						m_available++;
+						Monitor.Pulse(m_lock);
+					}
+					AssertInvariants();
+				}
 			}
-			finally
-			{
-				m_semaphore.Release();
-			}
+
+			AssertInvariants();
 		}
 
 		public static ConnectionPool GetPool(MySqlConnectionStringBuilder csb)
 		{
 			if (!csb.Pooling)
-				return null;
+				return new ConnectionPool(csb.Server, (int) csb.Port, csb.UserID, csb.Password, csb.Database, csb.ConnectionReset, true, 0, 1);
 
 			string key = csb.ConnectionString;
 			lock (s_lock)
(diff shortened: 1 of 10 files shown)


bgrainger:
I can look at this soon. Can you summarise the overall intent of these changes (so I can understand the purpose and review them according to that)?

You say this is "fixing the connection pool", but I'm not actually sure what is broken with it (#38 didn't provide any details). Is it something that can be simulated and verified in an integration test?

src/MySqlConnector/MySqlClient/ConnectionPool.cs
```diff
@@ -7,101 +9,191 @@ namespace MySql.Data.MySqlClient
...
 				}
 			}
 			finally
 			{
-				m_semaphore.Release();
+				// release all of the session slots
+				m_session_semaphore.Release(m_maximumSize);
+
+				// re-open the minimum sessions
+				await OpenMinimumSessions();
```
bgrainger: I'm not sure that `ClearAsync` should reopen connections; AFAIK it's commonly used to make sure all connections *are* shut down e.g., when exiting a program.

src/MySqlConnector/MySqlClient/ConnectionPool.cs
```diff
@@ -7,101 +9,191 @@ namespace MySql.Data.MySqlClient
...
 			if (!csb.Pooling)
 				return null;
 
-			string key = csb.ConnectionString;
-			lock (s_lock)
+			var key = csb.ConnectionString;
+
+			ConnectionPool pool;
+			// try without the lock
+			if (!s_pools.TryGetValue(key, out pool))
```
bgrainger: This isn't threadsafe.

src/MySqlConnector/MySqlClient/ConnectionPool.cs
```diff
@@ -7,101 +9,191 @@ namespace MySql.Data.MySqlClient
...
+				}
 
-				if (m_sessions.Count > 0)
-					return m_sessions.Dequeue();
+				session = new MySqlSession(this);
+				await session.ConnectAsync(m_connectionStringBuilder, cancellationToken);
+				return session;
 			}
-			finally
+			catch
```
bgrainger: Why not keep it as `finally`?

src/MySqlConnector/MySqlClient/ConnectionPool.cs
caleblloyd: Good point, this should be taken out

src/MySqlConnector/MySqlClient/ConnectionPool.cs
caleblloyd: This is a 2-part check. First part tries acquiring without a lock.  If it gets null, then it tries acquiring with lock.  Thought behind this is GetPoolAsync is called once per connection but it is only ever instantiated once in an application (and never removed).  This reduces lock contention.

src/MySqlConnector/MySqlClient/ConnectionPool.cs
caleblloyd: The `m_session_semaphore` is now set to the size of `m_max_connections`.  It is held until the connection is returned to the pool in the Return(session) method.  If there is an exception, it should be returned to the pool immediately since the session never leaves the pool.

src/MySqlConnector/MySqlClient/ConnectionPool.cs
bgrainger: `s_pools.Add` could be called (on line 126) from a different thread while this thread calls `TryGetValue`. That is not threadsafe and will crash if the dictionary's internal state is inconsistent.

caleblloyd:
> Can you summarise the overall intent of these changes (so I can understand the purpose and review them according to that)?

The current implementation does not track the number of Connections that are leased out of the pool.  Instead, it just checks to see if it over-created sessions once they are returned.  If load is high and sessions aren't getting returned to the pool, the current implementation will hand out unlimited connections and overwhelm the database.

This PR changes `m_session_semaphore` to track how many Connections are leased out of the pool.  It counts them back into the pool in Return(session).

I moved a few methods from MySqlConnection to MySqlSession because the ConnectionPool needed to access them, and they seemed more appropriate in MySqlSession.

src/MySqlConnector/MySqlClient/ConnectionPool.cs
caleblloyd: Ah.  Would you prefer that I change it to a ConcurrentDictionary or just use the Async lock every time?

caleblloyd: Fixed by switching to ConcurrentDictionary

caleblloyd:
Added the fix for making s_pools a ConcurrentDictionary and not re-opening minimum connections on ClearPools

Still having issues with SideBySide.New hanging though.  Any luck looking at that?

bgrainger:
> If load is high and sessions aren't getting returned to the pool fast enough, the current implementation will hand out unlimited connections and overwhelm the database.

That makes total sense. I need to find out why MySql.Data does in this instance... just block until a connection is available? Or fail immediately with an exception?

src/MySqlConnector/MySqlClient/ConnectionPool.cs
bgrainger: Using a well-tested framework class is a good approach. Can try to replace it if performance profiling indicates it's a bottleneck.

caleblloyd:
It definitely blocks until a connection is available. I confirmed by load testing it.

bgrainger:
Although it will cause merge conflicts, I'm going to start breaking this up into smaller commits, to narrow down the cause of test failures.

First pushed: moving `ConnectAsync` etc. into `MySqlSession`: 54e3a8df04af2030d990305ce96a4d88aeb8cd7c

bgrainger:
I abandoned the `Semaphore` approach and went with a simple `Monitor`/`lock` approach that let me more easily maintain invariants: 2dd346cee28c95e30b3977264ea2b10e72f4ece6

(The issue I ran into with the semaphore is that I couldn't atomically wait for it and increment a counter, so I couldn't find a way to assert invariants about the state of the pool--i.e., the number of active+idle connections--to verify it was behaving correctly. The downside with the new approach is that a thread will actually block in `GetSessionAsync` while it waits for a connection to become available. However, I'd like to make it correct first, then optimise it to be non-blocking later on.)

It doesn't implement `MinimumPoolSize` yet, but that can come later.

bgrainger:
The new code can maintain 400 req/s:

```
PS> .\stress.ps1 400 20s async
Rate = 400 RPS
Duration = 20s
Targets = targets-async.txt
Requests      [total, rate]            8000, 400.05
Duration      [total, attack, wait]    20.0086398s, 19.9975s, 11.1398ms
Latencies     [mean, 50, 95, 99, max]  5.626899ms, 5.1117ms, 11.6336ms, 14.1336ms, 19.6386ms
Bytes In      [total, mean]            92626005, 11578.25
Bytes Out     [total, mean]            2978000, 372.25
Success       [ratio]                  100.00%
Status Codes  [code:count]             200:8000
Error Set:
```

I am getting some failures above that, so will dig in further.

caleblloyd:
@bgrainger I'm seeing the same thing on perf tests.

Spinlocks are pretty detrimental to performance when you get a lot of threads waiting at load.  I think we should try to switch it back to SemaphoreSlim if possible

bgrainger:
I'm going to try porting the new code to `SemaphoreSlim` (if I can determine a good analogue to `Monitor.Pulse`).

bgrainger:
Switched to SemaphoreSlim: 187eb0bdb27010b38c780508b0575e2a3b5f332d

I have an idea for implementing an alternative to `Monitor.Pulse`, but don't have the time to implement it right now.

It still fails at 500 req/s but it doesn't grow to 100s or 1000s of blocked threads (as it did with the previous implementation).

greghroberts:
Npgsql's implementation is certainly something to refer to..
https://github.com/npgsql/npgsql/blob/74651fc696dbb84b46705de896870c038d4fd5b0/src/Npgsql/PoolManager.cs

Also System.Data.SqlClient 
https://github.com/dotnet/corefx/blob/master/src/System.Data.SqlClient/src/System/Data/SqlClient/SqlConnectionFactory.cs

caleblloyd:
Is there a reason we need to be able to keep track of what state the connections are in inside of the pool?  What's wrong with counting them out of the pool and counting them back into the pool with a `SemaphoreSlim` with max size of max_connections?

I've also corrected the behavior of `Pooling=False`.  There should be no connection pool at all with `Pooling=False`, the app should manage it's own connections at this point.

I've updated this PR to be based off the `tests` branch starting at 187eb0bdb27010b38c780508b0575e2a3b5f332d but reverted it to use the count in / count out method.  I'm maintaining 1500 RPS easily with this approach.

Still having an issue with some of the SideBySide.New tests.  The tests don't seem to be running serially and they don't print the test name, so it's very hard to track down deadlocks in the tests.  Any way we can at least get the tests to log their name?

src/MySqlConnector/Serialization/MySqlSession.cs
```diff
@@ -164,41 +170,44 @@ private void VerifyConnected()
...
 				{
-					cancellationToken.ThrowIfCancellationRequested();
-
-					Socket socket = null;
+					var socket = new Socket(ipAddress.AddressFamily, SocketType.Stream, ProtocolType.Tcp);
 					try
 					{
-						socket = new Socket(ipAddress.AddressFamily, SocketType.Stream, ProtocolType.Tcp);
-						using (cancellationToken.Register(() => socket.Dispose()))
+						var connectTasks = new List<Task>
```
caleblloyd: The Socket tasks do not take a cancellation token as an argument, so the Timeout Cancellation token will not do anything.  In order to find a timeout, I have leveraged Task.WaitAny with a timeout task, and checked the socket status.

tests/SideBySide.New/ConnectionPool.cs
```diff
@@ -80,37 +80,7 @@ public void ResetConnection(object connectionReset, int poolSize, long expected)
 		}
 
 		[Fact]
-		public async Task ExhaustConnectionPoolWithTimeout()
```
caleblloyd: I believe this test to be incorrect.  I think that ConnectionTimeout is a timeout for a TCP Connection to be made to the MySQL server.  I don't think it applies to a pooled connection being retrieved from the pool.

tests/SideBySide.New/ConnectionPool.cs
```diff
@@ -148,9 +118,5 @@ public void ResetConnection(object connectionReset, int poolSize, long expected)
 				connection.Dispose();
 		}
 
-		public void Dispose()
-		{
-			MySqlConnection.ClearAllPools();
```
caleblloyd: XUnit instantiates all test classes at the very beginning.  Since pools are shared among tests, `ClearAllPools` wreaks havoc on other tests that are still running.

tests/SideBySide.New/DatabaseFixture.cs
```diff
@@ -22,7 +22,6 @@ protected virtual void Dispose(bool disposing)
 			if (disposing)
 			{
 				Connection.Dispose();
-				MySqlConnection.ClearAllPools();
```
caleblloyd: XUnit instantiates all test classes at the very beginning.  Since pools are shared among tests, `ClearAllPools` wreaks havoc on other tests that are still running.

caleblloyd:
I've gotten tests working locally and on appveyor with this implementation.  Not sure why Travis is failing.  Running solid at 1500 RPS.  @bgrainger please read my code review comments.

tests/SideBySide.New/ConnectionPool.cs
bgrainger: MySql.Data uses this value to time out the connection when the pool is full. (This test passes when compiled with mysql-connector-net.) This behaviour makes sense to me and I think we should clone it. (The alternatives would be failing immediately or blocking forever, and none of those seems preferable.)

tests/SideBySide.New/ConnectionPool.cs
bgrainger: I agree that clearing pools at the end of each test isn't ideal, but I'm not clear on how this "wreaks havoc"; are there tests that specifically rely on idle connections being available? 

This may actually indicate a bug in `ClearAllPools`; it shouldn't be implemented as a "destructive" operation for existing open connections but as per the [documentation](https://dev.mysql.com/doc/dev/connector-net/html/M_MySql_Data_MySqlClient_MySqlConnection_ClearPool.htm)

> ClearPool clears the connection pool that is associated with the connection. If additional connections associated with connection are in use at the time of the call, they are marked appropriately and are discarded (instead of being returned to the pool) when Close is called on them.

(I don't think my implementation guarantees that, so I'll need to fix that up.)

src/MySqlConnector/Serialization/MySqlSession.cs
bgrainger: What was wrong with registering a handle on the cancellationToken that called `Socket.Dispose`?

As per [MSDN](https://msdn.microsoft.com/en-us/library/tad07yt6.aspx):

> To cancel a pending call to the BeginConnect method, close the Socket. When the Close method is called while an asynchronous operation is in progress, the callback provided to the BeginConnect method is called. A subsequent call to the EndConnect method will throw an ObjectDisposedException to indicate that the operation has been cancelled. 

(The Socket async API has lots of "oddities" and this is one of them.)

bgrainger: (.NET Core doesn't have `Close`, but it does have `Dispose` which I checked does the same thing.)

bgrainger:
> Not sure why Travis is failing.

Very odd; can't tell yet if something changed in the environment that's causing that failure.

bgrainger:
> Running solid at 1500 RPS.

Awesome! :+1: Will merge this once I can get the tests figured out.

tests/SideBySide.New/ConnectionPool.cs
caleblloyd: Ah, good find!  This should be an easy fix then - let's open another issue for it after this is merged.

bgrainger:
Shipped this in [0.1.0-alpha21](https://www.nuget.org/packages/MySqlConnector/0.1.0-alpha21).

src/MySqlConnector/Serialization/MySqlSession.cs
caleblloyd: I can't remember exactly - I'd have to go back and re-test.  Does an `await` command called on `socket.OpenAsync` without a cancellationToken block and not consider the handle on the CancellationToken?

tests/SideBySide.New/ConnectionPool.cs
caleblloyd: Ok, we should reimplement this using a timeout CancellationToken inside of `ConnectionPool.GetSessionAsync` then

PR #60, opened 2016-09-20
Add Async Transaction Methods

caleblloyd:
I'm seeing a big slowdown when using Transactions with Asynchronous query methods.  Threads are getting held up due to the mixing of Synchronous and Asynchronous requests to the database.

Right now, transactions can only be started, commited, or rolled back using Synchronous methods.  This PR adds Asynchronous methods for Transactions.

### DIFF
--- a/src/MySqlConnector/MySqlClient/MySqlConnection.cs
+++ b/src/MySqlConnector/MySqlClient/MySqlConnection.cs
@@ -23,7 +23,22 @@ namespace MySql.Data.MySqlClient
 
 		public new MySqlTransaction BeginTransaction() => (MySqlTransaction) base.BeginTransaction();
 
+		public async Task<MySqlTransaction> BeginTransactionAsync(CancellationToken cancellationToken = default(CancellationToken))
+		{
+			return await BeginDbTransactionAsync(IsolationLevel.Unspecified, cancellationToken).ConfigureAwait(false) as MySqlTransaction;
+		}
+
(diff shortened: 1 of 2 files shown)


src/MySqlConnector/MySqlClient/MySqlConnection.cs
```diff
@@ -23,8 +23,23 @@ public MySqlConnection(string connectionString)
...
+		{
+			return await BeginDbTransactionAsync(isolationLevel, cancellationToken).ConfigureAwait(false) as MySqlTransaction;
+		}
+
 		protected override DbTransaction BeginDbTransaction(IsolationLevel isolationLevel)
 		{
+			return BeginDbTransactionAsync(isolationLevel).GetAwaiter().GetResult();
+		}
+
+		private async Task<DbTransaction> BeginDbTransactionAsync(IsolationLevel isolationLevel, CancellationToken cancellationToken = default(CancellationToken))
```
bgrainger: Can the casts above be avoided if this returns `Task<MySqlTransaction>`? Since it's a private method, I see no reason to not return the most-derived type.

caleblloyd: `protected override DbTransaction BeginDbTransaction` needs a `DbTransaction` type out because the extended class `DbConnection` calls it

bgrainger: `MySqlTransaction` will be implicitly downcasted to `DbTransaction` in that instance.

bgrainger:
Shipped in [0.1.0-alpha23](https://www.nuget.org/packages/MySqlConnector/0.1.0-alpha23).

PR #63, opened 2016-09-25
make TryAsyncContinuation async

caleblloyd:
- Helps increase efficiency for Async calls by removing Synchronous calls in Async Code
- Helps increase efficiency for Async and Sync calls by reducing need for context switching

I used a method that retrieved 10 blog posts from the Performance Test database.  I logged every call to `GetAwaiter().GetResult()` to see what Managed Thread ID the calls were running on.

Async calls to `GetAwaiter().GetResult()` Before Change:

	Thread ID: 4 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 3 Kicking off task
	Thread ID: 3 Kicking off task
	Thread ID: 3 Kicking off task
	Thread ID: 3 Kicking off task
	Thread ID: 3 Kicking off task

Async calls to `GetAwaiter().GetResult()` After Change:

    Thread ID: 5 Kicking off task (this happens in MysqlDataReader.Dispose when it calls NextResult)

Sync calls to `GetAwaiter().GetResult()` Before Change:

	Thread ID: 5 Kicking off task
	Thread ID: 4 Kicking off task
	Thread ID: 10 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 10 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 4 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 10 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 4 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 10 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task

Sync calls to call to `GetAwaiter().GetResult()` After Change:

	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task
	Thread ID: 5 Kicking off task

### DIFF
--- a/src/MySqlConnector/Serialization/MySqlSession.cs
+++ b/src/MySqlConnector/Serialization/MySqlSession.cs
@@ -219,15 +219,16 @@ namespace MySql.Data.Serialization
 			if (task.Status == TaskStatus.RanToCompletion)
 				return task;
 
-			return task.ContinueWith(TryAsyncContinuation, cancellationToken, TaskContinuationOptions.LazyCancellation | TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
+			return TryAsyncContinuation(task);
 		}
 
-		private void TryAsyncContinuation(Task task)
+		private async Task TryAsyncContinuation(Task task)
 		{
+			await task.ConfigureAwait(false);
 			if (task.IsFaulted)
 			{
 				SetFailed();
-				task.GetAwaiter().GetResult();
+				await task.ConfigureAwait(false);
 			}
 		}
 
@@ -249,16 +250,15 @@ namespace MySql.Data.Serialization
 				return new ValueTask<PayloadData>(tcs.Task);
 #endif
 			}
-
-			return new ValueTask<PayloadData>(task.AsTask()
-				.ContinueWith(TryAsyncContinuation, cancellationToken, TaskContinuationOptions.LazyCancellation | TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default));
+			return new ValueTask<PayloadData>(TryAsyncContinuation(task.AsTask()));
 		}
 
-		private PayloadData TryAsyncContinuation(Task<PayloadData> task)
+		private async Task<PayloadData> TryAsyncContinuation(Task<PayloadData> task)
 		{
+			await task.ConfigureAwait(false);
 			if (task.IsFaulted)
 				SetFailed();
-			var payload = task.GetAwaiter().GetResult();
+			var payload = await task.ConfigureAwait(false);
 			payload.ThrowIfError();
 			return payload;
 		}


bgrainger:
This PR is replacing a hand-written continuation with a [compiler-generated one](http://www.filipekberg.se/2013/01/16/what-does-async-await-generate/) that is [less efficient due to the async state machine](https://github.com/dotnet/roslyn/issues/10449). It also allocates extra (unnecessary) Tasks.

`task.ContinueWith(TryAsyncContinuation)` schedules `TryAsyncContinuation` to run when the task is completed. There is no "blocking" because the antecedent task has finished, so `GetAwaiter().GetResult()` returns immediately (or throws).

Adding `await task.ConfigureAwait(false)` to the body of `TryAsyncContinuation` sets up an almost identical continuation, except it requires extra compiled-generated code and temporary objects. Behind the scenes, the compiler-generated code calls `GetResult()` on the task's awaiter.

Your change does suggest an approach that I haven't tested yet, though: using `GetAwaiter().OnCompleted()` to schedule the continuation instead of `ContinueWith`. However, the documentation explicitly says "This API supports the product infrastructure and is not intended to be used directly from your code." so it seems unwise. I haven't checked the implementation, but I'd be very surprised if `OnCompleted` wasn't implemented in terms of `ContinueWith`, though.

bgrainger:
> but I'd be very surprised if OnCompleted wasn't implemented in terms of ContinueWith, though

[`OnCompleted`](https://github.com/dotnet/coreclr/blob/775003a4c72f0acc37eab84628fcef541533ba4e/src/mscorlib/src/System/Runtime/CompilerServices/TaskAwaiter.cs#L206) calls `SetContinuationForAwait` which [calls `AddTaskContinuation`](https://github.com/dotnet/coreclr/blob/19346e58c6443638c245f85ee0b31a8c6e7af872/src/mscorlib/src/System/Threading/Tasks/Task.cs#L2984); `ContinueWith` calls `ContinueWithCore` which also [calls `AddTaskContinuation`](https://github.com/dotnet/coreclr/blob/19346e58c6443638c245f85ee0b31a8c6e7af872/src/mscorlib/src/System/Threading/Tasks/Task.cs#L4679). So the two should be roughly equivalent in terms of overhead.

caleblloyd:
Thanks for the information and references.

> There is no "blocking" because the antecedent task has finished

This makes sense.

> Behind the scenes, the compiler-generated code calls GetResult() on the task's awaiter

Good to know.  So it's running on a different thread anyways, I just can't log the result because it's compiler generated.  This makes total sense.  In both of the Sync Before/After examples there are 14 calls on `Thread ID: 5`.  The only difference is I've masked logging.

I'll read up more on these articles, it's great to know what's actually going on in the compiler!

PR #65, opened 2016-09-25
travis ci: use docker image for MySql 5.7

caleblloyd:
Travis CI seems to fail randomly.  I think it may be a problem with the MySQL 5.7 service not starting.

This PR switches Travis CI to use a Docker image for MySQL 5.7, which may be more reliable for us.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,15 @@
 sudo: required
 dist: trusty
-addons:
-  apt:
-    packages:
-    - mysql-server-5.6
-    - mysql-client-core-5.6
-    - mysql-client-5.6
-services: mysql
+services: docker
 
 before_install:
-  - bash .travis.install-mysql-5.7.sh
-  - |-
-    mysql -u root -e "CREATE USER mysqltest IDENTIFIED BY 'test;key=\"val'; GRANT ALL ON *.* TO mysqltest;"
-    mysql -u root -e "CREATE USER no_password;"
+  - docker pull mysql:5.7
+  - docker run --name mysql -e MYSQL_ROOT_PASSWORD='test' -p 3306:3306 -d mysql:5.7
   - sudo sh -c 'echo "deb [arch=amd64] https://apt-mo.trafficmanager.net/repos/dotnet-release/ trusty main" > /etc/apt/sources.list.d/dotnetdev.list'
   - sudo apt-key adv --keyserver apt-mo.trafficmanager.net --recv-keys 417A0893
   - sudo apt-get update
   - sudo apt-get install -y dotnet-dev-1.0.0-preview2-003121
+  - docker exec -it mysql mysql -uroot -ptest -e "CREATE USER 'mysqltest'@'%' IDENTIFIED BY 'test;key=\"val'; GRANT ALL ON *.* TO mysqltest; CREATE USER 'no_password'@'172.17.0.1';"
 
 script:
   - dotnet restore
--- a/.travis.install-mysql-5.7.sh
+++ b/.travis.install-mysql-5.7.sh
@@ -1,6 +1,0 @@
-echo mysql-apt-config mysql-apt-config/select-server select mysql-5.7 | sudo debconf-set-selections
-wget http://dev.mysql.com/get/mysql-apt-config_0.7.3-1_all.deb
-sudo dpkg --install mysql-apt-config_0.7.3-1_all.deb
-sudo apt-get update -q
-sudo apt-get install -q -y -o Dpkg::Options::=--force-confnew mysql-server
-sudo mysql_upgrade


.travis.yml
```diff
@@ -1,18 +1,10 @@
...
-    - mysql-client-core-5.6
-    - mysql-client-5.6
-services: mysql
+services: docker
 
 before_install:
-  - bash .travis.install-mysql-5.7.sh
-  - |-
-    mysql -u root -e "CREATE USER mysqltest IDENTIFIED BY 'test;key=\"val'; GRANT ALL ON *.* TO mysqltest;"
-    mysql -u root -e "CREATE USER no_password;"
```
bgrainger: Need to create this `no_password` user too.

bgrainger:
> I think it may be a problem with the MySQL 5.7 service not starting.

I haven't checked every failed build, but I've consistently seen this message (in `DataTypesFixture`), which implies that MySQL is running:

> Incorrect string value: '\xD0\x90\xD0\x91\xD0\x92...' for column 'cp1251' at row 4

`D0 90` are the UTF-8 bytes for U+0410, Cyrillic Capital Letter A, so it seems like a code page conversion is failing somewhere, and I don't know why.

I know correlation doesn't imply causation, but IIRC this only started (failing randomly) since I merged #56. I wonder if there's some race condition that's occasionally hit during test fixture initialisation? I suppose I could start a test branch off the commit before that one; if *that* fails, then it's more likely to be some environmental Travis CI change.

OTOH, if this is a race condition in the connector, why does it only happen with Travis CI? (I haven't seen it locally or on Appveyor).

caleblloyd:
@bgrainger yup, still failing with docker mysql 5.7.  Let me see if I can debug since I'm already on linux (and Travis is also linux)

caleblloyd:
I get it about 25% of the time.  Maybe it has to do with the first request that goes over the pooled connection?

Here's the buggy stream in pcap. (you can view with [Wireshark](https://www.wireshark.org/))

[error-stream.zip](https://github.com/bgrainger/MySqlConnector/files/491930/error-stream.zip)

bgrainger:
This is helpful; can you also upload a packet capture of a successful test run? (It may be significantly larger...)

caleblloyd:
Here is a successfully completed stream pcap:

[success-stream.zip](https://github.com/bgrainger/MySqlConnector/files/491941/success-stream.zip)

bgrainger:
I have a theory (but will probably need a successful test run packet capture to confirm).

In `error-stream`, it logged in successfully (setting the connection code page to `utf8mb4_bin`) in packet 6. It then sends multiple connection reset packets (as the existing connection is repeatedly retrieved from the pool) until it tries to create the `datatypes` schema in packet 57. This fails with the "incorrect string value" error.

If it's possible that resetting the connection has the (undocumented?) side-effect of resetting the connection code page to the server default, then we are in an unexpected state.

The successful test might happen when the connection pool is empty, so a new connection has to be created. In this case, the test would log in, setting the connection code page, then immediately execute its SQL (which works as expected).

(It ought to be possible to test this theory with a low-level test that sends packets that recreate this setup.)

caleblloyd:
> The successful test might happen when the connection pool is empty, so a new connection has to be created. In this case, the test would log in, setting the connection code page, then immediately execute its SQL (which works as expected).

That's exactly what happens in the success-stream that I just uploaded

bgrainger:
In the successful capture, the packets were (to server) login, (from server) OK, (to server) `create schema datatypes`. This seems to back up my guess above.

caleblloyd:
As far as the switch to Docker for the database - I think we should keep it.  It'd be super easy to run MySql 5.6 on port 3307 and MariaDB on port 3308 and run the SiteBySide.New tests over and over again

bgrainger:
[`character_set_connection`](http://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_character_set_connection), despite the name, is a session-level variable. 

[`COM_RESET_CONNECTION`](https://dev.mysql.com/doc/internals/en/com-reset-connection.html) "resets the session state". This probably means that `character_set_connection` is set back to the server default, and doesn't preserve the code page specified in the login packet.

I was considering the code page to be a *connection*-level setting, not a session-level one but that's apparently not how MySQL works.

Thus, after resetting the connection, we need to send a [`SET NAMES`](http://dev.mysql.com/doc/refman/5.7/en/set-names.html) command to change the code page back to what we want. Or just revert to the MySQL 5.6 behaviour of re-logging in (which would probably be the same number of server roundtrips?).

It'd be interesting to see whether MariaDB or Percona Server for MySQL implement the same behaviour.

bgrainger:
> As far as the switch to Docker for the database - I think we should keep it.

Agreed; this is a much less hacky way to run MySQL 5.7 for the tests.

PR #69, opened 2016-09-30
Handle pre-CLIENT_PROTOCOL_41 errors

### LINKED ISSUE #68: Consider error messages starting with 0x48
I have hit 2 MySql error messages starting with the byte 0x48 instead of 0x23.

These exceptions are:

- 1130 Host 'x.x.x.x' is not allowed to connect to this MySql Server
- 1203 Too Many Connections

Presently the exception comes across as:

    Unhandled Exception: System.FormatException: Expected to read 0x23 but got 0x48
       at MySql.Data.ByteArrayReader.ReadByte(Byte value)

I'm working up a fix to check for 0x23 and 0x48 that will print:

    Unhandled Exception: MySql.Data.MySqlClient.MySqlException: '172.17.0.1' is not allowed to connect to this MySQL server
       at MySql.Data.Serialization.PayloadData.ThrowIfError()

caleblloyd:
Fixes #68

### DIFF
--- a/src/MySqlConnector/Serialization/ErrorPayload.cs
+++ b/src/MySqlConnector/Serialization/ErrorPayload.cs
@@ -21,9 +21,18 @@ namespace MySql.Data.Serialization
 			reader.ReadByte(Signature);
 
 			var errorCode = reader.ReadUInt16();
-			reader.ReadByte(0x23);
-			var state = Encoding.ASCII.GetString(reader.ReadByteString(5));
-			var message = Encoding.UTF8.GetString(reader.ReadByteString(payload.ArraySegment.Count - 9));
+			var stateMarker = Encoding.ASCII.GetString(reader.ReadByteString(1));
+			string state, message;
+			if (stateMarker == "#")
+			{
+				state = Encoding.ASCII.GetString(reader.ReadByteString(5));
+				message = Encoding.UTF8.GetString(reader.ReadByteString(payload.ArraySegment.Count - 9));
+			}
+			else
+			{
+				state = "HY000";
+				message = stateMarker + Encoding.UTF8.GetString(reader.ReadByteString(payload.ArraySegment.Count - 4));
+			}
 			return new ErrorPayload(errorCode, state, message);
 		}
 


src/MySqlConnector/ByteArrayReader.cs
```diff
@@ -40,10 +41,20 @@ public byte ReadByte()
 			return m_buffer[m_offset++];
 		}
 
-		public void ReadByte(byte value)
+		public void ReadByte(params byte[] possibleValues)
```
bgrainger: This will cause an array allocation for every call to this methd.

src/MySqlConnector/ByteArrayReader.cs
```diff
@@ -40,10 +41,20 @@ public byte ReadByte()
...
 		{
-			if (ReadByte() != value)
-				throw new FormatException("Expected to read 0x{0:X2} but got 0x{1:X2}".FormatInvariant(value, m_buffer[m_offset - 1]));
+			var readByte = ReadByte();
+			if (possibleValues.Any(value => readByte == value))
```
bgrainger: This will allocate a lambda; a simple `foreach` might be better.

src/MySqlConnector/Serialization/ErrorPayload.cs
```diff
@@ -21,7 +21,7 @@ public static ErrorPayload Create(PayloadData payload)
 			reader.ReadByte(Signature);
 
 			var errorCode = reader.ReadUInt16();
-			reader.ReadByte(0x23);
+			reader.ReadByte(0x23, 0x48);
```
bgrainger: I think the real problem here is that I hard-coded `0x23`. The [documentation](https://dev.mysql.com/doc/internals/en/packet-ERR_Packet.html) just defines it as "# marker of the SQL State". I'm not sure where I got the value `0x23`; maybe I just observed that in practice and assumed it was a constant?

It does not appear to be a constant :smile: and the documentation doesn't specify the set of allowed values, so I would revert the changes to `ReadByte` and just change this line to `var sqlStateMarker = reader.ReadByte();` (and discard the value).

bgrainger: `0x23` is `#`, so I guess I interpreted "# marker of the SQL State" as meaning it was the constant `'#'`.

Still not sure what `0x48` (== `'H'`) means, but it must be an acceptable state marker. Since we're not doing anything with that value, it seems safe to ignore it if the rest of the packet looks right.

caleblloyd:
It looks like MariaDB's java driver holds all of the answers.  It's "old style" and "new style" error messages: https://github.com/MariaDB/mariadb-connector-j/blob/master/src/main/java/org/mariadb/jdbc/internal/packet/result/ErrorPacket.java#L70

I'll code it up according to their logic

bgrainger:
Oh, it's a pre-`CLIENT_PROTOCOL_41` error? That would make sense. Until we've sent back the `HandshakeResponse41Packet`, the server doesn't know if we support the new protocol.

bgrainger:
A simple approach may be to call `reader.ReadByteString` to read everything after the error code (as a string) then check if `message[0] == '#'` and perform the necessary Substring calls.

caleblloyd:
Updated PR with new changes

PR #82, opened 2016-10-04
Support MySqlConnection.ConnectionTimeout method

### LINKED ISSUE #57: Implement "Connect Timeout" connection string setting
Add `MySqlConnectionStringBuilder.ConnectionTimeout` setting (aliases: Connect Timeout, Connection Timeout, ConnectionTimeout). According to https://dev.mysql.com/doc/connector-net/en/connector-net-connection-options.html it has a default value of 15.

A "Timeout expired" `MySqlException` should be thrown when:
* the connection pool isn't full, but connecting to the server takes longer than the timeout
* the connection pool is full and a connection isn't returned in time (see #56)

caleblloyd:
This is the final change that I can see is needed to fully fix #57


bgrainger:
Shipped in [0.1.0](https://www.nuget.org/packages/MySqlConnector/0.1.0).

PR #83, opened 2016-10-04
set buffer size to 16KiB. fixes #80

### LINKED ISSUE #80: Consider Raising the default Buffer Size or using a Buffer Pool to decrease GC
The Performance Test's result set is ~14KB when it runs GET methods and retrieves 10 rows from the database.  Consider the following performance test setup, when the [buffer size](https://github.com/bgrainger/MySqlConnector/blob/master/src/MySqlConnector/Serialization/PacketTransmitter.cs#L15) is run with 4KiB (the default), and 64KiB.

	Rate = 1500 RPS
	Duration = 60s
	Targets = targets-async.txt

	                        mean    50%    95%    99%     max
	4KiB  Buffer Latencies  2.22s, 2.46s, 4.52s, 4.72s, 10.69s
	64KiB Buffer Latencies  0.19s, 0.08s, 0.76s, 0.86s,  6.65s

The 4KiB buffer is smaller than the packet that comes in, so a new buffer is [allocated](https://github.com/bgrainger/MySqlConnector/blob/master/src/MySqlConnector/Serialization/PacketTransmitter.cs#L206) on every request.  There is a huge slowdown due to all of the allocation and Garbage Collection that happens.

- A 64 KiB buffer size would allocate 6MiB of RAM with a connection pool of 100 connections.  This would cover pretty common use cases
- A buffer manager like [System.ServiceModel.Channels.BufferManager](https://msdn.microsoft.com/en-us/library/system.servicemodel.channels.buffermanager(v=vs.110).aspx) would be great.  Does something like this exist for .NET Core?

caleblloyd:
See comments in #80

### DIFF
--- a/src/MySqlConnector/Serialization/PacketTransmitter.cs
+++ b/src/MySqlConnector/Serialization/PacketTransmitter.cs
@@ -12,7 +12,7 @@ namespace MySql.Data.Serialization
 		{
 			m_socket = socket;
 			var socketEventArgs = new SocketAsyncEventArgs();
-			m_buffer = new byte[4096];
+			m_buffer = new byte[c_netBufferLength];
 			socketEventArgs.SetBuffer(m_buffer, 0, 0);
 			m_socketAwaitable = new SocketAwaitable(socketEventArgs);
 		}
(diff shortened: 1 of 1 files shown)


bgrainger:
Shipped in [0.1.0](https://www.nuget.org/packages/MySqlConnector/0.1.0).

PR #92, opened 2016-10-09
Support Synchronous connection string option. Fixes #91

### LINKED ISSUE #91: Support Synchronous connection string option.
If true, even Async methods run synchronously.

Theoretical benefits:
* Compare performance difference between sync and async without changing calling code.
* Better call stacks when debugging.

ejball:
If true, even Async methods run synchronously.

Theoretical benefits:
* Compare performance difference between sync and async without changing calling code.
* Better call stacks when debugging.

The setting has basic unit testing but I wasn't sure how to reliably test the change in behavior.

### DIFF
--- a/src/MySqlConnector/MySqlClient/MySqlConnectionStringBuilder.cs
+++ b/src/MySqlConnector/MySqlClient/MySqlConnectionStringBuilder.cs
@@ -118,6 +118,12 @@ namespace MySql.Data.MySqlClient
 			set { MySqlConnectionStringOption.UseAffectedRows.SetValue(this, value); }
 		}
 
+		public bool Synchronous
+		{
+			get { return MySqlConnectionStringOption.Synchronous.GetValue(this); }
+			set { MySqlConnectionStringOption.Synchronous.SetValue(this, value); }
+		}
+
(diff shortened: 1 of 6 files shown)


src/MySqlConnector/MySqlClient/MySqlConnectionStringBuilder.cs
```diff
@@ -118,6 +118,12 @@ public bool UseAffectedRows
 			set { MySqlConnectionStringOption.UseAffectedRows.SetValue(this, value); }
 		}
 
+		public bool Synchronous
```
bgrainger: `ForceSynchronous` might convey the meaning of this setting better; the default (`Synchronous = false`) doesn't stop the library being "synchronous".

ejball: Good point. I'll change it.

caleblloyd:
You won't notice much difference between sync and async until you start generating load. In that case, be sure to [set minimum threads in thread pool](https://github.com/bgrainger/MySqlConnector/issues/75)

Issue #88, opened 2016-10-05
Implement SSL Mode


caleblloyd:
From the [connection string reference](https://dev.mysql.com/doc/connector-net/en/connector-net-connection-options.html):

	SSL Mode, SslMode

	None - do not use SSL.
	Preferred - use SSL if the server supports it, but allow connection in all cases.
	Required - Always use SSL. Deny connection if server does not support SSL.
	VerifyCA - Always use SSL. Validate the CA but tolerate name mismatch.
	VerifyFull - Always use SSL. Fail if the host name is not correct.

I think this is a needed feature for 1.0 since many companies require SSL/TLS.  Also, some cloud MySQL providers such as Google Cloud SQL run over the public network so they pretty much require SSL/TLS.

A solution would most likely use the [SslStream](https://msdn.microsoft.com/en-us/library/system.net.security.sslstream(v=vs.110).aspx) library

Certificate validation based off the `SSL Mode` options would be performed in the [RemoteCertificateValidationCallback](https://msdn.microsoft.com/en-us/library/system.net.security.remotecertificatevalidationcallback(v=vs.110).aspx), checking the [SslPolicyErrors](https://msdn.microsoft.com/en-us/library/ms145055(v=vs.110).aspx)

I don't fully understand why someone would want to use the `SSL Mode=Preferred` option.  It seems to me like this would mask a server misconfiguration.  This may be an option that was added to support the legacy `Encrypt, UseSSL` connection string options.  I think we should consider not implementing `SSL Mode=Preferred`.

caleblloyd:
Work WIP branch at https://github.com/caleblloyd/MySqlConnector/commit/5d5c0e99228b8af65e8ec19799e779d7428fb0b3

How should we test this?  Just a small test file that connects and runs a query?  Or something more?

bgrainger:
My hope would be that we could re-run the full test suite with every combination of `SslMode=Required/None`, `UseCompression=true/false` (maybe `ForceSynchronous=true/false`). I'm not sure yet if xUnit has a way to re-run an entire test fixture with different settings; maybe CI would have to run `dotnet test` multiple times with different environment variables? I haven't dug into this yet.

caleblloyd:
> every combination of `SslMode=Required/None`, `UseCompression=true/false` (maybe `ForceSynchronous=true/false`)

I agree, we should test all 3.

[...]

Issue #31, opened 2016-08-10
Implement Compression


bgrainger:
Docs: https://dev.mysql.com/doc/internals/en/compression.html

`MySqlConnectionStringBuilder.UseCompression` already exists, but must be set to `false`; relax this restriction.

bgrainger:
Need to test compressed packets that span the 16MiB boundary (see [implementation note here](https://dev.mysql.com/doc/internals/en/example-several-mysql-packets.html)); the MySql.Data library does not currently handle those correctly:

* http://bugs.mysql.com/bug.php?id=83329
* http://bugs.mysql.com/bug.php?id=83330

mguinness:
When running a specific LINQ query (using Pomelo.EntityFrameworkCore.MySql) and with UseCompression set to true I get `System.InvalidOperationException: Packet received out-of-order. Expected 2; got 93.` error.  This is reproducible.

mguinness:
I did some more investigation and it appears related to the size of the data being returned.  In the example below the InvalidOperationException is raised, but if I set the limit to 74 it is not.

```C#
var cmd = db.CreateCommand();
cmd.CommandText = "SELECT * FROM `companies` LIMIT 75";

using (var rdr = cmd.ExecuteReader())
{
    while (rdr.Read()) { } //Packet received out-of-order. Expected 2; got 100.
}
```

It should also be noted that the loop works for the first 74 iterations, but it fails on the final one.

The following SQL returns 975 KB for the table in question if that is of any help.

```SQL
SELECT ROUND((DATA_LENGTH + INDEX_LENGTH) / 1024) AS `Size (KB)` 
FROM information_schema.TABLES
WHERE TABLE_NAME = 'companies'
```

bgrainger:
@mguinness The repro is probably highly dependent on the exact size of the data (either before or after compression). Would it be possible for me to get a copy of the data (solely for creating a repro of this bug, and to be deleted immediately afterwards)? My contact details are on my GitHub profile page.

bgrainger:
Splitting this bug out to #146.

PR #93, opened 2016-10-09
WIP: Refactor serialization

bgrainger:
This breaks `PacketTransmitter` up into multiple independent components:

* `IPayloadWriter`: writes a MySQL command and its data to an underlying transmission layer
 * standard implementation breaks it up into 16MiB packets with increasing sequence numbers
* `IPacketWriter`: serializes MySQL packets for transmission via a byte-oriented protocol
* `IByteWriter`: writes a byte array to a destination
 * standard implementation provides sync and async `Socket` support
* `Reader` versions of the three above interfaces, that invert their behaviour

Not only does this provide separation of concerns, it makes the "protocol stack" pluggable and should allow compression #31 and SSL #88 to be implemented much more easily.

This is a work-in-progress; pushed for early feedback. TODO:

- [x] squash bugfix commits
- [x] extract to one class per file
- [ ] code documentation
- [ ] decide on final namespaces
- [ ] implement compression (to validate the implementation)
- [ ] implement SSL proof-of-concept (to verify no major blocking issues)


caleblloyd:
The top level class that the driver interacts with should implement [Stream](https://msdn.microsoft.com/en-us/library/system.io.stream(v=vs.110).aspx) or [NetworkStream](https://msdn.microsoft.com/en-us/library/system.net.sockets.networkstream(v=vs.110).aspx)

This would allow for streams to be chained as Middleware.  For example, if compression and SSL were enabled, stream call stack would look like:

Write: Caller.Write -> MySqlStream.Write -> MySqlCompressedStream.Write -> SslStream.Write -> the wire
Read: Caller.Read -> MySqlStream.Read -> MySqlCompressedStream.Read -> SslStream.Read -> the wire

bgrainger:
I decided against that for the following reasons:

* The dichotomy between `Read`/`ReadAsync` and `Write`/`WriteAsync` is unhelpful for this project; it's proved to be more useful (so far) to use one method with `IOBehavior`
* `Stream` doesn't permit the use of `ValueTask`
* `Stream` has a lot of other functionality that is unneeded; it would be unclear which parts of `Stream` were necessary to override (or could be called)
* See other downsides of `Stream` here: https://github.com/davidfowl/channels
 * Avoiding `Stream` might make it easier to port to Channels in the future (low priority reason)
* If it would be useful to use a `Stream`, it should be possible to write an `IByteWriter` adapter that is implemented with a `Stream`, allowing a `Stream` to be easily plugged in as middleware

caleblloyd:
> I decided against that for the following reasons:

Another reason against it: In my testing, `SocketAwaitable` was faster than using async functions from the `NetworkStream` coming from `TcpClient.GetStream()`.  Maybe `SocketAwaitable` does a better job at avoiding copying?

> If it would be useful to use a Stream, it should be possible to write an IByteWriter adapter that is implemented with a Stream, allowing a Stream to be easily plugged in as middleware

This is probably what we'll need to do for SSL then.

The Compression piece looks to be MSQL Specific - on second thought, we can't just pipe the whole payload through a GZip stream.  Each MySql packet has to be compressed conditionally.

bgrainger:
> Maybe SocketAwaitable does a better job at avoiding copying?

It certainly does a lot less allocation; see Stephen Toub's writeup here: https://blogs.msdn.microsoft.com/pfxteam/2011/12/15/awaiting-socket-operations/

> Each MySql packet has to be compressed conditionally.

This was also part of the reason behind the separation into three separate interfaces: a `CompressingPacketWriter` should be able to send its output to an `IPayloadWriter` which will wrap the standard MySQL packet header around it (because the first four bytes in a regular and a compressed packet are the same).

PR #97, opened 2016-10-10
Combine read/write methods into single interface

caleblloyd:
**Re: Combining read/write methods into single interface**

Read/Write methods feel more natural in the same interface.  In order to support compression and SSL we are going to end up with the following classes:

- PacketHandler: an uncompressed MySQL packet reader / writer
- CompressedPacketHandler: a compressed MySQL packet reader / writer
- SocketByteHandler: reads/writes packets to a socket or SocketAwaitable
- SslByteHandler: reads/writes data to a SslStream

All of these implementations handles both reading and writing.  I cannot think of an implementation that handles just reading or just writing.

**Re: SSL**

For SslByteHandler - I can't figure out a good way to get `SslStream` working with `SocketAwaitable` without writing a ton code.  SSL is already a performance hit on encrypting/decrypting streams, and I think it's best to leave this implementation to the framework (which works with streams)

Therefore I think it would be best to open the connection using [TcpClient](https://msdn.microsoft.com/en-us/library/system.net.sockets.tcpclient(v=vs.110).aspx).

- Unencrypted will use `SocketByteHandler(TcpClient.Client)` (which calls to `SocketAwaitable` for async)
- Encrypted will use `SslByteHandler(TcpClient.GetStream())` (which calls to `Stream.ReadAsync` and `Stream.WriteAsync` for async)

I've added the `SslByteHandler` class so that we can decide on what this should look like.  I haven't done any of the work to connect it up in the `MySqlSession` yet though.


caleblloyd:
I have a working SSL WIP based off this change at https://github.com/caleblloyd/MySqlConnector/commit/5d5c0e99228b8af65e8ec19799e779d7428fb0b3

Once we finalize the interfaces, I'll open a PR for SSL

caleblloyd:
@bgrainger what are your thoughts on the interface reorganization in this PR?

bgrainger:
I think it was a mistake to break them into separate interfaces. (It makes some of the reader/writer classes more "SRP", but others need to share data to implement their functionality.) I will combine them into "XHandler" interfaces following your example. Then probably cherry-pick SSL code?

I'm also strongly considering whether to reduce it to just one interface (instead of three), possibly adding a separate `Flush` method (for hinting to future stages in the pipeline that a block of data has been written). But I'll complete this initial reorganisation first.

bgrainger:
I think all the code you pushed is now included in the HEAD of #93 (some I rewrote, then cherry-picked the `SslByteHandler`). So I'm going to close this; please let me know if I overlooked something.

I don't yet consider the new API final so I'm not going to merge it to master just yet. There are a couple of ideas I want to test out first.

PR #98, opened 2016-10-11
ssl connection string options

caleblloyd:
- Implements connection string options for #88 
- Adds an enumeration check for MySqlConnectionStringBuilder, needed for `SslMode`


caleblloyd:
I'll include this in the working SSL implementation PR

PR #99, opened 2016-10-12
Refactor Tests

caleblloyd:
- Refactors tests so that it will be easier to run tests against different setups (SSL Enabled, Compression On, Force Sync)
- Tests now use a gitignored config.json 
- Passwordless User tests and JSON tests can be disabled from config.json
- All test tables are created in the same schema, which is defined in the connection string in config.json.
  - This compartmentalizes test data
  - Should allow for multiple processes to test simultaneously, provided that they define different schemas in their connection strings (test SSL in one process with Database=ssltest, test compression in another with Database=compresstest, etc.)


tests/SideBySide.New/AppConfig.cs
```diff
@@ -0,0 +1,49 @@
...
+	    	.AddInMemoryCollection(DefaultConfig)
+		    .AddJsonFile("config.json")
+		    .Build();
+
+	    public static IConfiguration Config
+	    {
+		    get
+		    {
+			    if (Interlocked.Exchange(ref _configFirst, 1) == 0)
+				    Console.WriteLine("Config Read");
```
bgrainger: Debugging code?

caleblloyd: I was planning to use this from CI down the road.  If we are going to test 8 different combinations per database, we will want to do that in parallel with 8 different processes.  The "launching script" would read STDOUT until it sees "Config Read", then it would know it can swap the config.json out and launch the next process.

bgrainger:
Looks good for .NET Core; I'll verify (on Windows) that there aren't any showstoppers getting the baseline tests (against `MySql.Data`) running under the new framework.

tests/SideBySide.New/Constants.cs
```diff
@@ -1,23 +0,0 @@
...
-
-		public static MySqlConnectionStringBuilder CreateConnectionStringBuilder()
-		{
-			return new MySqlConnectionStringBuilder
-			{
-				Server = Server,
-				UserID = UserName,
-				Password = Password,
-				ConvertZeroDateTime = true,
-				UseAffectedRows = true,
```
bgrainger: Need this for the baseline tests; I'll add it to `config.json.example` even though it's superfluous for this connector.

PR #101, opened 2016-10-12
SSL Support

### LINKED ISSUE #88: Implement SSL Mode

caleblloyd:
- Fixes #88 
- Adds `IPayloadHandler.SetByteHandler` and `IPacketHandler.SetByteHandler` methods to their respective interfaces.  Needed to switch the underlying byte handler to SSL once the TLS handshake has been completed
- Uses `TcpClient` instead of `Socket` in `MySqlSession`
  - Unencrypted communication still happens at the `Socket` level using `TcpClient.Client`, `SocketByteHandler`, and `SocketAwaitable`
  - Encrypted communication happens at the `NetworkStream` level using `TcpClient.GetStream()`, `SslByteHandler`, and the appropriate stream methods
- Adds a second test run to TravisCI with `ssl mode=required`


bgrainger:
I'm getting the following exception under `dotnet test` on Windows:

```
    SideBySide.ConnectionPool.ResetConnection(connectionReset: False, poolSize: 11, expected: 0) [FAIL]
      System.IO.IOException : The handshake failed due to an unexpected packet format.
      Stack Trace:
           at System.Net.Security.SslState.StartReadFrame(Byte[] buffer, Int32 readBytes, AsyncProtocolRequest asyncRequest)
           at System.Net.Security.SslState.PartialFrameCallback(AsyncProtocolRequest asyncRequest)
        --- End of stack trace from previous location where exception was thrown ---
           at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
           at System.Net.Security.SslState.InternalEndProcessAuthentication(LazyAsyncResult lazyResult)
           at System.Net.Security.SslState.EndProcessAuthentication(IAsyncResult result)
           at System.Net.Security.SslStream.EndAuthenticateAsClient(IAsyncResult asyncResult)
           at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
```

(I think my local MySQL server is configured correctly because I'm able to connect with `mysql.exe --ssl-mode=REQUIRED (other ssl options) -u ssltest`. I'll debug this further to try to figure out what's happening.)

caleblloyd:
I'm seeing this also, running from Windows against the same MySQL database that is passing on linux, so I don't think it's your server configuration.  Let me try to get some decrypted packet captures.

src/MySqlConnector/Serialization/MySqlSession.cs
```diff
@@ -79,6 +86,80 @@ public MySqlSession(ConnectionPool pool, int poolGeneration)
...
+						default:
+							return sslMode == SslMode.Required;
+					}
+				};
+
+				var sslStream = new SslStream(m_tcpClient.GetStream(), false,
+					new RemoteCertificateValidationCallback(remoteCertificateCb),
+					new LocalCertificateSelectionCallback(localCertificateCb));
+				var clientCertificates = new X509CertificateCollection { certificate };
+				var sslProtocols = SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12;
```
caleblloyd: This is where the error on Windows is coming from. My MySQL server only speaks TLS1.1 but the handshake is trying to initiate with TLS1.2. Need to look into how to handle this properly.  Changing to `var sslProtocols = SslProtocols.Tls | SslProtocols.Tls11` on Windows works, but is not the correct solution.

caleblloyd:
MySQL doesn't like the `ClientHello` packet sent from windows when TLSv1.2 is enabled.  MySQL is responding to the `ClientHello` packet with "#08S01 Bad Handshake" and closing the TCP connection.  Here are the packet captures:

[tls-traces.zip](https://github.com/bgrainger/MySqlConnector/files/525350/tls-traces.zip)

I don't know if this is an issue with the `ClientHello` that windows is sending or an issue with MySQL, so I'm not sure who to open the upstream bug to.

In the short-term, our options are to:
1. Only enable TLS1 and TLS1.1 across all platforms
2. Enable TLS1 and TLS1.1 on Windows and enable TLS1, TLS1.1, and TLS1.2 on all other platforms

bgrainger:
I'm not very familiar with the specifics of TLS negotiation, but here's what I noticed when looking at the captures (thanks for grabbing them!):
* Linux sends a TLS 1.0 handshake containing a TLS 1.2 client hello; MySQL negotiates down to TLS 1.1 in its server hello
* Windows sends a TLS 1.2 handshake containing a TLS 1.2 client hello; MySQL doesn't recognise this.

My assumption would be that if your protocols are `SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12`, it should negotiate with the lowest-common-denominator, which doesn't appear to be happening. That makes me think the Windows `ClientHello` is the problem (but I'd really have to read the specs to confirm).

One other option would be to catch this exception and then renegotiate with just `Tls11`. However, there doesn't seem to be a lot of point to doing that unless we find a some combination of client and server that works; there's no value in adding a failing SSL handshake to every single connection.

> Enable TLS1 and TLS1.1 on Windows and enable TLS1, TLS1.1, and TLS1.2 on all other platforms

This seems like the best approach. TLS 1.2 wasn't available in MySQL Server [until 5.7.10](https://dev.mysql.com/doc/relnotes/mysql/5.7/en/news-5-7-10.html) (and isn't in Community Server) so it may not even be widely deployed. (This might become an issue if cloud MySQL providers started requiring TLS 1.2, but I haven't found any indication that this might happen after a quick search.)

bgrainger:
From [RFC 5246 Appendix E](https://tools.ietf.org/html/rfc5246#appendix-E):

> Earlier versions of the TLS specification were not fully clear on what the record layer version number (TLSPlaintext.version) should contain when sending ClientHello (i.e., before it is known which version of the protocol will be employed).  Thus, TLS servers compliant with this specification MUST accept any value {03,XX} as the record layer version number for ClientHello.
>
> TLS clients that wish to negotiate with older servers MAY send any value {03,XX} as the record layer version number.  Typical values would be {03,00}, the lowest version number supported by the client, and the value of ClientHello.client_version.  No single value will guarantee interoperability with all old servers, but this is a complex topic beyond the scope of this document.

It appears that the client code is permitted to send "TLS 1.2" as its record layer version number (as Windows does). However, it's also noted that this doesn't "guarantee interoperability".

Certainly it appears that MySQL Server is *not* "compliant with this specification" because it doesn't accept any value as the record layer version number.

bgrainger:
Neither the Windows .NET client nor MySQL Server are following [these recommendations](http://security.stackexchange.com/questions/26054/ssl-tls-version-change-during-renegotiation-or-resumption/26059#26059); this combination is breaking protocol negotiation.

bgrainger:
[Here's someone reporting](https://social.technet.microsoft.com/Forums/office/en-US/f1097c6d-8311-4ada-a0b5-1ec3100e92c7/schannel-tls-record-version-and-compatibility) essentially the same issue in Schannel (the native Windows SSL provider), but the problem doesn't seem to have been understood on the Microsoft side and no resolution was posted.

caleblloyd:
Very nice find.  While the Schannel side isn't a bug, it's still not following a best practice.  I'm not sure if it's worth trying to send them a bug report.  The MySQL side is definitely a bug and should be reported upstream.  I can report it or you can - let me know.

Working on disabling TLS 1.2 on windows now.  Getting the runtime information is turning out to be a pretty difficult task, this is the first language I've dealt with where it's not a one-liner.

bgrainger:
This potentially problematic behaviour in Schannel was [reported five years ago](https://www.ietf.org/mail-archive/web/tls/current/msg08218.html); I'm not sure we're going to see it changed any time soon...

I wouldn't be surprised if there would be significant interoperability problems caused by changing it; presumably Microsoft has good reasons for keeping it the way it is.

bgrainger:
> Getting the runtime information is turning out to be a pretty difficult task

I did not know (until now) that they took away `System.Environment.OSVersion` in .NET Core. The [workarounds](http://stackoverflow.com/questions/38790802/determine-operating-system-in-net-core) seem unnecessarily complicated...

caleblloyd:
I ended up importing [System.Runtime.InteropServices.RuntimeInformation](https://www.nuget.org/packages/System.Runtime.InteropServices.RuntimeInformation/) in both Netstandard 1.3 and Net45 so we'd have a common runtime information API

Tests are passing for me from Windows after the latest commit

caleblloyd:
TLS 1.2 bug opened upstream to MySQL at http://bugs.mysql.com/bug.php?id=83347

PR #103, opened 2016-10-13
Refactor protocol layer

bgrainger:
Reduces the three `I*Handler` interfaces down to one common `IProtocolLayer` interface. Adds a standard way of "injecting" a new layer into the protocol stack, which should work for both SSL and compression. Makes the "conversation" part of the protocol instead of being relegated to a separate interface.


bgrainger:
Will eventually introduce this code as part of implementing compression.

PR #111, opened 2016-10-20
ConnectionSettings class. fixes #110

### LINKED ISSUE #110: Create ConnectionSettings class in Serialization Namespace
As our list of supported `MySqlConnectionStringBuilder` options grows, the number of parameters that must be tracked by the `MySqlConnection`, `ConnectionPool`, and `MySqlSession` is also growing.

Add an internal, immutable `ConnectionSettings` class to hold connection-related options.  This class will take a `MySqlConnectionString` and check/convert it's values to optimal data types for internal use.

Pass the new `ConnectionSettings` class around in internal functions instead of individual parameters.

caleblloyd:
See #110 for more info

### DIFF
--- a/tests/SideBySide.New/AppConfig.cs
+++ b/tests/SideBySide.New/AppConfig.cs
@@ -20,8 +20,8 @@ namespace SideBySide
 		private static int _configFirst;
 
 		private static IConfiguration ConfigBuilder { get; } = new ConfigurationBuilder()
-			.SetBasePath(File.Exists(Path.Combine(Directory.GetCurrentDirectory(), "config.json")) ? Directory.GetCurrentDirectory() :
-				Path.Combine(Directory.GetCurrentDirectory(), "tests", "SideBySide.New"))
+			.SetBasePath(File.Exists(Path.Combine(Directory.GetCurrentDirectory(), "config.json")) || new DirectoryInfo(Directory.GetCurrentDirectory()).Name == "SideBySide.New"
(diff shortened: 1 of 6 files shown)


tests/SideBySide.New/AppConfig.cs
```diff
@@ -20,8 +20,8 @@ public static class AppConfig
 		private static int _configFirst;
 
 		private static IConfiguration ConfigBuilder { get; } = new ConfigurationBuilder()
-			.SetBasePath(File.Exists(Path.Combine(Directory.GetCurrentDirectory(), "config.json")) ? Directory.GetCurrentDirectory() :
-				Path.Combine(Directory.GetCurrentDirectory(), "tests", "SideBySide.New"))
+			.SetBasePath(Path.GetFileName(Path.GetDirectoryName(Path.Combine(Directory.GetCurrentDirectory(), "config.json"))) == "SideBySide.New"
```
bgrainger:
```
Path.GetDirectoryName(Path.Combine(Directory.GetCurrentDirectory(), "config.json"))
```

seems like a complicated way of writing `Directory.GetCurrentDirectory()` (i.e., adding and removing `\config.json` from the path).

Additionally, this breaks the Baseline tests, which are looking for the tests in the current working directory (set by `xunit.runner`) not in `SideBySide.New`. Was there a reason this needed to be changed?

caleblloyd:
Can't find a good way to get the currently executing test directory in XUnit.

The problem was when there was no config.json file in `tests/SideBySide.New`, the code would try to look in `tests/SideBySide.New/tests/SideBySide.New` and throw an unhelpful exception.

Added a check for this case and also merged #112, Baseline tests are working again.

PR #112, opened 2016-10-20
Build and run baseline tests (VS solution) under Appveyor.

bgrainger:


caleblloyd:
This is super helpful, running Baseline tests on Linux with mono has never really worked that well for me.

PR #115, opened 2016-10-21
Implement protocol compression

bgrainger:
Opening PR for code review. 

This fully implements MySQL protocol compression, but may need some code cleanup before merging it. It's fixed the major blocking issue from previous attempts, which is that sequence numbers couldn't be verified (due to compression rewriting them).


caleblloyd:
I ran through all of the performance tests yesterday on this branch and everything looks good.

The compressed connection is running slightly slower than uncompressed, but this is to be expected when connecting over localhost since the link speed is virtually unlimited.

My only reservation is the way buffers are allocated in the `BufferedByteReader`.  I think that a solution to #117 may need to employ a solution that leverages `ArrayPool<byte>.Shared.Rent/Return` that could be applied to all of the `IByteHandler` implementations, and cut down on Garbage Collection

My vote is to merge and move ahead with trying to speed things up for TCP in #117

PR #131, opened 2016-11-03
Stored Procedures

### LINKED ISSUE #19: Support for Stored Procedure command type
Current version of MySqlCommand only supports CommandType = Text. Supporting stored procedure commands is a must for any realistic migration from the official MySQL Connector.

ps: nice work so far!

caleblloyd:
- Fixes #19 
- Implements Stored Procedure Cache on a per-connection basis in >= [MySql 5.5.3](http://dev.mysql.com/doc/refman/5.5/en/parameters-table.html)
  - Fixes up Parameter Order, Direction, and Data Types to match the procedure on the Server
- Works without Stored Procedure Cache on < MySql 5.5.3, but caller is responsible for matching c# param Order, Direction, and Data Types to Server params

### DIFF
--- a/src/MySqlConnector/MySqlClient/Caches/CachedParameter.cs
+++ b/src/MySqlConnector/MySqlClient/Caches/CachedParameter.cs
@@ -1,0 +1,40 @@
+﻿using System.Data;
+using System.Linq;
+using MySql.Data.MySqlClient.Types;
+
+namespace MySql.Data.MySqlClient.Caches
+{
+	internal class CachedParameter
+	{
+		public CachedParameter(int ordinalPosition, string mode, string name, string dataType, bool unsigned)
+		{
+			Position = ordinalPosition;
+			if (Position == 0)
+			{
+				Direction = ParameterDirection.ReturnValue;
+			}
+			else
+			{
+				switch (mode.ToLowerInvariant())
+				{
+					case "in":
+						Direction = ParameterDirection.Input;
+						break;
+					case "inout":
+						Direction = ParameterDirection.InputOutput;
+						break;
+					case "out":
+						Direction = ParameterDirection.Output;
+						break;
+				}
+			}
+			Name = name;
+			DbType = TypeMapper.Mapper.GetDbTypeMapping(dataType, unsigned).DbTypes?.First() ?? DbType.Object;
+		}
+
+		internal readonly int Position;
+		internal readonly ParameterDirection Direction;
+		internal readonly string Name;
+		internal readonly DbType DbType;
+	}
+}
--- a/src/MySqlConnector/MySqlClient/Caches/CachedProcedure.cs
+++ b/src/MySqlConnector/MySqlClient/Caches/CachedProcedure.cs
@@ -1,0 +1,100 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Data;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using MySql.Data.Protocol.Serialization;
+
+namespace MySql.Data.MySqlClient.Caches
+{
+	internal class CachedProcedure
+	{
+		internal static async Task<CachedProcedure> FillAsync(IOBehavior ioBehavior, MySqlConnection connection, string schema, string component, CancellationToken cancellationToken)
+		{
+			var cmd = (MySqlCommand) connection.CreateCommand();
+
+			cmd.CommandText = @"SELECT ORDINAL_POSITION, PARAMETER_MODE, PARAMETER_NAME, DATA_TYPE, DTD_IDENTIFIER
+				FROM information_schema.parameters
+				WHERE SPECIFIC_SCHEMA = @schema AND SPECIFIC_NAME = @component
+				ORDER BY ORDINAL_POSITION";
+			cmd.Parameters.Add(new MySqlParameter
+			{
+				ParameterName = "@schema",
+				DbType = DbType.String,
+				Value = schema
+			});
+			cmd.Parameters.Add(new MySqlParameter
+			{
+				ParameterName = "@component",
+				DbType = DbType.String,
+				Value = component
+			});
+
+			var parameters = new List<CachedParameter>();
+			using (var reader = (MySqlDataReader) await cmd.ExecuteReaderAsync(CommandBehavior.Default, ioBehavior, cancellationToken).ConfigureAwait(false))
+			{
+				while (await reader.ReadAsync(cancellationToken).ConfigureAwait(false))
+				{
+					parameters.Add(new CachedParameter(
+						reader.GetInt32(0),
+						!reader.IsDBNull(1) ? reader.GetString(1) : null,
+						!reader.IsDBNull(2) ? reader.GetString(2) : null,
+						reader.GetString(3),
+						reader.GetString(4).IndexOf("unsigned", StringComparison.OrdinalIgnoreCase) != -1
+					));
+				}
+			}
+
+			return new CachedProcedure(schema, component, parameters.AsReadOnly());
+		}
+
+		protected CachedProcedure(string schema, string component, ReadOnlyCollection<CachedParameter> parameters)
+		{
+			m_schema = schema;
+			m_component = component;
+			m_parameters = parameters;
+		}
+
+		internal MySqlParameterCollection AlignParamsWithDb(MySqlParameterCollection parameterCollection)
+		{
+			var alignedParams = new MySqlParameterCollection();
+			var returnParam = parameterCollection.FirstOrDefault(x => x.Direction == ParameterDirection.ReturnValue);
+
+			foreach (var cachedParam in m_parameters)
+			{
+				MySqlParameter alignParam;
+				if (cachedParam.Direction == ParameterDirection.ReturnValue)
+				{
+					if (returnParam == null)
+						throw new InvalidOperationException($"Attempt to call stored function {FullyQualified} without specifying a return parameter");
+					alignParam = returnParam;
+				}
+				else
+				{
+					var index = parameterCollection.NormalizedIndexOf(cachedParam.Name);
+					if (index == -1)
+						throw new ArgumentException($"Parameter '{cachedParam.Name}' not found in the collection.");
+					alignParam = parameterCollection[index];
+				}
+
+				if (alignParam.Direction == default(ParameterDirection))
+					alignParam.Direction = cachedParam.Direction;
+				if (alignParam.DbType == default(DbType))
+					alignParam.DbType = cachedParam.DbType;
+
+				// cached parameters are oredered by ordinal position
+				alignedParams.Add(alignParam);
+			}
+
+			return alignedParams;
+		}
+
+		private string FullyQualified => $"`{m_schema}`.`{m_component}`";
+
+		readonly string m_schema;
+		readonly string m_component;
+		readonly ReadOnlyCollection<CachedParameter> m_parameters;
+	}
+}
--- a/src/MySqlConnector/MySqlClient/CommandExecutors/StoredProcedureCommandExecutor.cs
+++ b/src/MySqlConnector/MySqlClient/CommandExecutors/StoredProcedureCommandExecutor.cs
@@ -1,0 +1,118 @@
+﻿using System.Collections.Generic;
+using System.Data;
+using System.Data.Common;
+using System.Threading;
+using System.Threading.Tasks;
+using MySql.Data.MySqlClient.Types;
+using MySql.Data.Protocol.Serialization;
+
+namespace MySql.Data.MySqlClient.CommandExecutors
+{
+	internal class StoredProcedureCommandExecutor : TextCommandExecutor
+	{
+
+		internal StoredProcedureCommandExecutor(MySqlCommand command)
+			: base(command)
+		{
+			m_command = command;
+		}
+
+		public override async Task<DbDataReader> ExecuteReaderAsync(string commandText, MySqlParameterCollection parameterCollection,
+			CommandBehavior behavior, IOBehavior ioBehavior, CancellationToken cancellationToken)
+		{
+			var cachedProcedure = await m_command.Connection.GetCachedProcedure(ioBehavior, commandText, cancellationToken).ConfigureAwait(false);
+			if (cachedProcedure != null)
+				parameterCollection = cachedProcedure.AlignParamsWithDb(parameterCollection);
+
+			MySqlParameter returnParam = null;
+			m_outParams = new MySqlParameterCollection();
+			m_outParamNames = new List<string>();
+			var inParams = new MySqlParameterCollection();
+			var argParamNames = new List<string>();
+			var inOutSetParams = "";
+			for (var i = 0; i < parameterCollection.Count; i++)
+			{
+				var param = parameterCollection[i];
+				var inName = "@inParam" + i;
+				var outName = "@outParam" + i;
+				switch (param.Direction)
+				{
+					case 0:
+					case ParameterDirection.Input:
+					case ParameterDirection.InputOutput:
+						var inParam = param.WithParameterName(inName);
+						inParams.Add(inParam);
+						if (param.Direction == ParameterDirection.InputOutput)
+						{
+							inOutSetParams += $"SET {outName}={inName}; ";
+							goto case ParameterDirection.Output;
+						}
+						argParamNames.Add(inName);
+						break;
+					case ParameterDirection.Output:
+						m_outParams.Add(param);
+						m_outParamNames.Add(outName);
+						argParamNames.Add(outName);
+						break;
+					case ParameterDirection.ReturnValue:
+						returnParam = param;
+						break;
+				}
+			}
+
+			// if a return param is set, assume it is a funciton.  otherwise, assume stored procedure
+			commandText += "(" + string.Join(", ", argParamNames) +")";
+			if (returnParam == null)
+			{
+				commandText = inOutSetParams + "CALL " + commandText;
+				if (m_outParams.Count > 0)
+				{
+					m_setParamsFlags = true;
+					m_cancellationToken = cancellationToken;
+				}
+			}
+			else
+			{
+				commandText = "SELECT " + commandText;
+			}
+
+			var reader = (MySqlDataReader) await base.ExecuteReaderAsync(commandText, inParams, behavior, ioBehavior, cancellationToken).ConfigureAwait(false);
+			if (returnParam != null && await reader.ReadAsync(ioBehavior, cancellationToken).ConfigureAwait(false))
+				returnParam.Value = reader.GetValue(0);
+			
+			return reader;
+		}
+
+		internal void SetParams()
+		{
+			if (!m_setParamsFlags)
+				return;
+			m_setParamsFlags = false;
+			var commandText = "SELECT " + string.Join(", ", m_outParamNames);
+			using (var reader = (MySqlDataReader) base.ExecuteReaderAsync(commandText, new MySqlParameterCollection(), CommandBehavior.Default, IOBehavior.Synchronous, m_cancellationToken).GetAwaiter().GetResult())
+			{
+				reader.Read();
+				for (var i = 0; i < m_outParams.Count; i++)
+				{
+					var param = m_outParams[i];
+					if (param.DbType != default(DbType))
+					{
+						var dbTypeMapping = TypeMapper.Mapper.GetDbTypeMapping(param.DbType);
+						if (dbTypeMapping != null)
+						{
+							param.Value = dbTypeMapping.DoConversion(reader.GetValue(i));
+							continue;
+						}
+					}
+					param.Value = reader.GetValue(i);
+				}
+			}
+		}
+
+		readonly MySqlCommand m_command;
+		bool m_setParamsFlags;
+		MySqlParameterCollection m_outParams;
+		List<string> m_outParamNames;
+		private CancellationToken m_cancellationToken;
+	}
+}
--- a/src/MySqlConnector/MySqlClient/MySqlCommand.cs
+++ b/src/MySqlConnector/MySqlClient/MySqlCommand.cs
@@ -3,8 +3,8 @@ using System.Data;
 using System.Data.Common;
 using System.Threading;
 using System.Threading.Tasks;
+using MySql.Data.MySqlClient.CommandExecutors;
 using MySql.Data.Protocol.Serialization;
-using MySql.Data.Serialization;
 
 namespace MySql.Data.MySqlClient
 {
@@ -36,6 +36,7 @@ namespace MySql.Data.MySqlClient
 			DbConnection = connection;
 			DbTransaction = transaction;
 			m_parameterCollection = new MySqlParameterCollection();
+			CommandType = CommandType.Text;
 		}
 
 		public new MySqlParameterCollection Parameters
@@ -74,12 +75,20 @@ namespace MySql.Data.MySqlClient
 		{
 			get
 			{
-				return CommandType.Text;
+				return m_commandType;
 			}
 			set
 			{
-				if (value != CommandType.Text)
-					throw new ArgumentException("CommandType must be Text.", nameof(value));
+				if (value != CommandType.Text && value != CommandType.StoredProcedure)
+					throw new ArgumentException("CommandType must be Text or StoredProcedure.", nameof(value));
+				if (value == m_commandType)
+					return;
+
+				m_commandType = value;
+				if (value == CommandType.Text)
+					m_commandExecutor = new TextCommandExecutor(this);
+				else if (value == CommandType.StoredProcedure)
+					m_commandExecutor = new StoredProcedureCommandExecutor(this);
 			}
 		}
 
@@ -111,16 +120,7 @@ namespace MySql.Data.MySqlClient
 
 		internal async Task<int> ExecuteNonQueryAsync(IOBehavior ioBehavior, CancellationToken cancellationToken)
 		{
-			using (var reader = (MySqlDataReader) await ExecuteReaderAsync(CommandBehavior.Default, ioBehavior, cancellationToken).ConfigureAwait(false))
-			{
-				do
-				{
-					while (await reader.ReadAsync(ioBehavior, cancellationToken).ConfigureAwait(false))
-					{
-					}
-				} while (await reader.NextResultAsync(ioBehavior, cancellationToken).ConfigureAwait(false));
-				return reader.RecordsAffected;
-			}
+			return await m_commandExecutor.ExecuteNonQueryAsync(CommandText, m_parameterCollection, ioBehavior, cancellationToken).ConfigureAwait(false);
 		}
 
 		public override Task<object> ExecuteScalarAsync(CancellationToken cancellationToken) =>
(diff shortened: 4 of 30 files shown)


tests/SideBySide.New/StoredProcedureTests.cs
```diff
@@ -0,0 +1,149 @@
...
+		{
+			m_database.Connection.Close();
+		}
+
+		[Fact]
+		public async Task StoredProcedureBasic()
+		{
+			using (var cmd = m_database.Connection.CreateCommand())
+			{
+				cmd.CommandText = @"DROP PROCEDURE IF EXISTS simple;
```
caleblloyd: Failing to create in `MySql.Data`

tests/SideBySide.New/StoredProcedureTests.cs
```diff
@@ -0,0 +1,149 @@
...
+				Assert.Equal("simple", cmd.Parameters["@result"].Value);
+			}
+		}
+
+		[Fact]
+		public async Task StoredProcedureComplex()
+		{
+			using (var cmd = m_database.Connection.CreateCommand())
+			{
+				cmd.CommandText = @"DROP PROCEDURE IF EXISTS complex;
```
caleblloyd: Failing to create in `MySql.Data`

caleblloyd:
I want to separate these methods into an interface:

```csharp
interface ICommandExecutor

async Task<int> ExecuteNonQueryAsync(string commandText, MySqlParameterCollection parameterCollection, IOBehavior ioBehavior, CancellationToken cancellationToken)

async Task<object> ExecuteScalarAsync(string commandText, MySqlParameterCollection parameterCollection, IOBehavior ioBehavior, CancellationToken cancellationToken)

async Task<DbDataReader> ExecuteReaderAsync(string commandText, MySqlParameterCollection parameterCollection, CommandBehavior behavior, IOBehavior ioBehavior, CancellationToken cancellationToken)
```

Then create 2 implementations:

- `TextCommandExecutor`
- `StoredProcedureCommandExecutor`

@bgrainger do you have a recommendation for a namespace I should put these in?

src/MySqlConnector/MySqlClient/MySqlCommand.cs
```diff
@@ -111,7 +114,70 @@ protected override DbParameter CreateDbParameter()
 
 		internal async Task<int> ExecuteNonQueryAsync(IOBehavior ioBehavior, CancellationToken cancellationToken)
 		{
-			using (var reader = (MySqlDataReader) await ExecuteReaderAsync(CommandBehavior.Default, ioBehavior, cancellationToken).ConfigureAwait(false))
+			if (CommandType == CommandType.StoredProcedure)
+				return await ExecuteStoredProcedureAsync(ioBehavior, cancellationToken);
+
+			return await ExecuteTextCommandNonQueryAsync(CommandText, m_parameterCollection, ioBehavior, cancellationToken);
```
bgrainger: Missing `.ConfigureAwait(false)` on both lines.

src/MySqlConnector/MySqlClient/MySqlCommand.cs
```diff
@@ -111,7 +114,70 @@ protected override DbParameter CreateDbParameter()
...
+				}
+
+				if (parameter.Direction == ParameterDirection.ReturnValue)
+					returnParameter = parameter;
+				else
+					argParamNames.Add(parameter.ParameterName);
+			}
+
+			int affectedRows;
+			var commandText = "CALL " + CommandText + "(" + string.Join(", ", argParamNames) +")";
```
bgrainger: Does `argParamNames` need escaping/quoting?

src/MySqlConnector/MySqlClient/MySqlCommand.cs
```diff
@@ -111,7 +114,70 @@ protected override DbParameter CreateDbParameter()
...
+			if (outParams.Count > 0)
+			{
+				commandText = "SELECT " + string.Join(", ", outParamNames);
+				using (var reader = (MySqlDataReader) await ExecuteTextCommandReaderAsync(commandText, new MySqlParameterCollection(), CommandBehavior.Default, ioBehavior, cancellationToken).ConfigureAwait(false))
+				{
+					await reader.ReadAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
+					for (var i = 0; i < outParams.Count; i++)
+					{
+						var parameter = outParams[i];
+						parameter.Value = reader.GetValue(i);
```
bgrainger: Nitpick: unless you're going to use it for something else, I'd probably combine these into one line: `outParams[i].Value = reader.GetValue(i);`.

src/MySqlConnector/MySqlClient/MySqlCommand.cs
```diff
@@ -128,8 +194,16 @@ protected override DbParameter CreateDbParameter()
 
 		internal async Task<object> ExecuteScalarAsync(IOBehavior ioBehavior, CancellationToken cancellationToken)
 		{
+			if (CommandType == CommandType.StoredProcedure)
+				throw new MySqlException("Stored Procedures should be run with ExecuteNonQuery or ExecuteNonQueryAsync");
+
+			return await ExecuteTextCommandScalarAsync(CommandText, m_parameterCollection, ioBehavior, cancellationToken);
```
bgrainger: `.ConfigureAwait(false)`

tests/SideBySide.New/StoredProcedureTests.cs
bgrainger: FWIW, the exception is being thrown from line 45, i.e., `ExecuteNonQueryAsync` on the stored procedure.

bgrainger: It appears that when you use a `ReturnValue` parameter, Connector/Net assumes you are calling a function, not a stored procedure. The following code (inspired by https://dev.mysql.com/doc/connector-net/en/connector-net-tutorials-stored-procedures.html) works with Connector/Net:

```
			cmd.CommandText = "simple";
			cmd.CommandType = CommandType.StoredProcedure;
			{
				Assert.True(await reader.ReadAsync());
				Assert.Equal("simple", reader.GetString(0));
			}
```

tests/SideBySide.New/StoredProcedureTests.cs
```diff
@@ -0,0 +1,149 @@
...
+					DbType = DbType.String,
+					Direction = ParameterDirection.Output,
+				});
+				cmd.Parameters.Add(new MySqlParameter
+				{
+					ParameterName = "@result",
+					DbType = DbType.String,
+					Direction = ParameterDirection.ReturnValue,
+				});
+				await cmd.ExecuteNonQueryAsync();
```
bgrainger: Again, I think you need to `ExecuteReaderAsync` here and read the (single) row that is returned.

src/MySqlConnector/MySqlClient/MySqlCommand.cs
caleblloyd: Thanks, I'll definitely do a global grep for this before I push the final changeset

src/MySqlConnector/MySqlClient/MySqlCommand.cs
caleblloyd: Oh yea.  I had a foreach loop here at first, but it it wasn't detecting `parameter` as a `MySqlParameter` so I was having to use `as MySqlParameter`.  Changed to a for loop and it worked.  I guess `MySqlParameterCollection` should probably implement `IEnumerable<MySqlParameter>`

bgrainger:
The interfaces could go in `MySql.Data.MySqlClient` (I assume they would be `internal`).

src/MySqlConnector/MySqlClient/MySqlCommand.cs
caleblloyd: I thought about this too.  I think that all OUT parameters need to be [User Defined Vars](http://dev.mysql.com/doc/refman/5.7/en/user-variables.html), which all begin with `@`.  I think the caller should escape them per documentation if they want them escaped:

> User variables are written as @var_name, where the variable name var_name consists of alphanumeric characters, ., _, and $. A user variable name can contain other characters if you quote it as a string or identifier (for example, @'my-var', @"my-var", or @`my-var`).

Also, sub-queries are technically valid:

`CALL complex((SELECT 1), (SELECT 2), (SELECT "name"), @a, @b, @c, @d, @e)`

This would mean somebody would have to set their IN `MySqlParameter.ParameterName = "(SELECT 1)"`.  They should probably be working with Text Commands at this point anyways though.

caleblloyd:
They would be `internal`.  Should I create a sub-namespace, say `MySql.Data.MySqlClient.CommandExecutor` that contains all relevant files.  Right now I'm thinking there will be 4 and am worried about cluttering up the `MySql.Data.MySqlClient` ns.  Also it's hard to tell what's related when the first letter puts them far apart in the alphabetized directory listing with a bunch of other unrelated classes.

- `interface ICommandExecutor`
- `abstract BaseCommandExecutor`
- `TextCommandExecutor extends BaseCommandExecutor`
- `StoredProcedureCommandExecutor extends BaseCommandExecutor`

bgrainger:
`MySql.Data.MySqlClient.CommandExecutors` sounds good. ("[Consider using plural namespace names where appropriate."](https://msdn.microsoft.com/en-us/library/ms229026.aspx))

src/MySqlConnector/MySqlClient/CommandExecutors/StoredProcedureCommandExecutor.cs
```diff
@@ -0,0 +1,92 @@
...
+				commandText = "SELECT " + string.Join(", ", outParamNames);
+				using (var reader = await DoExecuteReaderAsync(commandText, new MySqlParameterCollection(), CommandBehavior.Default, ioBehavior, cancellationToken).ConfigureAwait(false))
+				{
+					await reader.ReadAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
+					for (var i = 0; i < outParams.Count; i++)
+						outParams[i].Value = reader.GetValue(i);
+				}
+			}
+
+			return null;
```
caleblloyd: I think `MySql.Data` may actually return a reader here.  How could we return a reader since it is used up when getting the return value?

tests/SideBySide.New/StoredProcedureTests.cs
```diff
@@ -0,0 +1,161 @@
...
+	SELECT 'simple';
+END";
+				await cmd.ExecuteNonQueryAsync();
+			}
+
+			using (var cmd = m_database.Connection.CreateCommand())
+			{
+				cmd.CommandText = "simple";
+				cmd.CommandType = CommandType.StoredProcedure;
+// MySql.Data makes the assumption that Stored Procedures with ParameterDirection.ReturnValue are functions
```
caleblloyd: I don't like the assumption `MySql.Data` makes about Stored Procedure vs. Function based off ParameterDirection.ReturnValue

Could we implement an extension of [System.Data.CommandType](https://msdn.microsoft.com/en-us/library/system.data.commandtype(v=vs.110).aspx) and include a `MySqlCommandType.Function` enum?

caleblloyd: Actually, since enums can't be extended, we could introduce a `StoredProcedureType` enum with options of `Procedure` and `Function`. It would be exposed in `MySqlCommand.StoredProcedureType`

bgrainger: Most stored procedures will return results in their result set (read by `ReadAsync`), even if it's a single row. So MySQL's assumption may not be that strange here. I'd need to do more investigation to be sure it's worth breaking API compatibility in this situation.

caleblloyd:
Latest change set implements stored procedures 1:1 with `MySql.Data`

- Check for a parameter with `ParameterDirection.ReturnValue` and call function, otherwise call stored procedure
- Added a `MySqlCommand.ReaderClosed` function that sets stored procedure `Out` parameter values after a stored procedure's reader is closed.
- Stored procedures returning result sets is now supported

bgrainger:
This code needs more test coverage to verify it's handling all the edge cases correctly (and there are probably a lot). For example, I just wrote some naive code to test stored procedures: 763716410130dbbe15622a84ade7a6e250288506

It succeeds with Connector/Net but fails with this code. (I haven't debugged it yet.)

Additionally, this code seems to be more sensitive to parameter names than Connector/Net; `@factor` works but `factor` doesn't.

caleblloyd:
It looks like `MySql.Data` makes up it's own user variable names:

https://github.com/mysql/mysql-connector-net/blob/6.9/Source/MySql.Data/StoredProcedure.cs#L179

I'll adjust the code to do this, which should eliminate problems stemming from what users choose to name their parameters.

tests/SideBySide.New/StoredProcedureTests.cs
```diff
@@ -0,0 +1,258 @@
...
+				cmd.CommandType = CommandType.StoredProcedure;
+				cmd.Parameters.Add(new MySqlParameter { ParameterName = paramaterName, Value = 3 });
+				using (var reader = await cmd.ExecuteReaderAsync())
+				{
+					Assert.True(await reader.ReadAsync());
+					Assert.Equal("six", reader.GetString(0));
+					Assert.True(await reader.ReadAsync());
+					Assert.Equal("three", reader.GetString(0));
+					Assert.False(await reader.ReadAsync());
+					//Assert.False(await reader.NextResultAsync()); // <-- This passes on MySql.Data but not MySqlConnector, why?
```
caleblloyd: Not sure why this assertion would fail on MySqlConnector.  Is there really another empty result set getting sent by the server?

caleblloyd:
Test coverage now includes all parameter directions (in, out, inout, and return)

`MySql.Data` checks parameter type names match function parameter names in the stored procedure.  It has code to retrieve the stored procedure from the database, parse the parameter names, and enforce that the C# parameter names match.

Our current implementation in this PR just passes positional parameters.  It is up to the caller to name them, choose the right parameter direction, choose the right DbType, and add them to the parameter collection in the correct order.

[This function](https://github.com/mysql/mysql-connector-net/blob/6.9/Source/MySql.Data/StoredProcedure.cs#L114) is where `MySql.Data` performs all of these operations.  Do we want to support all of this?

Current State Pros / Cons:
- Pro: We save a round trip to the Database every time a procedure is called (don't have to get procedure parameters)
- Con: We don't align data types for parameters (this doesn't happen for text SQL queries anyways)
- Con: We don't check to see if all parameters are defined with the proper direction
- Con: We aren't 1:1 backward compatible with `MySql.Data`

caleblloyd:
It's probably best to mirror what `MySql.Data` does. They do cache the procedure parameters for the life of the connection. INFORMATION_SCHEMA only started supporting parameter names after MySQL 5.5.3 so I'll add a version check. Anything lower than that will fallback on using positional parameters.

src/MySqlConnector/MySqlClient/Caches/NormalizedName.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Text.RegularExpressions;
+
+namespace MySql.Data.MySqlClient.Caches
+{
+	internal class NormalizedName
+	{
+		private static readonly Regex NameRe = new Regex(@"^\s*(?:`([^`]+)`|([^\.]+)?)\s*\.?\s*(?:`([^`]+)`|([^\.]+)?)?\s*$",
```
bgrainger: There are valid (but pathological) MySQL identifiers that will break this regex. 

> Identifier quote characters can be included within an identifier if you quote the identifier. If the character to be included within the identifier is the same as that used to quote the identifier itself, then you need to double the character.

http://dev.mysql.com/doc/refman/5.7/en/identifiers.html

May not be worth fixing, since it's such an edge case but FYI (in case you want to make a note or a known issue about it).

src/MySqlConnector/MySqlClient/Caches/Procedure.cs
```diff
@@ -0,0 +1,69 @@
...
+			{
+				ParameterName = "@component",
+				DbType = DbType.String,
+				Value = normalized.Component
+			});
+
+			var parameters = new List<Parameter>();
+			using (var reader = (MySqlDataReader) await cmd.ExecuteReaderAsync(CommandBehavior.Default, ioBehavior, cancellationToken))
+			{
+				while (reader.Read())
```
bgrainger: Do you want `ReadAsync` here?

src/MySqlConnector/MySqlClient/Caches/Procedure.cs
```diff
@@ -0,0 +1,69 @@
...
+			});
+			cmd.Parameters.Add(new MySqlParameter
+			{
+				ParameterName = "@component",
+				DbType = DbType.String,
+				Value = normalized.Component
+			});
+
+			var parameters = new List<Parameter>();
+			using (var reader = (MySqlDataReader) await cmd.ExecuteReaderAsync(CommandBehavior.Default, ioBehavior, cancellationToken))
```
bgrainger: `.ConfigureAwait(false)` :smile:

src/MySqlConnector/MySqlClient/Caches/NormalizedName.cs
caleblloyd: Wow. I feel sorry for anyone working with a database that contains back ticks in the schema name

src/MySqlConnector/MySqlClient/Caches/Procedure.cs
caleblloyd: Yes, thanks for catching

src/MySqlConnector/MySqlClient/MySqlConnection.cs
```diff
@@ -198,6 +200,34 @@ internal MySqlSession Session
...
+			{
+				cachedProcedure = await CachedProcedure.FillAsync(ioBehavior, this, normalized.Schema, normalized.Component, cancellationToken).ConfigureAwait(false);
+				m_cachedProcedures[normalized.CacheKey] = cachedProcedure;
+			}
+			catch (MySqlException ex)
+			{
+				if (ex.Number != 1109)
+					throw;
+				// 1109 means information_schema.parameters table not found on MySql < 5.5.3, see http://dev.mysql.com/doc/refman/5.5/en/parameters-table.html
+				m_cachedProceduresSupported = false;
```
caleblloyd: Opting for an Exception Error Number Check here instead of checking the Server version before the query.  This way we don't have to maintain a list of server types and versions that support `information_schema.parameters` (MySql >= 5.5.3, MariaDB >= whatever, etc).  Is this acceptable?

caleblloyd: Apparently their version numbers are supposed to be [1:1 compatible](https://mariadb.com/kb/en/mariadb/mariadb-vs-mysql-compatibility/), so I'll change this to a version check

src/MySqlConnector/MySqlClient/Caches/CachedParameter.cs
```diff
@@ -0,0 +1,96 @@
...
+					case "INOUT":
+						Direction = ParameterDirection.InputOutput;
+						break;
+					case "OUT":
+						Direction = ParameterDirection.Output;
+						break;
+				}
+			}
+			Name = name;
+			switch (dataType.ToLower())
```
bgrainger: Future enhancement: extract this (and the inverse code in `MySqlDataReader.GetDataTypeName`) to a shared location?

bgrainger: Wondering if we need to implement our own `MySqlType` (or similar) that's the superset of MySQL Server data type names, .NET types, `DbType` enumeration values, `ColumnType` enumeration values, etc., etc.

src/MySqlConnector/MySqlClient/Types/TypeMapper.cs
```diff
@@ -0,0 +1,141 @@
...
+using System.Linq;
+using MySql.Data.Serialization;
+
+namespace MySql.Data.MySqlClient.Types
+{
+	internal class TypeMapper
+	{
+		internal static TypeMapper Mapper = new TypeMapper();
+
+		private TypeMapper()
```
caleblloyd: Here are the beginnings of a Type Mapper.  I think that we can evolve this namespace (in further commits) to take care of most CLR Type / Column Type / DbType conversions.

src/MySqlConnector/MySqlClient/Caches/CachedParameter.cs
```diff
@@ -0,0 +1,40 @@
...
+		public CachedParameter(int ordinalPosition, string mode, string name, string dataType, bool unsigned)
+		{
+			Position = ordinalPosition;
+			if (Position == 0)
+			{
+				Direction = ParameterDirection.ReturnValue;
+			}
+			else
+			{
+				switch (mode.ToLower())
```
bgrainger: Need to use `ToLowerInvariant` to avoid the "[Turkish I problem](http://haacked.com/archive/2012/07/05/turkish-i-problem-and-why-you-should-care.aspx/)".

src/MySqlConnector/MySqlClient/Caches/CachedProcedure.cs
```diff
@@ -0,0 +1,103 @@
...
+			using (var reader = (MySqlDataReader) await cmd.ExecuteReaderAsync(CommandBehavior.Default, ioBehavior, cancellationToken).ConfigureAwait(false))
+			{
+				while (await reader.ReadAsync(cancellationToken).ConfigureAwait(false))
+				{
+					parameters.Add(new CachedParameter(
+						reader.GetInt32(0),
+						!reader.IsDBNull(1) ? reader.GetString(1) : null,
+						!reader.IsDBNull(2) ? reader.GetString(2) : null,
+						reader.GetString(3),
+						reader.GetString(4).ToLower().Contains("unsigned")
```
bgrainger: `reader.GetString(4).IndexOf("unsigned", StringComparison.OrdinalIgnoreCase) != -1`

Avoids culture-sensitive comparison ("Turkish I") and allocating a new string just for substring comparison.

src/MySqlConnector/MySqlClient/Types/TypeMapper.cs
```diff
@@ -0,0 +1,141 @@
...
+		{
+			ColumnTypeMapping columnTypeMapping;
+			if (!m_columnTypeMappingLookup.TryGetValue(ColumnTypeMapping.CreateLookupKey(columnTypeName, unsigned, length), out columnTypeMapping) && length != 0)
+				m_columnTypeMappingLookup.TryGetValue(ColumnTypeMapping.CreateLookupKey(columnTypeName, unsigned, 0), out columnTypeMapping);
+			return columnTypeMapping;
+		}
+
+		private Dictionary<Type, DbTypeMapping> m_dbTypeMappingsByClrType = new Dictionary<Type, DbTypeMapping>();
+		private Dictionary<DbType, DbTypeMapping> m_dbTypeMappingsByDbType = new Dictionary<DbType, DbTypeMapping>();
+		private Dictionary<string, ColumnTypeMapping> m_columnTypeMappingLookup = new Dictionary<string, ColumnTypeMapping>();
```
bgrainger: Could use `StringComparer.OrdinalIgnoreCase` as the key comparer (to match case-insensitively without having to lowercase all input strings).

caleblloyd: Ah I see, good tip.  Since everything runs through CreateLookupKey and this is private, I'll just call `ToLowerInvariant()` there (no need to double-tap)

caleblloyd:
I think this is about ready to merge.  @bgrainger do you see any features or test coverage that should be added?  I am going to squash the commits.

bgrainger:
I think I should have time to go through this in a bit more detail tonight.

src/MySqlConnector/MySqlClient/MySqlParameter.cs
```diff
@@ -153,9 +173,16 @@ internal void AppendSqlString(BinaryWriter writer, StatementPreparerOptions opti
 
 		internal static string NormalizeParameterName(string name)
 		{
-			return string.IsNullOrEmpty(name) ? name :
-				name[0] == '?' || name[0] == '@' ? name.Substring(1) :
-				name;
+			name = name.Trim();
+
+			if ((name.StartsWith("@`") || name.StartsWith("?`")) && name.EndsWith("`"))
```
bgrainger: Use the overloads that take `StringComparison.Ordinal`. The default behaviour is to be culture-sensitive which is horribly inefficient in CoreCLR on Linux: dotnet/coreclr#5612 dotnet/corefx#5027

src/MySqlConnector/MySqlClient/ServerVersions.cs
```diff
@@ -4,6 +4,10 @@ namespace MySql.Data.MySqlClient
 {
 	internal static class ServerVersions
 	{
+		// https://dev.mysql.com/doc/refman/5.7/en/mysql-reset-connection.html
```
bgrainger: 👍 for documentation

tests/MySqlConnector.Tests/TypeMapperTests.cs
```diff
@@ -0,0 +1,109 @@
...
+		[InlineData("multilinestring", false, 0, DbType.Binary)]
+		[InlineData("multipolygon", false, 0, DbType.Binary)]
+		[InlineData("datetime", false, 0, DbType.DateTime)]
+		[InlineData("date", false, 0, DbType.DateTime)] // todo: this should be DbType.Date
+		[InlineData("time", false, 0, DbType.Time)]
+		[InlineData("timestamp", false, 0, DbType.DateTimeOffset)]
+		[InlineData("year", false, 0, DbType.Int32)]
+		public void ColumnTypeMappingTest(string columnTypeName, bool unsigned, int length, DbType dbType)
+		{
+			Assert.Equal(dbType, TypeMapper.Mapper.GetDbTypeMapping(columnTypeName, unsigned, length).DbTypes.FirstOrDefault());
```
bgrainger: Suggestion: add UPPERCASE test values, or a second test with `columnTypeName.ToUpperInvariant()`?

tests/SideBySide.New/StoredProcedureTests.cs
```diff
@@ -0,0 +1,310 @@
...
+			using (var cmd = m_database.Connection.CreateCommand())
+			{
+				cmd.CommandText = "circle";
+				cmd.CommandType = CommandType.StoredProcedure;
+				cmd.Parameters.Add(new MySqlParameter
+				{
+					ParameterName = "@radius",
+					DbType = DbType.Double,
+					Direction = ParameterDirection.Input,
+					Value = (double)1,
```
bgrainger: Suggestion: `Value = 1.0`

tests/MySqlConnector.Tests/TypeMapperTests.cs
caleblloyd: second assertion with ToUpperInvariant() added

src/MySqlConnector/MySqlClient/Caches/CachedProcedure.cs
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		internal MySqlParameterCollection AlignParamsWithDb(MySqlParameterCollection parameterCollection)
+		{
+			var alignedParams = new MySqlParameterCollection();
+
+			MySqlParameter returnParam = null;
+			for (var i = 0; i < parameterCollection.Count; i++)
+				if (parameterCollection[i].Direction == ParameterDirection.ReturnValue)
+					returnParam = parameterCollection[i];
```
caleblloyd: Done.  Had to make `MySqlParameterCollection` implement `IEnumerable<MySqlParameter>`

caleblloyd:
Style fixes are in

tests/SideBySide.New/StoredProcedureTests.cs
bgrainger: There is. https://dev.mysql.com/doc/internals/en/multi-resultset.html says:

> ... and a closing empty resultset, an OK_Packet. ... The trailing OK_Packet is the response to the CALL statement and contains the affected-rows count of the last statement. 

I think the way to address this might be to add extra logic in `NextResultAsync` to detect this empty result set condition (when `CommandType == StoredProcedure`) and return an accurate `false` value.

I can think of one edge case where this will be a problem: someone using Dapper's `QueryMultiple` to call a sproc with multiple result sets. Their code will only read _n_ result sets but Dapper will think there are _n+1_, so will `Cancel` the command because it thinks there was an incomplete read. This will currently throw an exception in this library. (Dapper's regular `Query` for a sproc will work just fine.)

It seems fine to me to ship with this as a known issue and address it in a future version. (Shipping support for stored procedures may turn up other problems too, and we can fix them all at once.)

Issue #135, opened 2016-11-13
NextResult returns true after stored procedure


bgrainger:
`MySqlDataReader.NextResultAsync` will return `true` after reading the results returned by the stored procedure. This appears to be because MySQL Server returns an empty resultset after the main results (https://dev.mysql.com/doc/internals/en/multi-resultset.html): 

> ... and a closing empty resultset, an OK_Packet. … The trailing OK_Packet is the response to the CALL statement and contains the affected-rows count of the last statement.

This will cause a crash with Dapper's QueryMultiple, because Dapper will think that the server returns an extra resultset that the client isn't reading:

```
connection.Execute(@"drop procedure if exists multiple_selects;
create procedure multiple_selects() begin select 1; select 2; end;");

using (var reader = connection.QueryMultiple("multiple_selects", commandType: CommandType.StoredProcedure))
{
	reader.Read<int>().Dump();
	reader.Read<int>().Dump();
} // <-- crash here, disposing the GridReader
```

caleblloyd:
I think we can only handle this behavior when `commandType: CommandType.StoredProcedure`.  If the caller executes a `CommandType.Text` that contains `CALL multiple_selects`, we'd have to do a lot of query parsing to understand where we should remove the empty result set.

bgrainger:
Agreed; it seems that if you're operating at that level, you need to be aware of how MySQL handles the `CALL` statement.

caleblloyd:
I've got some WIP for this that involves creating a `MySqlResultSet` class and buffering `NextResult` for stored procedures.  Should have a PR open soon

bgrainger:
Fixed in [0.7.1](https://www.nuget.org/packages/MySqlConnector/0.7.1).

Issue #117, opened 2016-10-22
Mitigate TCP Retransmissions


caleblloyd:
I've been thinking that performance _should_ be better- I expect it to be more in the 10K RPS range than the 1K RPS range for a fast machine.

The way that we read TCP sockets is slightly out of the norm.  Currently, we only read one MySqlPacket at a time from the socket.  When dealing with result sets, this is equivalent to a single row.  Result set rows are usually pretty small, so this means we are calling into the OS's socket API for every single row in a result set.

We may call socket.read a handful of times before we ever read a single TCP packet.  This means that the TCP packet is sitting in the OS's socket buffer, taking up space.  The socket buffer is unable to accept more packets until it has been completely read.

When activity is high, and many MySQL connections are reading a lot of data, the OS socket buffer is sitting full while connections read from it row by row.  MySQL may try to send more data, but when the socket buffer is full, these packets get dropped and cause a TCP retransmission.  TCP retransmissions are bad for performance.

To test this theory, I ran our stress tests and checked retransmissions after each test:

```
# targets-async2.txt just stresses the GET /api/async endpoint
cat vegeta/targets-async2.txt 
GET http://localhost:5000/api/async

# check retransmissions
 netstat -s | grep retrans
    60320 segments retransmited


# use normal async targets to seed the database
./stress.sh 1000 1s async
Latencies     [mean, 50, 95, 99, max]  2.105138734s, 2.224804101s, 2.905801837s, 3.17405192s, 3.462670324s

# check retransmissions
 netstat -s | grep retrans
    60322 segments retransmited
# caused 2 retransmissions


# use async2 targets to stress 3000 GETs
./stress.sh 3000 1s async2
Latencies     [mean, 50, 95, 99, max]  8.101027972s, 8.310701883s, 12.76829299s, 13.952514362s, 13.981650988s

 netstat -s | grep retrans
    60380 segments retransmited
# caused 78 retransmissions


# use async2 targets to stress 3000 more GETs
./stress.sh 3000 1s async2
Latencies     [mean, 50, 95, 99, max]  5.509762976s, 5.698767446s, 9.626680815s, 9.804948553s, 9.97223801s

 netstat -s | grep retrans
    60794 segments retransmited
# caused 414 retransmissions
```

[...]

PR #136, opened 2016-11-14
Split out MySqlDataReader ResultSet and Row operations into classes

caleblloyd:
- Creates a `MySql.Data.MySqlClient.Results` namespace
  - `ResultSet` class holds Result Set oriented behavior
  - `Row` class holds Row oriented behavior
- Allows for Buffering Result Set or Row in order to look ahead where needed, such as in #135 
- Makes `MySqlDataReader` a lot easier to comprehend :smile: 
- Opening PR for design feedback and suggestions

### DIFF
--- a/src/MySqlConnector/MySqlClient/MySqlDataReader.cs
+++ b/src/MySqlConnector/MySqlClient/MySqlDataReader.cs
@@ -1,11 +1,11 @@
 ﻿using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
-using System.Globalization;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using MySql.Data.MySqlClient.Results;
 using MySql.Data.Protocol.Serialization;
 using MySql.Data.Serialization;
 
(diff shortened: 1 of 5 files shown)


src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -628,205 +165,77 @@ protected override void Dispose(bool disposing)
...
 		{
-			None,
-			ReadResultSetHeader,
-			AlreadyReadFirstRow,
-			ReadingRows,
-			HasMoreData,
-			NoMoreData,
+			get { return _resultSet;  }
+			set { _resultSet = value?.Actiate() ?? new ResultSet(this); }
 		}
```
bgrainger: From an API design perspective, it's somewhat confusing to have `_resultSet` field, `m_ResultSet` property and `GetResultSet` method that all do something slightly different. Also, a setter that transforms `m_ResultSet = null;` into a non-null value smells like a bit of a hack that would be confusing to understand.

Perhaps the property should be removed and its setter become an explicit `SetAndActivateResultSet` method that manipulates the field?

src/MySqlConnector/MySqlClient/Results/ResultSet.cs
```diff
@@ -0,0 +1,358 @@
...
+					m_lastInsertId = -1;
+					State = ResultSetState.ReadResultSetHeader;
+					break;
+				}
+			}
+			m_bufferState = State;
+			return this;
+		}
+
+		internal ResultSet Actiate()
```
bgrainger: Typo: `Actiate`

src/MySqlConnector/MySqlClient/Results/ResultSet.cs
```diff
@@ -0,0 +1,358 @@
...
+	{
+		internal static readonly Task<bool> FalseTask = Task.FromResult(false);
+		internal static readonly Task<bool> TrueTask = Task.FromResult(true);
+
+		public ResultSet(MySqlDataReader dataReader)
+		{
+			DataReader = dataReader;
+		}
+
+		internal async Task<ResultSet> ReadResultSetHeaderAsync(IOBehavior ioBehavior, CancellationToken cancellationToken)
```
bgrainger: Nitpick: I would consider this member to be part of the "public" API of the class (irrespective of the class's `internal` accessibility), following Eric's arguments here: https://ericlippert.com/2014/09/15/internal-or-public/

(Most of the methods on this type I would make `public`.)

bgrainger: I prefer `ValueTask<T>` (fewer allocations and async state machines on the happy path) to `Task` for APIs that aren't required by ADO.NET.

src/MySqlConnector/MySqlClient/Results/ResultSet.cs
```diff
@@ -0,0 +1,358 @@
...
+				State = m_bufferState;
+				return false;
+			}
+			State = ResultSetState.ReadingRows;
+			return true;
+		}
+
+		internal async Task<Row> BufferReadAsync(IOBehavior ioBehavior, CancellationToken cancellationToken)
+		{
+			var read = await DoReadAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
```
bgrainger: `row` would be a better name.

bgrainger:
It seems like this might be doing a lot more allocations than the previous code, but (without measuring) I don't know if it's significant enough to impact performance when reading large resultsets.

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -16,578 +16,105 @@ public sealed class MySqlDataReader : DbDataReader
...
+				m_ResultSet = null;
+				throw;
 			}
-
-			// TODO: Correct exception
-			throw new IndexOutOfRangeException("The column name '{0}' does not exist in the result set.".FormatInvariant(name));
 		}
 
-		public override string GetDataTypeName(int ordinal)
+		internal async Task<ResultSet> DoNextResultAsync(IOBehavior ioBehavior, ResultSet currentResultSet, CancellationToken cancellationToken)
```
bgrainger: `private`

caleblloyd:
Allocations should be limited to a few pointers per result set and a few pointers per row if no buffering is used (the default).  I'm trying to get the buffering working reliably right now.  I'll see if I can make it re-use objects when no buffering is used to cut down on allocations.

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -16,578 +16,105 @@ public sealed class MySqlDataReader : DbDataReader
...
+				return resultSet != null;
 			}
-		}
-
-		public override int GetOrdinal(string name)
-		{
-			VerifyHasResult();
-
-			for (int column = 0; column < m_columnDefinitions.Length; column++)
+			catch (Exception)
```
bgrainger: `Exception` is a broad type to catch here; are there specific exceptions that should be getting handled?

bgrainger:
I ran the following code under VS' .NET Memory Allocation profiler:

```
await connection.OpenAsync();

using (var cmd = new MySqlCommand(@"drop table if exists integers; create table integers(value integer not null primary key); insert into integers(value) values(0), (1), (2), (3), (4), (5), (6), (7), (8), (9);", connection))
	await cmd.ExecuteNonQueryAsync();

long sum = 0;
using (var cmd = new MySqlCommand(@"select i1.value, i2.value, i3.value from integers i1 join integers i2 join integers i3;", connection))
{
		sum += reader.GetInt64(0) + reader.GetInt64(1) + reader.GetInt64(2);
}
Console.WriteLine(sum);
```

Current `master` results:

image

This PR results (highlighting added from manual comparison):

image

There are obviously more `Row` objects being created, but this seems to be more than offset by a reduction in `Func` allocations. (I want to understand that to see if I can remove more unnecessary Func allocations in other places.)

With a switch from `Task` to `ValueTask` in this PR, we might be able to see that number come down under the baseline.

src/MySqlConnector/MySqlClient/Results/Row.cs
```diff
@@ -0,0 +1,394 @@
...
+		{
+			if (ordinal < 0 || ordinal > ResultSet.ColumnDefinitions.Length)
+				throw new ArgumentOutOfRangeException(nameof(ordinal), "value must be between 0 and {0}.".FormatInvariant(ResultSet.ColumnDefinitions.Length));
+
+			if (ResultSet.DataOffsets[ordinal] == -1)
+				return DBNull.Value;
+
+			var data = new ArraySegment<byte>(m_payload, ResultSet.DataOffsets[ordinal], ResultSet.DataLengths[ordinal]);
+			var columnDefinition = ResultSet.ColumnDefinitions[ordinal];
+			var isUnsigned = columnDefinition.ColumnFlags.HasFlag(ColumnFlags.Unsigned);
```
bgrainger: Memory profiler says there is an allocation (!) here: #137.

caleblloyd:
That memory allocation is very useful, is there a good reference on how to get that working?  I have VS Community 2015 installed on my Windows Partition but I couldn't find it.

I'm seeing some other interesting behavior.  While buffering in this PR does not work perfectly yet, it appears to work fine for a single result set for the performance tests.  I've done a performance test with a large result set (return the top 200 rows in the BlogPosts table).  When I buffer the entire ResultSet into memory before reading any of it, I get around 30-50% faster response times.

This is drastically more allocations (it makes heavy use of `Buffer.BlockCopy` into `ArrayPool<byte>.Shared` buffers), yet it is faster.  Maybe .NET incurs a performance penalty for jumping from socket to socket.  In #117 I predicted that it may be due to TCP Retransmissions, but I see the same behavior over Unix Domain Sockets so I'm not sure why this is the case.

bgrainger:
Not sure if it's in Community Edition, but I got to it through:

* create a console app that targets .NET Framework 4.6; set it as Startup Project
* Analyze > Performance Profiler
* Select Performance Wizard; click Start
* Select .NET memory allocation; Next; Next; Finish
* the app will run and close, when it does, the analysis results will be displayed automatically
* select Current View: Allocation at the top to show this list

bgrainger:
As "simple" as that! (I'm hoping there's a quicker way to start a new session because 10+ clicks every time will get tiresome...)

caleblloyd:
I think I've fixed everything in the code review, with the exception of the switch to `ValueTask`.  Most (all?) of the async code here is making or awaiting network requests, which will always result in `Task` and not `Value`.  I also find it difficult to follow the program flow when switching to continuation-style programming as opposed to async/await (I think you posted previously that ValueTask will be allowed to be used with async/await soon - can't wait!)

Allocations to `ResultSet` and `Row` should be way down now (objects are re-used when they aren't buffered).

Buffering is only used in the case of #135, when the final `NextResult` must return `false` in the case of a Stored Procedure.  I've tested it more extensively, and will post a sample of how to buffer all of the result sets in a reader.

One final good code review is needed.  I like most everything here except one part is a little confusing:
- `MySqlDataReader` contains:
  - `m_resultSet` (active result set)
  - `m_resultSetBuffered` (most recent buffered result set)
- `ResultSet` contains:
  - `m_row` (active read)
  - `m_rowBuffered` (most recent buffered read)

The references to the current and most recent buffered objects in each of the classes makes the code a little difficult to read.  I couldn't think of a better way to do it efficiently though, happy to refactor if you can.

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -628,205 +199,71 @@ protected override void Dispose(bool disposing)
...
+		internal MySqlCommand Command { get; private set; }
+		internal MySqlConnection Connection => Command?.Connection;
+		internal MySqlSession Session => Command?.Connection.Session;
 
 		internal static async Task<MySqlDataReader> CreateAsync(MySqlCommand command, CommandBehavior behavior, IOBehavior ioBehavior, CancellationToken cancellationToken)
 		{
 			var dataReader = new MySqlDataReader(command, behavior);
-			await dataReader.ReadResultSetHeaderAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
+			await dataReader.ReadFirstResultSetAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
 			return dataReader;
```
caleblloyd: In order to buffer the entire result set, add `while (await BufferNextResultAsync(ioBehavior, cancellationToken).ConfigureAwait(false) != null){}` before the return

src/MySqlConnector/MySqlClient/Results/ResultSet.cs
```diff
@@ -0,0 +1,373 @@
...
+					payload = await Session.ReceiveReplyAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
+					EofPayload.Create(payload);
+
+					LastInsertId = -1;
+					State = ResultSetState.ReadResultSetHeader;
+					break;
+				}
+			}
+			BufferState = State;
+			//await BufferEntireAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
```
bgrainger: Did you mean to leave this in?

src/MySqlConnector/MySqlClient/Results/ResultSet.cs
```diff
@@ -0,0 +1,373 @@
...
+		private async Task<Row> DoReadAsync(IOBehavior ioBehavior, Row row, CancellationToken cancellationToken)
+		{
+			// if we've already read past the end of this resultset, Read returns false
+			if (BufferState == ResultSetState.HasMoreData || BufferState == ResultSetState.NoMoreData || BufferState == ResultSetState.None)
+				return null;
+
+			var payloadTask = Session.ReceiveReplyAsync(ioBehavior, cancellationToken);
+			if (payloadTask.IsCompletedSuccessfully)
+				row = ReadAsyncRemainder(row, payloadTask.Result);
+			row = await ReadAsyncAwaited(row, payloadTask.AsTask()).ConfigureAwait(false);
```
bgrainger: This looks like a bug to assign `row` twice if the condition on line 131 is true.

src/MySqlConnector/MySqlClient/Results/ResultSet.cs
```diff
@@ -0,0 +1,373 @@
...
+			return row;
+		}
+
+		private async Task<Row> ReadAsyncAwaited(Row row, Task<PayloadData> payloadTask)
+		{
+			var payload = await payloadTask.ConfigureAwait(false);
+			return ReadAsyncRemainder(row, payload);
+		}
+
+		private Row ReadAsyncRemainder(Row row, PayloadData payload)
```
bgrainger: It would be more idiomatic C# to return void and take a `ref Row row` parameter; this more typically implies the variable would be reused.

src/MySqlConnector/MySqlClient/Results/ResultSet.cs
```diff
@@ -0,0 +1,373 @@
...
+				return false;
+			}
+			State = ResultSetState.ReadingRows;
+			return true;
+		}
+
+		public async Task<Row> BufferReadAsync(IOBehavior ioBehavior, CancellationToken cancellationToken)
+		{
+			m_rowBuffered?.BufferData();
+			m_rowBuffered = await DoReadAsync(ioBehavior, null, cancellationToken).ConfigureAwait(false);
```
bgrainger: Won't `m_rowBuffered` already be set by `DoReadAsync`? Unless it returns `null`, in which case `null` will be enqueued in `m_readBuffer`? Could that cause problems later?

It's confusing, and potentially error-prone, to have multiple functions responsible for setting `m_rowBuffered`. (Still trying to ensure I understand the logic properly before suggesting an alternative.)

src/MySqlConnector/MySqlClient/Results/ResultSet.cs
bgrainger: Oh, right, I forgot about that limitation.

src/MySqlConnector/MySqlClient/Results/ResultSet.cs
caleblloyd: Yes, this is not the correct behavior.  I think I can change this whole method to return a ValueTask

caleblloyd:
I've made the following updates:

- Changed name of `DoNextResultSetAsync` to `ScanResultSetAsync` in `MySqlDataReader.cs`
  - It is now returns a `ValueTask`
  - It is solely responsible for setting `m_resultSetBuffered`

- Changed name of `DoReadAsync` to `ScanRowAsync` in `ResultSet.cs`
  - It now returns a `ValueTask`
  - It is solely responsible for setting `m_rowBuffered`

Reason for the name change it to reflect that the Result Set / Row is being "scanned" into the object that is being passed.  I've seen similar terminology in Golang's database/sql package.

I ran performance tests and saw similar numbers.  I don't think this swings performance much in one way or another, but it does give us the ability to buffer result sets / rows now.

PR #143, opened 2016-12-06
Bump Dependencies to 4.3.0

caleblloyd:
The .NET Dependencies have been updated to 4.3.0 across the board, and we should upgrade!

I've also changed the `System.Threading` dependency to `System.Threading.Tasks`.  The former is not compatible with UWP but the latter is.

### DIFF
--- a/.ci/config/config.json
+++ b/.ci/config/config.json
@@ -1,0 +1,8 @@
+{
+    "Data": {
+      "ConnectionString": "server=127.0.0.1;user id=mysqltest;password='test;key=\"val';port=3306;database=mysqltest;Use Affected Rows=true",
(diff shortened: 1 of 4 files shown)


caleblloyd:
I re-ran TravisCI three times and keep getting this error:

    Error starting userland proxy: listen tcp 0.0.0.0:3306: bind: address already in use

Not sure if they updated their image to run MySQL by default or if we are just pinned to a jacked up VM.  I'll give it some time then re-run the build.  If we keep hitting we may need to either switch base images or find another port to run MySQL on.

caleblloyd:
Travis CI fixes caused this to get marked as merged.  Moved this PR to #144

PR #144, opened 2016-12-07
Bump Dependencies to 4.3.0

caleblloyd:
The .NET Dependencies have been updated to 4.3.0 across the board, and we should upgrade!

I've also changed the `System.Threading` dependency to `System.Threading.Tasks`.  The former is not compatible with UWP but the latter is.

### DIFF
--- a/src/MySqlConnector/project.json
+++ b/src/MySqlConnector/project.json
@@ -25,29 +25,29 @@
 				"System.Transactions": "4.0.0.0"
 			},
 			"dependencies": {
-				"System.Buffers": "4.0.0",
-				"System.Runtime.InteropServices.RuntimeInformation": "4.0.0",
-				"System.Threading.Tasks.Extensions": "4.0.0"
+				"System.Buffers": "4.3.0",
+				"System.Runtime.InteropServices.RuntimeInformation": "4.3.0",
+				"System.Threading.Tasks.Extensions": "4.3.0"
 			}
 		},
 		"netstandard1.3": {
 			"frameworkAssemblies": {
 			},
 			"dependencies": {
-				"System.Buffers": "4.0.0",
-				"System.Collections.Concurrent": "4.0.12",
-				"System.Data.Common": "4.1.0",
-				"System.IO": "4.1.0",
-				"System.IO.Compression": "4.1.0",
-				"System.IO.FileSystem": "4.0.1",
-				"System.Linq": "4.1.0",
-				"System.Net.NameResolution": "4.0.0",
-				"System.Net.Security": "4.0.0",
-				"System.Net.Sockets": "4.1.0",
-				"System.Runtime.Extensions": "4.1.0",
-				"System.Runtime.InteropServices.RuntimeInformation": "4.0.0",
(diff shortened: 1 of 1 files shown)


bgrainger:
I don't know what the right philosophy is here. It's certainly nice to test against the latest packages, and ensure that there are no accidental breaking changes. However, does upgrading our dependencies force clients to update from (say) System.Buffers 4.0.0 to 4.3.0 unnecessarily (since this library will work fine with the older version)?

bgrainger:
> I've also changed the `System.Threading` dependency to `System.Threading.Tasks`

Does that require 4.3, or can we fix UWP without that? (Update, we can: #147.)

caleblloyd:
I'm of the opinion that using the latest stable dependencies available is the best idea.  They may contain underlying performance or errata fixes.

If library implementors always update their .NET Core to the latest stable version, these dependencies will match their .NET Core dependencies.  For example, if an app uses .NET Core 1.1, it uses version `4.3.0` of all of the `System.` libraries already, so `dotnet restore` will avoid requiring that they install an older version of the `System.` as long as we are also on `4.3.0`.

Conversely, if library implementors stay on .NET Core 1.0 then they would pull less dependencies if we don't update...  With Software, however, I think it's always better to update as long as BC is maintained.

We can't have UWP support until we figure out a way to get `System.Net.Security` or an alternate package to work with UWP.  This is the package that provides `SslStream` which is pretty important :)  I'm wondering if this is the reason that the only Entity Framework [sample for UWP](https://docs.microsoft.com/en-us/ef/core/get-started/uwp/getting-started) is SQLITE.

I can only think of a couple practical reasons for talking directly to an SQL Server from UWP in the first place:

1. You are building a Database Browser (similar to MySqlWorkbench)
2. Some internal application to your company where you are not worried about people trying to hack your database (since the connection string could be easily decompiled out of the binary)

bgrainger:
> I'm of the opinion that using the latest stable dependencies available is the best idea. They may contain underlying performance or errata fixes.

I agree, but I don't think that's *our* decision. 

We should be compatible with them, but we don't need to force any consumers of MySqlConnector to update past the minimum package version that they need for their app. For all we know, they may have extensive compliance testing for each set of dependencies that are shipped and the timeframe of updating to .NET Core 1.1 should be in their hands, not forced because they update MySqlConnector to get a small bugfix.

What if the library uses the minimum version (for compatibility with consumers), but the SideBySide tests use the latest dependencies? That should test that MySqlConnector does work with the latest platform/libraries.

caleblloyd:
> For all we know, they may have extensive compliance testing...

Good point

> What if the library uses the minimum version (for compatibility with consumers), but the SideBySide tests use the latest dependencies?

Ah I think I misunderstood how NuGet works.  Now I see on our [nuget page](https://www.nuget.org/packages/MySqlConnector/) that our `project.json` dependencies are the _minimum_ library version required.  I mistakenly thought that dependencies were pinned to that exact version.

So if apps are running .NET core 1.1, they should default to using the `4.3.0` even if we specify `4.0.0` in the project.json.  That's all I was after.

PR #145, opened 2016-12-07
Implement Keepalive

### LINKED ISSUE #132: Implement Keepalive Connection String Options
`Keep Alive, Keepalive` (default: 0) For TCP connections, idle connection time measured in seconds, before the first keepalive packet is sent. A value of 0 indicates that keepalive is not used.

Should call `socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, value)` on TCP Socket if `value > 0`

caleblloyd:
- Fixes #132 
- A value of 0 indicates that the OS Default keepalive settings are used
- On Windows, a value greater than 0 is the idle connection time, measured in seconds, before the first keepalive packet is sent.
- Unix-based Operating Systems will always use the OS Default keepalive settings.

### DIFF
--- a/docs/content/connection-options.md
+++ b/docs/content/connection-options.md
@@ -148,6 +148,13 @@ These are the other options that MySqlConnector supports.  They are set to sensi
     <td>True to have MySqlDataReader.GetValue() and MySqlDataReader.GetDateTime() return DateTime.MinValue for date or datetime columns that have disallowed values.</td>
   </tr>
   <tr>
+    <td>Keep Alive, Keepalive</td>
+    <td>0</td>
+    <td>TCP Keepalive idle time.  A value of 0 indicates that the OS Default keepalive settings are used.
+    On Windows, a value greater than 0 is the idle connection time, measured in seconds, before the first keepalive packet is sent.
+    Due to limitations in .NET Core, Unix-based Operating Systems will always use the OS Default keepalive settings.</td>
+  </tr>
+  <tr>
     <td>Old Guids, OldGuids</td>
     <td>false</td>
     <td> The backend representation of a GUID type was changed from BINARY(16) to CHAR(36). This was done to allow developers to use the server function UUID() to populate a GUID table - UUID() generates a 36-character string. Developers of older applications can add 'Old Guids=true' to the connection string to use a GUID of data type BINARY(16).</td>
--- a/src/MySqlConnector/Serialization/SerializationUtility.cs
+++ b/src/MySqlConnector/Serialization/SerializationUtility.cs
@@ -1,3 +1,7 @@
+using System;
+using System.Net.Sockets;
+using System.Runtime.InteropServices;
+
 namespace MySql.Data.Serialization
 {
 	internal static class SerializationUtility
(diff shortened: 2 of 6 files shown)


caleblloyd:
I've included a basic test, but I couldn't think of a better way to unit test this since everything happens at the network driver layer.  Here's a screenshot of a packet capture running this commit with `Keepalive=3` in the connection string.

keepalive

As you can see, the keepalive runs every 3 seconds from time 792-810.  At time 812, I shutdown the MySql server, and the server sends FIN.

The client does not send FIN, so the client believes the connection is still open.  The next keepalive packet is not ACK'd, so the client retries keepalives every 1s, which is the proper interval.  Eventually 10 keepalives are missed, so the client RST the connection.

I think that this demonstrates correct behavior.

tests/SideBySide.New/ConnectAsync.cs
```diff
@@ -101,6 +103,21 @@ public ConnectAsync(DatabaseFixture database)
...
+		{
+			var csb = AppConfig.CreateConnectionStringBuilder();
+			csb.Keepalive = 1;
+			using (var connection = new MySqlConnection(csb.ConnectionString))
+			{
+				await connection.OpenAsync();
+				var stopwatch = Stopwatch.StartNew();
+				await Task.Delay(3000);
+				stopwatch.Stop();
+				Assert.InRange(stopwatch.ElapsedMilliseconds, 2500, 3500);
```
bgrainger: What is this testing specifically? That no exception is thrown?

docs/content/connection-options.md
```diff
@@ -148,6 +148,13 @@ These are the other options that MySqlConnector supports.  They are set to sensi
     <td>True to have MySqlDataReader.GetValue() and MySqlDataReader.GetDateTime() return DateTime.MinValue for date or datetime columns that have disallowed values.</td>
   </tr>
   <tr>
+    <td>Keep Alive, Keepalive</td>
+    <td>0</td>
+    <td>TCP Keepalive idle time.  A value of 0 indicates that the OS Default keepalive settings are used.
```
bgrainger: Connector/Net documents "A value of 0 indicates that keepalive is not used."; do we want to be different by default?

src/MySqlConnector/Serialization/SerializationUtility.cs
```diff
@@ -47,5 +51,30 @@ public static int GetBytesPerCharacter(CharacterSet characterSet)
 				return 1;
 			}
 		}
+
+		public static void SetKeepalive(Socket socket, uint keepAliveTimeSeconds)
+		{
+			// Always use the OS Default Keepalive settings
+			socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, true);
```
bgrainger: Following up on question above, should this option only be set if user has set keepAlive != 0?

If you think that the benefits of turning it on by default are worthwhile, then I have no objection. (I'm in favour of changing to use sensible defaults.)

tests/SideBySide.New/ConnectAsync.cs
caleblloyd: Correct.  I guess the stopwatch is not necessary, I'll just remove that

src/MySqlConnector/Serialization/SerializationUtility.cs
caleblloyd: I think that always enabling it to use the OS Default would be sensible.  It allows the OS to clean up old sockets should the connection go away.  If data is consistently being sent across the network, Keepalive packets will never need to be sent.

[Windows](https://blogs.technet.microsoft.com/nettracer/2010/06/03/things-that-you-may-want-to-know-about-tcp-keepalives/) and [Linux](http://www.tldp.org/HOWTO/html_single/TCP-Keepalive-HOWTO/#configuringkernel) both default to send the first Keepalive packet after 2 hours of inactivity on the socket.

PR #147, opened 2016-12-07
Remove unnecessary netstandard1.3 dependencies.

bgrainger:
Various dependencies have accumulated during development, and it appears that not all of them are still necessary.


bgrainger:
Shipped in [0.7.4](https://www.nuget.org/packages/MySqlConnector/0.7.4).

PR #155, opened 2016-12-29
Change Minimum Version from .NET 4.5 to .NET 4.5.1

### LINKED ISSUE #154: Not compatible with .net 4.5.1
@caleblloyd, The following depedencies refered to `System.Runtime` and the `Guid` in mscorlib will conflict with the `Guid` based in `System.Runtime`

image

image

I don't know why Full .NET Framework lib need refer to the .net standard lib.

caleblloyd:
- Fixes #154 
- Change Minimum Version from .NET 4.5 to .NET 4.5.1
- Changes framework assemblies to type=build

### DIFF
--- a/tests/SideBySide.New/QueryTests.cs
+++ b/tests/SideBySide.New/QueryTests.cs
@@ -129,19 +129,16 @@ create table query_invalid_sql(id integer not null primary key auto_increment);"
 					Assert.Throws<MySqlException>(() => cmd2.ExecuteReader());
 					Assert.Throws<MySqlException>(() => cmd2.ExecuteScalar());
 
-#if NET45
+#if NET451
 					reader1.Close();
-					using (var reader2 = cmd2.ExecuteReader())
+#else
+					reader1.Dispose();
+#endif
+					using (cmd2.ExecuteReader())
 					{
 					}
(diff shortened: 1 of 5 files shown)


tests/SideBySide.New/QueryTests.cs
```diff
@@ -128,17 +128,9 @@ public void NextResultBeforeRead()
 
 					Assert.Throws<MySqlException>(() => cmd2.ExecuteReader());
 					Assert.Throws<MySqlException>(() => cmd2.ExecuteScalar());
-
-#if NET45
```
bgrainger: Can this be a `#if !NETSTANDARD1_3` (or `#if NET451`?), too? (I assume the `.Close()` method is still available there (and that it's worth testing that it fully closes an open reader).

tests/SideBySide.New/QueryTests.cs
```diff
@@ -129,19 +129,16 @@ public void NextResultBeforeRead()
 					Assert.Throws<MySqlException>(() => cmd2.ExecuteReader());
 					Assert.Throws<MySqlException>(() => cmd2.ExecuteScalar());
 
-#if NET45
+#if NET451
```
caleblloyd: Put this back and made it read a little clearer for the .NET Core case

PR #158, opened 2017-01-11
Implement SslMode=preferred

### LINKED ISSUE #157: Support SSL without Client Certificates and add SslMode=preferred
When I initially implemented SSL in #88, I erroneously assumed that Client Certificates were required.  This is not the case.  SSL can be initiated without client certificates.

This clears up my confusion about `SslMode=Preferred`.  That option is meant to connect to servers that have SSL enabled by default with no extra client configuration.  MySqlConnector should check the initial handshake packet when `SslMode=Preferred` and attempt to connect via SSL if the `SSL` flag is set.

caleblloyd:
- Fixes #157 
- Adds `SslMode=preferred` as default functionality for SslMode (this now matches Oracle's connector)
- Allows for SSL connection without client certs
- Adds test cases for:
  - `SslMode=preferred`
  - Good client cert (signed by same CA as server's `ssl-ca`)
  - Bad client cert (signed by different CA than server's `ssl-ca`)


src/MySqlConnector/MySqlClient/MySqlSslMode.cs
```diff
@@ -6,7 +6,12 @@
 	public enum MySqlSslMode
 	{
 		/// <summary>
-		/// Do not use SSL. This is the default.
+		/// Use SSL if the server supports it. This is the default.
+		/// </summary>
+		Preferred,
```
bgrainger: We should keep the same order as https://github.com/mysql/mysql-connector-net/blob/5864e6b21a8b32f5154b53d1610278abb3cb1cee/Source/MySql.Data/MysqlDefs.cs#L363 so that `default(MySqlSslMode)` is the same (source compatibility). (Even though that's not the same default that the connector would actually use.)

src/MySqlConnector/Serialization/MySqlSession.cs
```diff
@@ -87,8 +87,13 @@ public MySqlSession(ConnectionPool pool, int poolGeneration)
 			if (cs.UseCompression && (initialHandshake.ProtocolCapabilities & ProtocolCapabilities.Compress) == 0)
 				cs = cs.WithUseCompression(false);
 
-			if (cs.SslMode != MySqlSslMode.None)
+			var serverSupporsSsl = (initialHandshake.ProtocolCapabilities & ProtocolCapabilities.Ssl) != 0;
```
bgrainger: Typo: `Supports`

caleblloyd:
> We should keep the same order as https://github.com/mysql/mysql-connector-net/blob/5864e6b21a8b32f5154b53d1610278abb3cb1cee/Source/MySql.Data/MysqlDefs.cs#L363 

I changed the order to match.  I left out the `Prefered=Preferred` alias.  I guess this would help with misspelling, but it also makes `MySqlSslMode.Prefered` the first auto-suggested ENUM in my IDE, which is going to promote misspelling.

I don't think that Aliases count as an integer value for enumerations, so the integer values should still match up without it.

caleblloyd:
I ran `SideBySide.Baseline` tests with SSL.  MySql.Data has an issue reading client certs from our default path (maybe the `.` in `.ci` is causing an issue for them?), but once I move them out to a different path the Baseline tests pass.

bgrainger:
Shipped in [0.11.0](https://www.nuget.org/packages/MySqlConnector/0.11.0).

PR #167, opened 2017-01-25
Buffer Socket Reads

### LINKED ISSUE #164: Improve async performance
I have a (non-public) test case that executes two bulk read queries (in a batch); one that returns 31K rows of `int,string,int,int` and a second that returns 37K rows of `int,int,int,int`. The remote server is MySQL 5.6 on Linux, about 10ms away from my test box.

On full .NET 4.6.2 on Windows 10, the sync codepath uses less than one core and completes in 4s; the async code path uses over 300% CPU and takes 20s wall time.

```
sync:
real   4.043s
total  1.922s
user   1.438s
sys    0.484s

async:
real   20.840s
total  68.422s
user   47.531s
sys    20.891s
```

With dotnet, the results are even worse:

```
sync:
real   4.336s
total  1.953s
user   1.469s
sys    0.484s

async:
real   24.468s
total  85.703s
user   59.219s
sys    26.484s
```

Some investigation with dotTrace and dotPeek on the full .NET Framework shows that `Socket.ReceiveAsync` (used in the async code path) takes 2s of CPU time while `Socket.Receive` (which blocks until the data is received) only takes 1.5s. Moreover, `Socket.ReceiveAsync` appears to only return `false` (i.e., receive completed synchronously) on failure; a successful result *always* completes asynchronously, meaning that we suffer the penalty of a callback being queued to the threadpool on top of the extra CPU time just to initiate the operation!

Additionally, network bandwidth (shown in Task Manager) is noticeably lower for the async case than the sync case.

We may need to find some way to get the benefits of asynchronous I/O without the overhead of `Socket.ReceiveAsync`, perhaps by #163 or by writing custom P/Invoke for the full .NET Framework on Windows.

caleblloyd:
- Fixes #164 
- Uses `BufferedByteReader` to read up to 16384 bytes out of the OS Socket Buffer at once
- Changes `IByteHandler.ReadAsync` method to accept an `ArraySegment<byte> buffer` parameter that new data is read into

### DIFF
--- a/src/MySqlConnector/Protocol/Serialization/SocketByteHandler.cs
+++ b/src/MySqlConnector/Protocol/Serialization/SocketByteHandler.cs
@@ -11,20 +11,18 @@ namespace MySql.Data.Protocol.Serialization
 			m_socket = socket;
 			var socketEventArgs = new SocketAsyncEventArgs();
 			m_socketAwaitable = new SocketAwaitable(socketEventArgs);
-			m_buffer = new byte[16384];
 		}
 
-		public ValueTask<ArraySegment<byte>> ReadBytesAsync(int count, IOBehavior ioBehavior)
+		public ValueTask<int> ReadBytesAsync(ArraySegment<byte> buffer, IOBehavior ioBehavior)
 		{
-			var buffer = count < m_buffer.Length ? m_buffer : new byte[count];
-			if (ioBehavior == IOBehavior.Asynchronous && m_socket.Available < count)
+			if (ioBehavior == IOBehavior.Asynchronous)
 			{
-				return new ValueTask<ArraySegment<byte>>(DoReadBytesAsync(buffer, 0, count));
+				return new ValueTask<int>(DoReadBytesAsync(buffer));
 			}
 			else
 			{
-				var bytesRead = m_socket.Receive(buffer, 0, count, SocketFlags.None);
-				return new ValueTask<ArraySegment<byte>>(new ArraySegment<byte>(buffer, 0, bytesRead));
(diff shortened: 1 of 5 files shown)


src/MySqlConnector/Protocol/Serialization/SocketByteHandler.cs
```diff
@@ -41,30 +39,21 @@ public ValueTask<int> WriteBytesAsync(ArraySegment<byte> data, IOBehavior ioBeha
...
-			if (payload.Count <= m_buffer.Length)
-			{
-				Buffer.BlockCopy(payload.Array, payload.Offset, m_buffer, 0, payload.Count);
-				m_socketAwaitable.EventArgs.SetBuffer(m_buffer, 0, payload.Count);
-			}
-			else
-			{
-				m_socketAwaitable.EventArgs.SetBuffer(payload.Array, payload.Offset, payload.Count);
-			}
+			m_socketAwaitable.EventArgs.SetBuffer(payload.Array, payload.Offset, payload.Count);
```
bgrainger: The previous code was intentional: d705e209dd34f21cab4eeb4b036427adf37c3d11

I wonder if keeping `m_buffer` in this new code and copying in/out of it would be an improvement or not.

bgrainger:
This PR brings the timings for my test harness to:

```
.NET Framework sync
real   3.426s
total  1.156s
user   1.156s
sys    0.000s

.NET Framework async
real   3.056s
total  1.438s
user   1.344s
sys    0.094s

.NET Core sync
real   3.564s
total  1.063s
user   1.031s
sys    0.031s

.NET Core async
real   2.769s
total  1.359s
user   1.250s
sys    0.109s
```

A substantial improvement! :+1:

PR #171, opened 2017-01-30
Fixes #170 - Refactoring checks for Windows platform to work with AWS/Lambda.

### LINKED ISSUE #170: Checks for OS fail on AWS/Lambda
When executed on within AWS/Lambda, the following line throws a `PlatformNotSupportedException`

```csharp
RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
```

This function is used by `ConnectionSettings` and a few other classes to modify their behavior by platform (such as attempting to use unix sockets vs tcp).  While I have an open support issue with the Lambda team, I don't believe they will resolve this anytime soon.

SebastianC:
Fixes #170


bgrainger:
Thanks for the bug report and the PR!

PR #173, opened 2017-02-03
Adding support for DateTimeOffset in MySqlParameter

### LINKED ISSUE #172: DateTimeOffset is not handled in MySqlParameter
It throws on https://github.com/mysql-net/MySqlConnector/blob/master/src/MySqlConnector/MySqlClient/MySqlParameter.cs#L169

sebastienros:
Fixes #172


sebastienros:
The "baseline" set of tests fails on the new test. Not sure what this the differences are between the two.

caleblloyd:
Baseline tests are the same test cases, except run against the MySQL driver from Oracle

Oracle probably does not support DateTimeOffset, you'd need to disable them for baseline using a conditional compilation flag `#if BASELINE`

sebastienros:
Updated

caleblloyd:
Storing `DateTimeOffset` as a `DATETIME` in MySql in this manner strips the Time Zone information from the object.  Let's discuss in #172

sebastienros:
Updated.

I based the change on the fact that DateTimeOffset is mapped to a timestamp column type. Otherwise we would need to assume all DateTime are parsed as UTC and I didn't want to take this decision.

bgrainger:
@caleblloyd Do you still want the "do not merge" label on this, or have your objections been satisfied?

caleblloyd:
It doesn't look like it supports reading into a DateTimeOffset.

So did we decide that forcing conversation to UTC for DateTimeOffset is OK? Or did we want to throw an exception if the DateTimeOffset was not in UTC before saving?

caleblloyd:
I can implement reading into a DateTimeOffset in a separate PR once this is merged though, let me do a code review and remove the label

sebastienros:
If you want to do it as part of this PR, just edit this PR to target a new branch and continue to work on it.

caleblloyd:
OK, I can do it that way. @sebastienros can you squash your commits on your branch and then I'll pick it up and add the reader portion?

sebastienros:
Tip: Use the new Github features to squash it when you merge, it's just one click:

First you edit this PR to target a new branch:

image

Then you select "Squash and Merge" in Github

image

It will squash all the commits and rebase on top of the target branch.
Sorry if you are already aware of this.

bgrainger:
I was planning to squash and merge when it came time.

caleblloyd:
Changes look good to me, let's go ahead and squash/merge this and I will open another issue for the Reader portion.

caleblloyd:
@sebastienros thanks for the tip on Squash in GitHub PRs, I hadn't seen that before

PR #174, opened 2017-02-07
Improve ExecuteAsync Checks and Readability

caleblloyd:
I was reading back over `TextCommandExecutor.cs` and found that things could be simplified.

I also moved the `VerifyValid` back to being a private call in `MySqlCommand.cs` so that it is run no matter which Command Executor interface is used.


caleblloyd:
Going to go ahead and merge since all test cases including `MultipleReaders` still pass

PR #177, opened 2017-02-08
Reader Functions for DateTimeOffset

### LINKED ISSUE #175: Support Reading into DateTimeOffset
#172 adds support for using `DateTimeOffset` in parameters.

MySqlDataReader should also support reading into a `DateTimeOffset` object.

`DateTimeOffset` objects should always be set to UTC timezone when read.

caleblloyd:
- Fixes #175 
- Adds public `MySqlDataReader.GetDateTimeOffset` function
- Adds overrides for `MySqlDataReader.GetFieldValue<T>` `MySqlDataReader.GetFieldValueAsync<T>`
  - Needed to convert `DateTimeOffset` if `MySqlDataReader.GetFieldValue<typeof(DateTimeOffset)>` is called
  - May be useful in the future
- Added test path for `MySqlDataReader.GetFieldValue<T>` and `MySqlDataReader.GetFieldValueAsync<T>`

### DIFF
--- a/src/MySqlConnector/MySqlClient/MySqlDataReader.cs
+++ b/src/MySqlConnector/MySqlClient/MySqlDataReader.cs
@@ -169,6 +169,8 @@ namespace MySql.Data.MySqlClient
 
 		public override DateTime GetDateTime(int ordinal) => GetResultSet().GetCurrentRow().GetDateTime(ordinal);
 
+		public DateTimeOffset GetDateTimeOffset(int ordinal) => GetResultSet().GetCurrentRow().GetDateTimeOffset(ordinal);
+
 		public override string GetString(int ordinal) => GetResultSet().GetCurrentRow().GetString(ordinal);
 
 		public override decimal GetDecimal(int ordinal) => GetResultSet().GetCurrentRow().GetDecimal(ordinal);
@@ -190,6 +192,14 @@ namespace MySql.Data.MySqlClient
 			DoClose();
 		}
 #endif
+
+		public override T GetFieldValue<T>(int ordinal)
+		{
+			if (typeof(T) == typeof(DateTimeOffset))
+				return (T) Convert.ChangeType(GetDateTimeOffset(ordinal), typeof(T));
+
+			return base.GetFieldValue<T>(ordinal);
+		}
 
 		protected override void Dispose(bool disposing)
 		{
--- a/tests/SideBySide.New/DataTypes.cs
+++ b/tests/SideBySide.New/DataTypes.cs
@@ -2,6 +2,7 @@
 using System.Data.Common;
 using System.Globalization;
 using System.Linq;
+using System.Reflection;
 using System.Threading.Tasks;
 using Dapper;
 using MySql.Data.MySqlClient;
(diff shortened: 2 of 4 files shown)


src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -191,6 +193,22 @@ public override void Close()
 		}
 #endif
 
+		public override T GetFieldValue<T>(int ordinal)
+		{
+			if (typeof(T) == typeof(DateTimeOffset))
+				return (T) Convert.ChangeType(GetDateTimeOffset(ordinal), typeof(T));
```
bgrainger: Isn't `(T) Convert.ChangeType` redundant here because we know `T` is `DateTimeOffset`?

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -191,6 +193,22 @@ public override void Close()
...
+
+			return base.GetFieldValue<T>(ordinal);
+		}
+
+		public override async Task<T> GetFieldValueAsync<T>(int ordinal, CancellationToken cancellationToken)
+		{
+			if (typeof(T) == typeof(DateTimeOffset))
+				return (T) Convert.ChangeType(GetDateTimeOffset(ordinal), typeof(T));
```
bgrainger: Consider using `Task.FromResult` here to avoid an unnecessary `await` below. (Stephen Cleary has some thoughts on that [here](http://blog.stephencleary.com/2016/12/eliding-async-await.html) but I think it's worth it in this case since `GetFieldValueAsync` ought to be low-overhead.)

tests/SideBySide.New/DataTypes.cs
```diff
@@ -632,9 +660,33 @@ private void DoQuery<TException>(string table, string column, string dataTypeNam
...
-							Assert.Equal(value.GetType(), reader.GetFieldType(0));
+
+							// test `reader.GetValue` and `reader.GetFieldType` if value matches default type
+							if (matchesDefaultType)
+							{
+								Assert.Equal(value, reader.GetValue(0));
+								Assert.Equal(value.GetType(), reader.GetFieldType(0));
+							}
+
+							// test `reader.GetFieldValue<value.GetType()>`
```
bgrainger: :+1:

tests/SideBySide.New/DataTypes.cs
```diff
@@ -632,9 +660,33 @@ private void DoQuery<TException>(string table, string column, string dataTypeNam
...
+							var syncMethod = typeof(MySqlDataReader)
+								.GetMethod("GetFieldValue")
+								.MakeGenericMethod(value.GetType());
+							Assert.Equal(value, syncMethod.Invoke(reader, new object[]{ 0 }));
+
+							// test `reader.GetFieldValueAsync<value.GetType()>`
```
bgrainger: :+1:

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
caleblloyd: I tried without it first but got an error that I wasn't returning type `T`

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
caleblloyd: It turns out that the [base method already does this](https://github.com/dotnet/corefx/blob/release/1.0.0/src/System.Data.Common/src/System/Data/Common/DbDataReader.cs/#L199)!  The override can be eliminated, although I will keep the test in place to ensure if the base method ever changes we are still in good shape.

caleblloyd:
Code review changes complete

bgrainger:
Didn't mean to squash-and-merge these changes; GitHub just saves the last "merge" action you took as the default.

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
sebastienros: I think it would work with `(T)(object)GetDateTimeOffset(ordinal)`

Issue #169, opened 2017-01-28
Switch authentication command fails with Azure MySQL


bgrainger:
Forked from PomeloFoundation/Pomelo.EntityFrameworkCore.MySql#181.

When connecting to MySQL on Azure, even if the selected MySQL version is 5.7, the server announces its version as `5.5.0.0`. This causes the [`ServerVersions.SupportsResetConnection` test](https://github.com/mysql-net/MySqlConnector/blob/db8179046f652e18292037553adf6321f549549b/src/MySqlConnector/Serialization/MySqlSession.cs#L113) to be `false` because MySQL Server 5.5 didn't support the "reset connection" command. The code then tries to reset the connection by switching users. Azure MySQL sends an authentication method switch request packet. We send a response, which is rejected by the server.

The same failure to log in can be reproduced with `MySql.Data` by setting the `ConnectionReset=true` connection string setting (it's false by default). This seems to indicate it may be a problem with Azure MySQL, and not with the client.

One workaround would be to detect server version `5.5.0.0` as >= 5.7.3. This may be safe because MySQL never uses four-part version numbers (that's a Microsoft kind of thing to do), so we should never see that number when connecting to a real MySQL Server 5.5 instance. Additionally, the first GA version of MySQL Server 5.5 was 5.5.8 according to [the release notes](https://dev.mysql.com/doc/relnotes/mysql/5.5/en/). Where this might fail is if the user is actually running MySQL Server 5.5 on Azure; presumably it would also announce its version as `5.5.0.0`.

bgrainger:
Now that Azure Database for MySQL is available globally, I've been able to reproduce this issue and [report it in the MSDN forums](https://social.msdn.microsoft.com/Forums/en-US/adb35274-2f20-4968-8026-49a8d7ade1f0/comchangeuser-command-not-supported?forum=AzureDatabaseforMySQL).

bgrainger:
If we change the auth method in [`ChangeUserPayload.Create`](https://github.com/mysql-net/MySqlConnector/blob/master/src/MySqlConnector/Serialization/ChangeUserPayload.cs#L18), e.g., `unsupported_auth`, I would expect Azure Database for MySQL to respond with an authentication method switch request payload, because `unsupported_auth` is not an auth plugin name that it recognises. Instead, it still replies directly with an error packet with `Access denied for user`.

Kagamine:
@bgrainger could you have a try now, we have a new deploy yesterday.

[...]

Issue #178, opened 2017-02-15
Reset session when returning to pool


MrSmoke:
I currently have a use case were I get far better performance setting `ConnectionReset` to `false` but I still want the connection reset for safety purposes. 

I was thinking that when `ReturnToPool` is called the connection could be reset there (if the enabled) instead of when it's pulled out of the pool, leaving the reset to be done in the background (before being added back into the pool).

It would however mean that the pool is easier to deplete.

bgrainger:
It seems like this should work; I've created #179 to investigate it.

In my current PR, the connection is reset synchronously when `MySqlConnection.Dispose` is called; since the caller is still blocked, it's possible that this is not enough to increase performance.

The final solution may need to queue the work of resetting the connection (or disposing it) to a threadpool thread so that the client is freed up immediately when closing the connection. Presumably in that implementation, any exceptions would simply be swallowed and just cause the connection to be dropped from the pool. The downside would be that any errors would be silently ignored and only manifest themselves as a performance problem (because the pool would never contain live connections).

This would also be a behavioural change from the official connector, although I'm not sure it's one that would affect anyone. However, I did have to [turn ConnectionReset off](https://github.com/mysql-net/MySqlConnector/pull/179/files#diff-ec9bcc1180d65ef50e9c683747878812R140) for a test because it was causing a session variable to be reset immediately, rather than having the desired side-effect.

@caleblloyd Any thoughts on the suggestion or proposed implementation?

caleblloyd:
@MrSmoke what is the Round-Trip time from your client to your server (ping your MySql Host from wherever your client code resides)

MrSmoke:
Around 1-2ms.

caleblloyd:
How big is the speedup when you turn off connection reset?  2-4ms or is it higher?  I'm wondering if what you're seeing is environment specific.

bgrainger:
Another benefit of resetting the connection (in the background) as soon as it's returned to the pool is that it would immediately free up any session-specific server variables (e.g., temp tables). Right now, I'm assuming that those persist in memory on the server until the connection is pulled from the pool and reset.

caleblloyd:
> Right now, I'm assuming that those persist in memory on the server until the connection is pulled from the pool and reset.

Or until the connection it's reaped

I can see the benefits, my main concern is forcing the reset commands to run over Sync I/O for everyone. The concurrency tests have shown that even 1ms of synchronous execution can cause a huge difference in throughput.

[...]

PR #179, opened 2017-02-16
Reset connection on return to pool

bgrainger:
Issue #178 suggests that the connection be reset when it's returned to the pool, so that it's ready immediately when a new `MySqlConnection` is opened (and reuses a pooled connection).

In the current PR, the connection is reset synchronously when the session is closed; this may not be sufficient to increase performance since the caller will still be blocked (albeit at the end of the connection lifetime, not at the beginning). It may be necessary to offload the work of resetting the connection to a background thread.

### DIFF
--- a/src/MySqlConnector/MySqlClient/ConnectionPool.cs
+++ b/src/MySqlConnector/MySqlClient/ConnectionPool.cs
@@ -61,7 +61,7 @@ namespace MySql.Data.MySqlClient
 				if (session.PoolGeneration == m_generation)
 					m_sessions.Enqueue(session);
 				else
-					session.DisposeAsync(IOBehavior.Synchronous, CancellationToken.None).ConfigureAwait(false);
+					session.DisposeAsync(IOBehavior.Synchronous, CancellationToken.None).GetAwaiter().GetResult();
 			}
 			finally
 			{
--- a/src/MySqlConnector/MySqlClient/ConnectionPool.cs
+++ b/src/MySqlConnector/MySqlClient/ConnectionPool.cs
@@ -33,11 +33,6 @@ namespace MySql.Data.MySqlClient
 					}
 					else
 					{
-						// session is valid, reset if supported
-						if (m_connectionSettings.ConnectionReset)
-						{
-							await session.ResetConnectionAsync(m_connectionSettings, ioBehavior, cancellationToken).ConfigureAwait(false);
-						}
 						// pooled session is ready to be used; return it
 						return session;
 					}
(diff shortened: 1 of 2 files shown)


caleblloyd:
This is an interesting proposition. I'll do some review and performance testing on it tomorrow.

caleblloyd:
Use of Synchronous I/O in this method is correct.  Switching the Connection Reset to Async causes lock-ups when using Sync I/O.

I think that awaiting the result of the Connection Reset should be deferred until the connection is drawn from the pool again.  Since connections are inserted at the back of the Queue, this should give it a nice amount of time to complete before it is drawn again.

While we're at it - the `TryPingAsync` command really only needs to be run after a certain amount of idle time (say 30 seconds).  If the connection has been active in a short period of time, this is pretty much a wasted round trip.

I pushed a branch with a commit on top of this PR that has this behavior: https://github.com/mysql-net/MySqlConnector/commit/518c72cacc4beaf0adafbb7073cc7a0e29c00b5e you should be able to cherry-pick that commit into this PR if it looks good.

caleblloyd:
I think this will unintentionally fix #169 because the exception in `ResetConnectionAsync` will get caught and the MySqlSession will be disposed.  It will mask that exception every single time, I'm not sure if that's a good thing...

bgrainger:
As I wrote on #178:

> The downside would be that any errors would be silently ignored and only manifest themselves as a performance problem (because the pool would never contain live connections).

I also don't currently have a good idea for handling this.

Were you able to do any performance testing on this yet, to see if the change is worthwhile?

caleblloyd:
> I also don't currently have a good idea for handling this.

If we want to mirror the current behavior, we could still initiate the reset task upon returning, but check for an exception upon re-drawing from the pool.  It would essentially be [removing this try-catch](https://github.com/mysql-net/MySqlConnector/commit/5313c564239b22c60eb0474439f01ec9112c560e#diff-709a9503470224e906bdecb8736e16c4R122)

> Were you able to do any performance testing on this yet, to see if the change is worthwhile?

The main benefit here would be for clients which exhibit behaviors:

(A) Connection pool is not constantly in use
(B) There is a high Round Trip Time (RTT) between Client and Server

Looking at our current traffic for drawing from the connection pool, we have 3 round trips:

image

RT 1 is from `PingSessionAsync`.  This takes RTT + 0.07ms
RT 2 is from actual resetting of the connection.  This takes RTT + 0.06ms
RT 3 is from `SET NAMES utf8mb4`.  This takes RTT + 0.06ms

If the connection pool is constantly in use, there will be no difference whether connection is reset at the beginning or end because those events will occur back-to-back.  If there is a negligible RTT (e.g. localhost like my timings here), then all of this will  happen in a fraction of a millisecond.  But, if (A) and (B) are both true, then we will move 2 of the 3 Round Trips when leasing a connection to a time when the connection would normally be idle, and that will benefit clients with a high RTT.

One thing to consider is that by implementing this, we would now  be forcing these 2 round trips over Sync I/O, which will block a thread pool thread.  I don't think we force Sync I/O on any other happy paths in the codebase currently.

Another consideration might be to run `ResetConnectionAsync` in the place of `PingSessionAsync`, and catch a MySqlException in the case that the connection was closed by the server.  This would eliminate one round trip.

caleblloyd:
As a follow-up to my most recent comment.

> One thing to consider is that by implementing this, we would now be forcing these 2 round trips over Sync I/O, which will block a thread pool thread. I don't think we force Sync I/O on any other happy paths in the codebase currently.

After reflecting on this a little more, I think that staying 100% async on the happy path should be the primary concern of our library.

> Another option is to run `ResetConnectionAsync` in the place of `PingSessionAsync`, and catch a MySqlException in the case that the connection was closed by the server.  This would eliminate one round trip.

This is still an option but would mask errors in `ResetConnectionAsync` by catching the exception.  It would have to be done by parsing exception messages.  If everyone else agrees, my vote is to abandon this PR and open this change as an "up for grabs" issue.

bgrainger:
Closing in favour of #264.

PR #156, opened 2017-01-03
Updated pull request for MySqlBulkLoader

gitsno:
This is an update to my original pull request #148, all the requested changes have been made.

Sean

PR #148, opened 2016-12-10
MySqlBulkLoader implementation

gitsno:
This pull request adds support for the LOCAL_INFILE_Request protocol and a complete implementation of the MySqlBulkLoader class that is available in the standard Connector/NET. It includes sync and async tests that use the LOAD DATA INFILE statement with a MySqlCommand as well as tests that use the MySqlBulkLoader class. The test file and tables include INT, VARCHAR, DATETIME and BLOB data types.

The MySqlBulkLoader class includes a new property - InfileStream - that is not available in the standard Connector/NET. InfileStream allows loading data directly from a Stream that follows the same formatting as a regular file loaded with LOAD DATA INFILE. Tests for the InfileStream are also implemented.

PR #180, opened 2017-02-17
Add MySqlBulkLoader

### LINKED ISSUE #15: Implement MySqlBulkLoader
Clone the `MySqlBulkLoader` API (specified here: https://dev.mysql.com/doc/connector-net/en/connector-net-programming-bulk-loader.html).

As an extension, support a `Stream` property instead of a `FileName`. Behind the scenes, create a GUID and pass that to MySQL as the `LOAD DATA LOCAL INFILE` file name (https://dev.mysql.com/doc/refman/5.7/en/load-data.html).

When the [`LOCAL_INFILE_Request`](https://dev.mysql.com/doc/internals/en/com-query-response.html#packet-Protocol::LOCAL_INFILE_Request) is received with that GUID as the name, stream the in-memory data to the server.

bgrainger:
This PR builds on top of #156 (which was a modification of #148) to get the tests working and fix bugs in the implementation.

Fixes #15.


bgrainger:
The Baseline tests fail under Appveyor when this new code is added. I haven't been able to consistently reproduce the failure locally yet.

src/MySqlConnector/MySqlClient/MySqlBulkLoader.cs
```diff
@@ -0,0 +1,210 @@
...
+                throw new InvalidOperationException("Cannot set both FileName and SourceStream");
+
+            if (string.IsNullOrWhiteSpace(FileName) && SourceStream != null)
+            {
+                if (!Local)
+                    throw new InvalidOperationException("Cannot use SourceStream when Local is not true.");
+
+                FileName = StreamPrefix + Guid.NewGuid().ToString("N");
+                lock (s_lock)
+                    s_streams.Add(FileName, SourceStream);
```
caleblloyd: Lock is fine, but could use `ConcurrentDictionary` instead

tests/SideBySide.New/BulkLoaderAsync.cs
```diff
@@ -0,0 +1,377 @@
...
+			string testClient;
+#if BASELINE
+			testClient = "Baseline";
+#else
+			testClient = "New";
+#endif
+			m_testTable = "test.BulkLoaderAsyncTest" + testClient;
+
+			var initializeTable = @"
+				create schema if not exists test;
```
caleblloyd: Right now we only use the Schema from the connection string in tests.  This is so that folks can do local testing against an existing MySQL database and not have random schemas appear/disappear when tests are run.  It also eliminates the need for the MySQL user to have SCHEMA level privileges.  Could this test be adapted to use the schema in the connection string like the others?

tests/SideBySide.New/BulkLoaderSync.cs
```diff
@@ -0,0 +1,381 @@
...
+			string testClient;
+#if BASELINE
+			testClient = "Baseline";
+#else
+			testClient = "New";
+#endif
+			m_testTable = "test.BulkLoaderSyncTest" + testClient;
+
+			var initializeTable = @"
+				create schema if not exists test;
```
caleblloyd: use connection string schema

tests/SideBySide.New/LoadDataInfileAsync.cs
```diff
@@ -0,0 +1,68 @@
...
+			string testClient;
+#if BASELINE
+			testClient = "Baseline";
+#else
+			testClient = "New";
+#endif
+			m_testTable = "test.LoadDataInfileAsyncTest" + testClient;
+
+			var initializeTable = @"
+				create schema if not exists test;
```
caleblloyd: use connection string schema

tests/SideBySide.New/LoadDataInfileSync.cs
```diff
@@ -0,0 +1,67 @@
...
+			string testClient;
+#if BASELINE
+			testClient = "Baseline";
+#else
+			testClient = "New";
+#endif
+			m_testTable = "test.LoadDataInfileSyncTest" + testClient;
+
+			var initializeTable = @"
+				create schema if not exists test;
```
caleblloyd: use connection string schema

src/MySqlConnector/MySqlClient/Results/ResultSet.cs
```diff
@@ -46,9 +48,31 @@ public ResultSet(MySqlDataReader dataReader)
...
 				{
-					throw new NotSupportedException("Don't support LOCAL_INFILE_Request");
+					try
+					{
+						var localInfile = LocalInfilePayload.Create(payload);
+						using (var stream = localInfile.FileName.StartsWith(MySqlBulkLoader.StreamPrefix, StringComparison.Ordinal) ?
+							MySqlBulkLoader.GetAndRemoveStream(localInfile.FileName) :
+							File.OpenRead(localInfile.FileName))
+						{
+							byte[] readBuffer = new byte[65536];
```
caleblloyd: could use ArrayBuffer.Shared<byte>.Rent/Return.  Since it's only one allocation though it won't make a big difference

caleblloyd: 65536 seems like a logical size though.  Not too much RAM but should be enough to keep the socket buffer busy.

tests/SideBySide.New/BulkLoaderAsync.cs
bgrainger: Yes, I forgot about that. I'll adjust all these.

src/MySqlConnector/MySqlClient/MySqlBulkLoader.cs
bgrainger: Seems overkill for something that shouldn't typically have any contention.

src/MySqlConnector/MySqlClient/Results/ResultSet.cs
bgrainger: And (at least on desktop .NET Framework) it doesn't go to the LOH (which is >85k).

bgrainger: I'd be interested in profiling with different sizes, but I'm not sure this is a commonly-used type/feature.

caleblloyd: I agree, bulk loading is usually a one-off.  my write socket buffer default size on Ubuntu 16.04 is 16384 so this should give it plenty of time to work before looping again

bgrainger:
I noticed data corruption in the `LoadDataInfile` tests; will investigate.

PR #181, opened 2017-02-17
Unify how config.json is found and used by tests

bgrainger:
* Use the same `config.json` file for (no compression, no SSL) Appveyor and Travis tests
* Find the same file at runtime in Baseline tests, instead of copying it to the `bin` folder
* Clean up namespaces in code


bgrainger:
These commits were cherry-picked from #180 into their own logical PR.

PR #185, opened 2017-02-26
Add test for MySqlCommand.LastInsertedId

caleblloyd:
- Add explicit test for `MySqlCommand.LastInsertedId`


caleblloyd:
Appveyor failure but I don't think it's due to the new test.

@bgrainger can you add me on AppVeyor?  I think there's a setting where you can add contributors to the `mysql-net` organization as "users" in AppVeyor so that we can re-queue builds and such.

PR #188, opened 2017-03-09
Switch to MSBuild

caleblloyd:
- Switch project to MSBuild
- Switch appveyor to install .NET Core using [korebuild](https://github.com/aspnet/KoreBuild) so that tooling 1.0 works
- Merge `SideBySide.New` and `SideBySide.Baseline` tests into one project called `SideBySide`.  The `netcoreapp1.1.1` runs the New tests and the `net462` runs the Baseline tests.


caleblloyd:
Merged `SideBySide` tests are working on AppVeyor.  I don't like that `netcoreapp1.1.1` runs new tests and `net462` runs baseline tests though.  Ideally we would be able to run tests against `MySqlConnector` on both `netcoreapp1.1.1` and `net462`

I could not get specifying dependencies based off of configuration does working with `dotnet restore`.  It can only restore one configuration at a time.  Technically it does work, but we would have to run something like `dotnet restore /configuration=Baseline` before executing the baseline tests.  I wonder if this could be achieved with MSBuild Targets?

I believe what I have implemented right now is giving us the same coverage as before though...

caleblloyd:
`msbuild` generates a bunch of warnings that look like this:

```
warning : Detected package downgrade: System.Collections.Concurrent from 4.3.0 to 4.0.12  [/home/travis/build/mysql-net/MySqlConnector/MySqlConnector.sln]
warning : Detected package downgrade: System.IO from 4.3.0 to 4.1.0  [/home/travis/build/mysql-net/MySqlConnector/MySqlConnector.sln]
...
```

I think we may need to do something different with dependency versioning in `msbuild`.  Yesterday I published a small class library to NuGet: [Pomelo.JsonObject](https://www.nuget.org/packages/Pomelo.JsonObject/), and once I converted it to msbuild, `dotnet publish` automatically added `NETStandard.Library (>= 1.6.1)` as a dependency, which in-turn imports version `4.3.0` of many of the system libraries that we require.

caleblloyd:
> I don't like that netcoreapp1.1.1 runs new tests and net462 runs baseline tests though. Ideally we would be able to run tests against MySqlConnector on both netcoreapp1.1.1 and net462

I think the best path forward is just to require an extra `dotnet restore` when switching between New and Baseline tests.  This allows us to run New tests against .NET Core and full .NET Framework.  Appveyor now tests MySqlConnector against `netcoreapp1.1.1` and `net462`.  I've made appropriate updates in this PR, here's how to use it:

To run the New tests against MySqlConnector:

`dotnet restore; dotnet test`

To run the Baseline tests against MySql.Data:

`dotnet restore /p:Configuration=Baseline; dotnet test -c Baseline`

src/MySqlConnector/MySqlConnector.csproj
```diff
@@ -0,0 +1,57 @@
...
+    <PackageReference Include="System.Buffers" Version="4.0.0" />
+    <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.0.0" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.0.0" />
+    <Reference Include="System" />
+    <Reference Include="System.Data" />
+    <Reference Include="System.Transactions" />
+    <Reference Include="Microsoft.CSharp" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
```
iamcarbon: You can add : <PackageReference Remove="NETStandard.Library" /> 

to remove the implicit reference to .NETStandard 1.6.1 made by the SDK.

bgrainger:
Thanks for doing this! I'm not completely caught up on all the .NET Core 1.0 tooling changes, but this looks good to me. 

I'll merge now and you can try @iamcarbon's suggestion later; doesn't seem like a blocking issue.

PR #198, opened 2017-03-20
dispose reader when disposing connection. fixes #196

### LINKED ISSUE #196: MySql Connector throw an exception with a big concurrency volume
image

caleblloyd:
- Change `bool HasActiveReader` to `MySqlDataReader ActiveReader`
- Dispose active reader if one exists when connection is closed


caleblloyd:
Added a test in #199.  When I run it prior to this getting merged, there's a "Packet received out-of-order" because there's still buffered data in the reader.  Test proves that this fix is working.

PR #200, opened 2017-03-20
Test failure to dispose MySqlDataReader.

bgrainger:
This tests changes made for #196.

I wasn't able to repro the exact failure in that bug report, but did consistently get the following exception:

```
Error Message:
 System.InvalidOperationException : Packet received out-of-order. Expected 1; got 5.
Stack Trace:
   at SideBySide.QueryTests.UseReaderWithoutDisposing() in MySqlConnector\tests\SideBySide\QueryTests.cs:line 468
```

The exception was fixed when tested with the changes in #198.

### DIFF
--- a/tests/SideBySide/QueryTests.cs
+++ b/tests/SideBySide/QueryTests.cs
@@ -1,6 +1,8 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Data;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


caleblloyd:
Haha we wrote pretty much the same test case.  Mine doesn't use threads so yours may be more of a "stress" test that could catch random behavior.  See #199

bgrainger:
:smile: We should probably keep them both; I'll merge them together.

PR #209, opened 2017-03-27
Move version history to the docs website.

bgrainger:
Change `<PackageReleaseNotes>` to just link to the online documentation (instead of needing to be updated with every release).


caleblloyd:
Good idea - looks good!

Issue #211, opened 2017-03-30
What's the connection lifetime in the pool? / When are idle connections released?


nvivo:
Hi,

I'm investigating why I'm getting more open connections with this driver than with the oracle one, and I got stuck on this question: after a connection is returned to the pool, how long does it stay idle before released?

I noticed that the original driver has the `Connection Lifetime` option, the doc says: 
> When a connection is returned to the pool, its creation time is compared with the current time, and the connection is destroyed if that time span (in seconds) exceeds the value specified by Connection Lifetime. This is useful in clustered configurations to force load balancing between a running server and a server just brought online. A value of zero (0) causes pooled connections to have the maximum connection timeout.

How does it work with this driver?

bgrainger:
`Connection Lifetime` is not currently supported; see #105.

Thus, the lifetime will be controlled by the server, which should close them after [`wait_timeout`](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_wait_timeout) seconds (which defaults to `28800` = eight hours).

bgrainger:
Opened #212 to implement this connection string option.

nvivo:
That explains it. I have a cluster, and after a peak of queries during startup, each node is retaining 100+ connections, exhausting the server after some nodes go up. I took a look at the `wait_timeout`, but I got a little worried about TIME_WAITs like described [here](http://serverfault.com/a/355938/229225), and since I'm using RDS, I'm not sure how Amazon will handle this.

Right now I have a deadline to put a get a new cluster running and I'd rather not go back to the old driver, so let me ask this:

Is it possible to control this manually with ClearPool? Should there be a problem is I call ClearPoolAsync() every 10 minutes while having around 30 connections being used? 

And is there any way to get the number of open sessions in the pool without reflection?

bgrainger:
> Should there be a problem is I call ClearPoolAsync() every 10 minutes while having around 30 connections being used?

This should not be a problem at all; it will iterate over all idle connections in the pool (or all pools, if you call `MySqlConnector.ClearAllPools`) and close them. Connections that are in use will not be affected.

> And is there any way to get the number of open sessions in the pool without reflection?

[...]

Issue #85, opened 2016-10-04
Implement MinimumPoolSize


bgrainger:
The connection pool should respect the `MySqlConnectionStringBuilder.MinimumPoolSize` property and open that many connections when the first connection is made.

bgrainger:
Creating initial connections to meet `MinimumPoolSize` was shipped in [0.31.0](https://github.com/mysql-net/MySqlConnector/releases/tag/0.31.0).

PR #213, opened 2017-03-30
Implement TrimPool

nvivo:
This is an idea to work around Connection Lifetime that should have a similar effect. Related to #211.

This patch transforms `ClearPool` into `TrimPool`, and makes `ClearPool` = TrimPool(conectionsToKeep: 0);

Then it adds a static method `MySqlConnection.TrimPool(int connectionsToKeep)` that should release only the required number of connections to keep the pool at some sane size without releasing more than required.

Ideally, it would be better to trim connections based on lifetime, but I'm not confortable changing this codebase yet to add this. If it was possible to add some method like "TrimPool(TimeSpan)" that would trim based on idle time, I could easily implement the timer on the app side calling it every 3 minutes or so.

### DIFF
--- a/src/MySqlConnector/MySqlClient/ConnectionPool.cs
+++ b/src/MySqlConnector/MySqlClient/ConnectionPool.cs
@@ -69,13 +69,16 @@ namespace MySql.Data.MySqlClient
 			}
 		}
 
-		public async Task ClearAsync(IOBehavior ioBehavior, CancellationToken cancellationToken)
+		public async Task TrimAsync(IOBehavior ioBehavior, int connectionsToKeep, CancellationToken cancellationToken)
 		{
+			if (m_sessions.Count <= connectionsToKeep)
+				return;
+
 			// increment the generation of the connection pool
 			Interlocked.Increment(ref m_generation);
 
 			var waitTimeout = TimeSpan.FromMilliseconds(10);
-			while (true)
+			while (m_sessions.Count <= connectionsToKeep)
 			{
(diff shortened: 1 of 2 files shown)


bgrainger:
Would having `MinimumPoolSize` #85 implemented solve your problem? Then you could set `MinimumPoolSize=20; MaximumPoolSize={your peak usage}` and calling `ClearPool` would remove all but 20 idle connections.

And if the [watchdog thread](https://github.com/mysql-net/MySqlConnector/issues/211#issuecomment-290528075) were implemented, it seems like no manual workarounds would be necessary; after a few minutes, idle connections would be cleaned up automatically.

nvivo:
@bgrainger Yes, this is quite a dirty workaround, I wouldn't leave it there if there are better options. As I just commented [here](https://github.com/mysql-net/MySqlConnector/issues/211#issuecomment-290539552), after some extra thinking, it seems Maximum Pool Size is a better workaround, so I'll close this.

But as for a solution, only a watchdog would really work. See, ideally, I don't want to set a minimum number of connections to leave open, neither a maximum. If my node is idle, it should have zero connections to the database. If it's at peak, I'd like to pool as many as needed. During these peak times, I may reuse those 100's connections for a while, so I don't really want to lose the pool. Once those connections are not used anymore for 30 seconds or so, they can be released.

It seems that having a miminum pool size would work to keep latency at minimum on a webserver, but that's not my case. these are background services.

So, for now, a Max pool size would be my workaround, it will cause a lot of connections to keep opening and closing, but I don't have any latency requirements on these servers, I don't mind paying this price now.

PR #216, opened 2017-03-31
Throw MySqlException for oversize packets.

bgrainger:
When a MySQL Server receives a packet larger than `max_allowed_packet` it will reply with an error payload ("Got a packet bigger than 'max_allowed_packet' bytes") and then reset the connection.

This may manifest on the client side as an error payload with that message, or a `SocketException` (when not using SSL) or an `IOException` (when using SSL).

The connector now handles those exceptions and throws a more descriptive exception if it's likely that the underlying problem is exceeding `max_allowed_packet`.

Note that one unhandled problem is that (when this happens) the socket has been closed (and `MySqlSession` is in the failed state) but `MySqlConnection.State` will still be `Open`; reusing that connection will throw an exception from `MySqlSession`. The workaround is to close/dispose the connection and return the session to the pool (which will detect that it's broken and dispose it).


bgrainger:
Test failed with `System.InvalidOperationException : Packet received out-of-order. Expected 9; got 7.`

I haven't been able to reproduce this locally, but I'm guessing it's due to the `Got a packet bigger than 'max_allowed_packet' bytes` being received before the socket is reset. MySQL Server probably sends that as soon as it sees the first packet which takes the cumulative size > `max_allowed_packet` (recall that payloads are broken into 16MiB packets). If so, it would reply with `current-sequence-number + 1` while the connector keeps sending more packets to complete the payload. We don't expect this out-of-order response from the server.

Issue #67, opened 2016-09-27
Implement CommandTimeout


caleblloyd:
**1. Implement `DefaultCommandTimeout` in the connection string**

[Connection String Documentation](https://dev.mysql.com/doc/connector-net/en/connector-net-connection-options.html)

**2. Implement `CommandTimeout` in `MySqlCommand`**

```
MySqlCommand cmd = new MySqlCommand();
cmd.CommandTimeout = 60;
```

[MySqlCommand Documentation](https://dev.mysql.com/doc/connector-net/en/connector-net-programming-mysqlcommand.html): 

**Implementation Notes:**

> MySQL Connector/Net 6.2 introduced timeouts that are aligned with how Microsoft handles SqlCommand.CommandTimeout. This property is the cumulative timeout for all network reads and writes during command execution or processing of the results. A timeout can still occur in the MySqlReader.Read method after the first row is returned, and does not include user processing time, only IO operations. The 6.2 implementation uses the underlying stream timeout facility, so is more efficient in that it does not require the additional timer thread as was the case with the previous implementation.

It sounds like Oracle's implementation of `CommandTimeout` is setting [Socket.SendTimeout](https://msdn.microsoft.com/en-us/library/system.net.sockets.socket.sendtimeout%28v=vs.110%29.aspx) and [Socket.ReceiveTimeout](https://msdn.microsoft.com/en-us/library/system.net.sockets.socket.receivetimeout%28v=vs.110%29.aspx).  They start at the timeout value (e.g. 30 seconds) and subtract only time spent in I/O.  This becomes the send or receive timeout for the next socket command.

caleblloyd:
I have a WIP of this here: https://github.com/caleblloyd/MySqlConnector/commit/d061332a495e08ac9282f2a1f97aee58141e59b7

Commands are properly timing out, but the next command that goes over the same connection hits issues with reading packets out of order.  Should the default logic be to reset a connection that hits a command timeout?

bgrainger:
This is potentially going to be complicated.

I would expect a `CommandTimeout` to behave the same as passing in a `CancellationToken` that cancels after a timeout. As per the notes on #3 this needs to involve `KILL QUERY` to free up server resources (and make the connection usable again). In that situation, the library shouldn't cancel socket operations, but actually read the final response from the server and leave the connection in a good state (so another command can be executed, it can be returned to the pool, etc.).

caleblloyd:
[...]

PR #229, opened 2017-04-07
Implement reader cancellation

### LINKED ISSUE #3: Implement cancellation of active reader
The consumer should be able to cancel an active reader via `MySqlCommand.Cancel()` or the `CancellationToken` passed to `ExecuteReaderAsync` etc.

This probably needs to be implemented via [`KILL QUERY`](http://dev.mysql.com/doc/refman/5.7/en/kill.html) which may require a separate connection. It may be helpful to implement connection pooling #2 first and keep a spare connection around to issue this command.

bgrainger:
Fixes #3.

Implements both `MySqlCommand.Cancel` (pre-async ADO.NET cancellation) and cancelling a `CancellationToken` passed to `ExecuteXAsync`/`ReadAsync`/`NextResultAsync`.

Cancellation is handled by opening a separate connection to the (IP address of the) current DB server and executing `KILL QUERY n` (where *n* is the `ServerThread` of the current connection). The `MySqlSession` stores the current command and reader so that calling `Cancel` on a non-active command does not incorrectly kill a different command executing on the same connection.

### DIFF
--- a/src/MySqlConnector/MySqlClient/MySqlDataReader.cs
+++ b/src/MySqlConnector/MySqlClient/MySqlDataReader.cs
@@ -108,8 +108,20 @@ namespace MySql.Data.MySqlClient
 
 		private async Task<ResultSet> ScanResultSetAsyncAwaited(IOBehavior ioBehavior, ResultSet resultSet, CancellationToken cancellationToken)
 		{
-			m_resultSetBuffered = await resultSet.ReadResultSetHeaderAsync(ioBehavior, cancellationToken);
-			return m_resultSetBuffered;
+			using (cancellationToken.Register(Command.Cancel))
+			{
+				try
+				{
+					m_resultSetBuffered = await resultSet.ReadResultSetHeaderAsync(ioBehavior, CancellationToken.None).ConfigureAwait(false);
+					return m_resultSetBuffered;
+				}
+				catch (MySqlException ex) when (ex.Number == (int) MySqlErrorCode.QueryInterrupted)
+				{
+					m_resultSetBuffered = null;
+					cancellationToken.ThrowIfCancellationRequested();
+					throw;
+				}
+			}
 		}
 
 		public override string GetName(int ordinal) => GetResultSet().GetName(ordinal);
--- a/src/MySqlConnector/MySqlClient/MySqlDataReader.cs
+++ b/src/MySqlConnector/MySqlClient/MySqlDataReader.cs
@@ -215,14 +215,29 @@ namespace MySql.Data.MySqlClient
 		internal static async Task<MySqlDataReader> CreateAsync(MySqlCommand command, CommandBehavior behavior, IOBehavior ioBehavior, CancellationToken cancellationToken)
 		{
 			var dataReader = new MySqlDataReader(command, behavior);
-			await dataReader.ReadFirstResultSetAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
-			command.Connection.ActiveReader = dataReader;
-			if (command.Connection.BufferResultSets)
+			command.Connection.Session.SetActiveReader(dataReader);
+
+			try
 			{
-				while (await dataReader.BufferNextResultAsync(ioBehavior, cancellationToken).ConfigureAwait(false) != null);
-				command.Connection.ActiveReader = null;
+				await dataReader.ReadFirstResultSetAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
+				if (command.Connection.BufferResultSets)
+				{
+					while (await dataReader.BufferNextResultAsync(ioBehavior, cancellationToken).ConfigureAwait(false) != null)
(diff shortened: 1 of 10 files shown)


src/MySqlConnector/Serialization/MySqlSession.cs
```diff
@@ -35,16 +36,97 @@ public MySqlSession(ConnectionPool pool, int poolGeneration)
...
+		public void DoCancel(MySqlCommand commandToCancel, MySqlCommand killCommand)
+		{
+			lock (m_lock)
+			{
+				if (m_activeCommand != commandToCancel)
+					return;
+
+				try
+				{
+					killCommand.ExecuteNonQuery();
```
caleblloyd: This appears to be the only locked section where a remote call runs.  Any way to move `killCommand.ExecuteNonQuery();` outside of the lock here?  Locking remote calls can invite deadlocks when the network is flaky...

bgrainger: I wasn't a fan of this, but I wanted to avoid the following problem:

thread 1 | thread 2
--- | ---
cmd1.Execute | 
_ | cmd1.Cancel
cmd1 finishes | open new connection
cmd2.Execute | wait for new connection to open
_ | KILL QUERY x
_ | ^ cmd2 is killed

We could dial down the `CommandTimeout` (once that is implemented #67) but I couldn't think of another approach here that makes sure cancellation (which has to run asynchronously) actually cancels the right command instead of just killing what happens to be actively running on that connection.

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -108,8 +108,20 @@ private ValueTask<ResultSet> ScanResultSetAsync(IOBehavior ioBehavior, ResultSet
 
 		private async Task<ResultSet> ScanResultSetAsyncAwaited(IOBehavior ioBehavior, ResultSet resultSet, CancellationToken cancellationToken)
 		{
-			m_resultSetBuffered = await resultSet.ReadResultSetHeaderAsync(ioBehavior, cancellationToken);
-			return m_resultSetBuffered;
+			using (cancellationToken.Register(Command.Cancel))
+			{
+				try
+				{
+					m_resultSetBuffered = await resultSet.ReadResultSetHeaderAsync(ioBehavior);
```
bgrainger: Yes, will do.

src/MySqlConnector/Serialization/MySqlSession.cs
bgrainger: I suppose I could have added this as a comment/justification in the code itself... 🤦‍♂️

bgrainger: I haven't actually fixed this, but I'm now setting `CommandTimeout` in expectation of that being implemented soon :smile:

Also added a comment about why this is being done.

caleblloyd: > wanted to avoid the following problem...

Ah, that is a good point.  

> We could dial down the `CommandTimeout`

Maybe should also dial down the `ConnectionTimeout` for the cancellation query.  I believe `CommandTimeout` will simply be calling this `.Cancel()` method, right?

bgrainger: Yes, we should be reducing the ConnectionTimeout as well; I'll add that.

bgrainger: Set to three seconds [here](https://github.com/mysql-net/MySqlConnector/pull/229/files#diff-7d25f7d140291c4c3a2bfb66fcc7a5a5R225). I don't have any good way of ensuring that failure to cancel is silently ignored (ADO.NET provides no mechanism to report if cancellation worked or not) or verify that it can be retried until `Cancel` succeeds. However the code *looks* good to me.

src/MySqlConnector/MySqlClient/MySqlErrorCode.cs
```diff
@@ -0,0 +1,18 @@
+namespace MySql.Data.MySqlClient
+{
+	/// <summary>
+	/// MySQL Server error codes. Taken from <a href="https://dev.mysql.com/doc/refman/5.7/en/error-messages-server.html">Server Error Codes and Messages</a>.
+	/// </summary>
+	public enum MySqlErrorCode
```
ejball: Could do `public static class MySqlErrorCodes` with `public const int` fields to avoid casting.

bgrainger: This is for eventual API compatibility: #232

(We could also code-gen the class you suggest although I'm not sure the duplication would be worth it.)

ejball: Ah, makes sense. Code-gen does not seem worth it.

PR #237, opened 2017-04-12
Allow passing of MySqlConnectionStringBuilder in constructor

MrSmoke:
Since `MySqlConnectionStringBuilder` is public anyway, I found that I could get more performance by passing in an instance of `MySqlConnectionStringBuilder` into `MySqlConnection` as opposed to having to parse the connection string every time I created a new `MySqlConnection`.

According to dotTrace, `MySqlConnectionStringBuilder` was spending around 10% of CPU time in Regex

### DIFF
--- a/src/MySqlConnector/MySqlClient/MySqlConnection.cs
+++ b/src/MySqlConnector/MySqlClient/MySqlConnection.cs
@@ -22,7 +22,10 @@ namespace MySql.Data.MySqlClient
 
 		public MySqlConnection(MySqlConnectionStringBuilder connectionStringBuilder)
 		{
-			SetConnectionSettings(connectionStringBuilder);
+			if (connectionStringBuilder == null)
+				throw new ArgumentNullException(nameof(connectionStringBuilder));
+
+			SetConnectionSettings(connectionStringBuilder.Clone());
 		}
 
 		public new MySqlTransaction BeginTransaction() => (MySqlTransaction) base.BeginTransaction();
--- a/src/MySqlConnector/MySqlClient/MySqlConnection.cs
+++ b/src/MySqlConnector/MySqlClient/MySqlConnection.cs
@@ -20,6 +20,11 @@ namespace MySql.Data.MySqlClient
 
 		public MySqlConnection(string connectionString) => ConnectionString = connectionString;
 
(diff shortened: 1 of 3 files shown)


src/MySqlConnector/MySqlClient/MySqlConnection.cs
```diff
@@ -349,6 +354,12 @@ private void CloseDatabase()
 			}
 		}
 
+		private void SetConnectionSettings(MySqlConnectionStringBuilder connectionStringBuilder)
+		{
+			m_connectionStringBuilder = connectionStringBuilder;
```
bgrainger: We need a private copy of this object, to avoid problems if the caller mutates it after passing it to the constructor.

src/MySqlConnector/MySqlClient/MySqlConnection.cs
```diff
@@ -20,6 +20,11 @@ public MySqlConnection()
 
 		public MySqlConnection(string connectionString) => ConnectionString = connectionString;
 
+		public MySqlConnection(MySqlConnectionStringBuilder connectionStringBuilder)
+		{
+			SetConnectionSettings(connectionStringBuilder);
```
bgrainger: We need a private copy of this object, to avoid problems if the caller mutates it after passing it to the constructor.

MrSmoke: Whoops yeah. I'm not sure if there's a clone pattern already in the library?

bgrainger:
The straightforward implementation of `Clone` adds keys from the current instance to a new one. However, the default implementation of `Add(key, value)` calls `this[key] = value`. According to my research in #238, the base class implementation of the setter for `this[string key]` is the problem, so this will probably be just as slow as the current code.

My performance benchmark measured a typical call to `new MySqlConnectionStringBuilder(connectionString)` at 20µs, which should be minuscule compared to the network round-trips involved in opening a connection. 

Can you elaborate a bit further on how instantiating a `MySqlConnection` object is a bottleneck in your scenario? I want to make sure I'm understanding it properly.

bgrainger:
Unclear whether this provides any performance improvement over the current code, nor whether this is a bottleneck worth optimising. See #238 for further discussion.

PR #241, opened 2017-04-15
Support System.Transactions.

bgrainger:
Adds support for ambient transactions, and for explicitly enlisting with `EnlistTransaction`.

Unlike Connector/NET, this is implemented using MySQL's [XA Transactions](https://dev.mysql.com/doc/refman/5.7/en/xa.html). This may be a compatibility problem for users migrating from Connector/NET, but better matches the purpose/intent of `TransactionScope`.

Currently only implemented for .NET Framework 4.5.1, since .NET Core won't support `System.Transactions` until netstandard-2.0: dotnet/corefx#10089

### DIFF
--- a/docs/content/tutorials/migrating-from-connector-net.md
+++ b/docs/content/tutorials/migrating-from-connector-net.md
@@ -39,8 +39,16 @@ MySqlConnector has some different default connection string options:
 Some command line options that are supported in Connector/NET are not supported in MySqlConnector.  For a full list of options that are
 supported in MySqlConnector, see the [Connection Options](connection-options)
 
+### TransactionScope
+
+MySqlConnector adds full distributed transaction support (for client code using [`TransactionScope`](https://msdn.microsoft.com/en-us/library/system.transactions.transactionscope.aspx)),
+while Connector/NET uses regular database transactions. As a result, code that uses `TransactionScope`
+may execute differently with MySqlConnector. To get Connector/NET-compatible behavior, remove
+`TransactionScope` and use `BeginTransaction`/`Commit` directly.
+
 ### Bugs present in Connector/NET that are fixed in MySqlConnector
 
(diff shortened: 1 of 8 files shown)


bgrainger:
After thinking about it some more, I am going to implement the `AutoEnlist` connection string setting. That will let a client convert this Connector/NET code:

```
using (var ts = new TransactionScope())
using (var conn = new MySqlConnection(connStr))
{
    conn.Open();
    // ...
    ts.Complete();
}
```

to the equivalent:

```
using (var ts = new TransactionScope())
using (var conn = new MySqlConnection(connStr + ";AutoEnlist=False"))
{
    conn.Open();
    using (var mySqlTrans = conn.BeginTransaction())
    {
        // ...
        mySqlTrans.Commit();
        ts.Complete();
    }
}
```

[These statements](https://dev.mysql.com/doc/refman/5.7/en/implicit-commit.html) aren't

bgrainger:
I'm going to implement the `AutoEnlist` connection string setting before merging this. Because this implementation uses XA Transactions it's subject to [these restrictions](https://dev.mysql.com/doc/refman/5.7/en/xa-restrictions.html) and clients can't use [these statements](https://dev.mysql.com/doc/refman/5.7/en/implicit-commit.html) during one. Supporting `AutoEnlist` will let clients opt into Connector/NET compatible behaviour as follows:

Old:

```
using (var ts = new TransactionScope())
using (var conn = new MySqlConnection(connStr))
{
    conn.Open();
    // ...
    ts.Complete();
}
```

Compatible:
```
using (var ts = new TransactionScope())
using (var conn = new MySqlConnection(connStr + ";AutoEnlist=False"))
{
    conn.Open();
    using (var mySqlTrans = conn.BeginTransaction())
    {
        // ...
        mySqlTrans.Commit();
        ts.Complete();
    }
}
```

bgrainger:
Implemented in [0.18.0](http://www.nuget.org/packages/MySqlConnector/0.18.0).

PR #248, opened 2017-04-21
Handle QueryInterrupted exceptions when canceling completed queries.

adampoit:
KILL QUERY will interrupt a subsequent query if the command it was intended to cancel has already completed, and the connection is idle. In order to handle this case, we issue a dummy query to catch the QueryInterrupted exception. See https://bugs.mysql.com/bug.php?id=45679.

I ran into this issue while using Dapper's `QueryImpl` method. If `reader.FieldCount == 0`, then Dapper will try to cancel the query. Upon the next command issued by the connection, it is immediately interrupted. 

Generally, using `Query` without selecting any columns is wrong, but this would be a problem for any query that completed before it was canceled. The MySQL.Data implementation from Oracle handles both of these cases without an error.

### DIFF
--- a/src/MySqlConnector/MySqlClient/MySqlDataReader.cs
+++ b/src/MySqlConnector/MySqlClient/MySqlDataReader.cs
@@ -293,16 +293,8 @@ namespace MySql.Data.MySqlClient
 					// KILL QUERY will kill a subsequent query if the command it was intended to cancel has already completed.
 					// In order to handle this case, we issue a dummy query to catch the QueryInterrupted exception.
 					// See https://bugs.mysql.com/bug.php?id=45679
-					var killClearCommand = new MySqlCommand("SELECT * FROM fake_table LIMIT 0;", connection);
-					try
-					{
-						killClearCommand.ExecuteReader();
-					}
-					catch (MySqlException ex)
-					{
-						if (ex.Number != (int) MySqlErrorCode.QueryInterrupted && ex.Number != (int) MySqlErrorCode.NoSuchTable)
-							throw;
-					}
+					var killClearCommand = new MySqlCommand("DO SLEEP(0);", connection);
+					killClearCommand.ExecuteNonQuery();
 				}
 
 				Command.ReaderClosed();
--- a/src/MySqlConnector/MySqlClient/MySqlDataReader.cs
+++ b/src/MySqlConnector/MySqlClient/MySqlDataReader.cs
@@ -287,6 +287,24 @@ namespace MySql.Data.MySqlClient
 				var connection = Command.Connection;
 				if (!connection.BufferResultSets)
 					connection.Session.FinishQuerying();
+
+				if (Command.IsCanceled)
+				{
+					// KILL QUERY will kill a subsequent query if the command it was intended to cancel has already completed.
+					// In order to handle this case, we issue a dummy query to catch the QueryInterrupted exception.
+					// See https://bugs.mysql.com/bug.php?id=45679
+					var killClearCommand = new MySqlCommand("SELECT * FROM fake_table LIMIT 0;", connection);
+					try
+					{
+						killClearCommand.ExecuteReader();
+					}
+					catch (MySqlException ex)
+					{
+						if (ex.Number != (int) MySqlErrorCode.QueryInterrupted && ex.Number != (int) MySqlErrorCode.NoSuchTable)
+							throw;
+					}
+				}
+
 				Command.ReaderClosed();
 				if ((m_behavior & CommandBehavior.CloseConnection) != 0)
 				{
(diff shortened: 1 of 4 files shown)


src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -287,6 +287,24 @@ private void DoClose()
 				var connection = Command.Connection;
 				if (!connection.BufferResultSets)
 					connection.Session.FinishQuerying();
+
+				if (Command.IsCanceled)
+				{
+					// KILL QUERY will kill a subsequent query if the command it was intended to cancel has already completed.
+					// In order to handle this case, we issue a dummy query to catch the QueryInterrupted exception.
+					// See https://bugs.mysql.com/bug.php?id=45679
+					var killClearCommand = new MySqlCommand("SELECT * FROM fake_table LIMIT 0;", connection);
```
bgrainger: Did you try something like `DO SLEEP(0.001)` to clear the pending cancellation?

src/MySqlConnector/Serialization/MySqlSession.cs
```diff
@@ -84,6 +84,8 @@ public void DoCancel(MySqlCommand commandToCancel, MySqlCommand killCommand)
 				// blocking the other thread for an extended duration.
 				killCommand.CommandTimeout = 3;
 				killCommand.ExecuteNonQuery();
+
+				commandToCancel.IsCanceled = true;
```
bgrainger: I think this `IsCanceled` property should be combined with the existing `MySqlSession.m_state` so that all state related to cancellation is in one place.

tests/SideBySide/CancelTests.cs
```diff
@@ -228,6 +228,18 @@ public void DapperQueryMultiple()
 			}
 		}
 
+		[Fact]
+		public async Task CancelCompletedCommand()
```
bgrainger: I'm not getting any kind of failure from this test when running it without the new code (so it's difficult to verify that the problem is fixed).

adampoit: What version of MySQL are you testing with? I tested this on 5.6.29. According to a [duplicate](https://bugs.mysql.com/bug.php?id=79838) of the bug I linked, the issue may be 

> repeatable with server version 5.1/5.5/5.6 and NOT REPEATABLE with server version 5.0 and 5.7

bgrainger: I was using 5.7.

I still can't repro a failure with this test with 5.6, although I can get an unhandled exception when using Dapper's `Query<int>` followed by `Execute`.

tests/SideBySide/CancelTests.cs
```diff
@@ -228,6 +228,18 @@ public void DapperQueryMultiple()
...
+		{
+			using (var cmd = m_database.Connection.CreateCommand())
+			{
+				cmd.CommandText = @"select 1;";
+
+				using (await cmd.ExecuteReaderAsync().ConfigureAwait(false))
+					cmd.Cancel();
+			}
```
bgrainger: Does this need to be followed by a second command that will be the unintentional victim of cancellation?

adampoit: The [Dispose](https://github.com/adampoit/MySqlConnector/blob/ef4f7eeb62e960ccb446a9a0ec5af12711931093/tests/SideBySide/CancelTests.cs#L17) method should be executing the second command, but perhaps it would be better to be explicit.

bgrainger: I tried `SELECT 2;` but that doesn't repro the problem, as per bullet 2 of [bug 45679](https://bugs.mysql.com/bug.php?id=45679). So I think the `INSERT`/`UPDATE` combination of commands is a good one.

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
bgrainger: `DO SLEEP(0)` is even better (for some reason I was assuming it had to be a non-zero wait, but `SLEEP(0)` seems to work fine).

adampoit: 👍

src/MySqlConnector/Serialization/MySqlSession.cs
bgrainger: I want to handle this situation in `MySqlSession.FinishQuerying`; I can push that change.

adampoit: I started messing around with the states, but if you know how you want to tackle this I have `Allow edits from maintainers` turned on, so go for it.

PR #250, opened 2017-04-26
Properly sum affected row count.

adampoit:
Before the number of affected rows would always get set to 1, regardless of the number of inserted rows.


bgrainger:
Looks like this was regressed in 8cfc592d5b3c3234a038d7a3c7309534311fb705 but we didn't have a test to catch it. Thanks for fixing it!

bgrainger:
Shipped in [0.19.1](https://www.nuget.org/packages/MySqlConnector/0.19.1).

Issue #251, opened 2017-04-30
Connection pool entry is leaked if MySqlConnection is not disposed


bgrainger:
Execute the following code:

```
var csb = new MySqlConnectionStringBuilder
{
	// server, username, password, etc.
	Pooling = true,
	MaximumPoolSize = 5,
	ConnectionTimeout = 5u,
};

for (int i = 0; i < 6; i++)
{
	var connection = new MySqlConnection(csb.ConnectionString);
	connection.Open(); // throws when i == 5
	GC.Collect();
}
```

If `MySqlConnection` is not disposed, then `MySqlSession.ReturnToPool` is not called, so `ConnectionPool.Return` doesn't release `m_sessionSemaphore`. If this happens repeatedly in client code, then eventually all open slots in the pool are leaked, and clients start timing out in `ConnectionPool.GetSessionAsync`.

One strategy to mitigate this might be for the connection pool to maintain a list of `WeakReference<MySqlSession>` objects. If a weak reference's target was `null`, we would know that a connection had been GC'ed without having been disposed, so could manually `Release` the semaphore to free up that slot. This could be done explicitly in `GetSessionAsync` (if `CurrentCount == 0`) or perhaps periodically in the background, e.g., by the connection pool reaper. (Disposing the connection would have to remove the weak reference from the list.)

bgrainger:
A potential implementation would be to use a finalizer; this would be bad for [the following reasons](https://msdn.microsoft.com/en-us/library/b1yfkh5e(v=vs.100).aspx):

> * Implement Finalize only on objects that require finalization. There are performance costs associated with Finalize methods.
> * The Finalize method should not reference any other objects.

And [from here](https://msdn.microsoft.com/en-us/library/ddae83kx(v=vs.71).aspx):

> An object's Finalize method should not call a method on any objects other than that of its base class.

PR #252, opened 2017-04-30
Recover leaked connection pool entries

bgrainger:
Uses a `WeakReference` to track `MySqlSession` objects, and restores free slots to the connection pool if it detects that a `MySqlSession` object has been garbage-collected (without the associated `MySqlConnection` having been disposed).

Although `MySqlConnection` objects should always be disposed, this connector should be resilient to client code that fails to do so.

### DIFF
--- a/src/MySqlConnector/MySqlClient/ConnectionPool.cs
+++ b/src/MySqlConnector/MySqlClient/ConnectionPool.cs
@@ -14,7 +14,13 @@ namespace MySql.Data.MySqlClient
 		{
 			cancellationToken.ThrowIfCancellationRequested();
 
-			// wait for an open slot
+			// if all sessions are used, see if any have been leaked and can be recovered
+			// check at most once per second (although this isn't enforced via a mutex so multiple threads might block
+			// on the lock in RecoverLeakedSessions in high-concurrency situations
+			if (m_sessionSemaphore.CurrentCount == 0 && unchecked(((uint) Environment.TickCount) - m_lastRecoveryTime) >= 1000u)
+				RecoverLeakedSessions();
+
+			// wait for an open slot (until the cancellationToken is cancelled, which is typically due to timeout)
 			if (ioBehavior == IOBehavior.Asynchronous)
 				await m_sessionSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
 			else
@@ -46,14 +52,19 @@ namespace MySql.Data.MySqlClient
 						{
 							await session.ResetConnectionAsync(m_connectionSettings, ioBehavior, cancellationToken).ConfigureAwait(false);
 						}
+
 						// pooled session is ready to be used; return it
+						lock (m_leasedSessions)
+							m_leasedSessions.Add(session.Id, new WeakReference<MySqlSession>(session));
 						return session;
 					}
 				}
 
 				// create a new session
-				session = new MySqlSession(this, m_generation);
+				session = new MySqlSession(this, m_generation, Interlocked.Increment(ref m_lastId));
 				await session.ConnectAsync(m_connectionSettings, ioBehavior, cancellationToken).ConfigureAwait(false);
+				lock (m_leasedSessions)
+					m_leasedSessions.Add(session.Id, new WeakReference<MySqlSession>(session));
 				return session;
 			}
 			catch
@@ -82,6 +93,8 @@ namespace MySql.Data.MySqlClient
 		{
 			try
 			{
+				lock (m_leasedSessions)
+					m_leasedSessions.Remove(session.Id);
 				if (SessionIsHealthy(session))
 					lock (m_sessions)
 						m_sessions.AddFirst(session);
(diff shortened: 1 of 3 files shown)


src/MySqlConnector/MySqlClient/ConnectionPool.cs
```diff
@@ -14,6 +14,13 @@ internal sealed class ConnectionPool
 		{
 			cancellationToken.ThrowIfCancellationRequested();
 
+			// if all sessions are used, see if any have been leaked and can be recovered
+			if (m_sessionSemaphore.CurrentCount == 0)
+			{
+				GC.Collect();
+				RecoverLeakedSessions();
```
caleblloyd: This could happen potentially thousands of times in a second, and calls locked code. Maybe adapt it to run at most once per second?

src/MySqlConnector/MySqlClient/ConnectionPool.cs
```diff
@@ -241,5 +280,7 @@ private ConnectionPool(ConnectionSettings cs)
 		readonly SemaphoreSlim m_sessionSemaphore;
 		readonly LinkedList<MySqlSession> m_sessions;
 		readonly ConnectionSettings m_connectionSettings;
+		readonly Dictionary<int, WeakReference<MySqlSession>> m_leasedSessions;
+		int m_lastId;
```
caleblloyd: `long` or `ulong` may be better. Although it'd be rare, `int`could overflow in a high load scenario given enough time

src/MySqlConnector/MySqlClient/ConnectionPool.cs
bgrainger: The other option would be just to delete this whole `if` statement. and hope that connections are being leaked slowly enough that they've been GCed and cleaned up by the background task before the pool is emptied. 

This would stop "punishing" clients who are correctly disposing their connections but are legitimately hitting an empty pool.

src/MySqlConnector/MySqlClient/ConnectionPool.cs
bgrainger: I considered this, but even if you opened 10,000 connections/sec, it would take 4.97 days to wrap around, then you'd only have a problem (ID collision) if one of those connections had been kept open for five days. (Under more typical load, ~150 conn/s, it wouldn't wrap around for a year.)

The downside of `long` is that `Interlocked.Increment` has to be simulated on a 32-bit system, which is still a potential target platform for this code.

src/MySqlConnector/MySqlClient/ConnectionPool.cs
bgrainger: Or... perform the full GC and scan only if a timeout exception is thrown so we try to auto-recover only in a completely-leaked pool situation.

caleblloyd: I was under the assumption that once the connection was Garbage Collected, it's Dispose method would return it to the pool. When is that not the case?

src/MySqlConnector/MySqlClient/ConnectionPool.cs
caleblloyd: As long as it wraps around I think it's fine as `int`. I wasn't sure​ if C# throws an OverflowException.

src/MySqlConnector/MySqlClient/ConnectionPool.cs
bgrainger: The .NET GC does not invoke objects' `Dispose' methods; that has to be done manually by the consumer (typically using a `using` block).

The .NET finalizer will run a finalizable object's destructor (i.e., `~MyObject()`) after the object has been collected. In the typical implementation of the dispose pattern, the finalizer will call `Dispose(false)`. (So in that sense "the GC calls dispose", but in general `Dispose` is not called deterministically, unlike say C++ destructors.)

I enumerated some reasons why I don't think using a finalizer is the right solution in #251.

bgrainger: I'm forcing a GC here so that `MySqlSession` objects leaked by clients are collected, so that their associated `WeakReference` will have a `null` target.

src/MySqlConnector/MySqlClient/ConnectionPool.cs
bgrainger: > This method handles an overflow condition by wrapping: if _location_ = `Int32.MaxValue`, _location_ + 1 = `Int32.MinValue`. No exception is thrown.

[MSDN](https://msdn.microsoft.com/en-us/library/dd78zt0c(v=vs.110).aspx)

src/MySqlConnector/MySqlClient/ConnectionPool.cs
bgrainger: > perform the full GC and scan only if a timeout exception is thrown so we try to auto-recover only in a completely-leaked pool situation

This approach doesn't really work because the cancellation token is now permanently cancelled (because the connect timeout has elapsed) so all further operations with it abort immediately.

bgrainger: Pushed an update that tries to throttle it to once per second. The check doesn't use a mutex, so if multiple threads arrive at an empty pool at the same time, they could still all attempt to take the `lock (m_leasedSessions)` lock and block on it. But it should stop *every* thread that arrives at an empty pool from calling `GC.Collect` and taking the lock (as was being done in the initial PR).

PR #253, opened 2017-05-03
ChangeDatabaseAsync method and CancellationToken parameter

Nicholi:
Looks like minor error, forgot to pass along parameter.


bgrainger:
Thanks for the fix, @Nicholi !

bgrainger:
Fixed in [0.19.3](https://www.nuget.org/packages/MySqlConnector/0.19.3).

PR #265, opened 2017-05-25
Escape single quote and backslash when writing GUIDs as byte arrays.

adampoit:
This code is copied from [writing a byte array](https://github.com/mysql-net/MySqlConnector/blob/master/src/MySqlConnector/MySqlClient/MySqlParameter.cs#L124-L129). 

These byte values were not escaped when writing a GUID as a byte array, causing syntax errors when doing something like

```
var query = @"insert into table(value) select @value from dual";
database.Connection.Execute(query, new { value = Guid.NewGuid() });
```

when `OldGuids=true`.

Since the code is duplicated, I can refactor it into a method if we feel that's appropriate.


bgrainger:
Fixed in [0.19.5](https://github.com/mysql-net/MySqlConnector/releases/tag/0.19.5).

PR #271, opened 2017-06-06
Add fake MySQL Server

bgrainger:
Exploration of adding a (simple so far) fake MySQL Server, for testing the protocol without using a real MySQL Server instance.

This allows introspection of the server state, so (for example) we can test that `Minimum Pool Size` opens the right number of initial connections. It should also allow unit testing of (simulated) behavior observed from various servers in the wild without having to recreate the exact conditions that put the server in that state.

Drawbacks are obviously that (due to bugs in the fake server) it could lead to testing conditions that never happen in real-world usage. Alternatively, a bug in the fake server could mask a bug in the client, resulting in a false passing test. (This is currently mitigated by reimplementing much of the protocol from scratch instead of reusing code from the connector.) Additionally, the state machine required to implement the server could become extremely complex, especially if faking query results were ever added to the server.

The current tests simply verify that connections aren't closed if pooling is enabled, but are closed if it's disabled.

### DIFF
--- a/tests/SideBySide/ConnectionPool.cs
+++ b/tests/SideBySide/ConnectionPool.cs
@@ -173,6 +173,9 @@ namespace SideBySide
 
 				// have to GC for leaked connections to be removed from the pool
 				GC.Collect();
+
+				// HACK: have to sleep (so that RecoverLeakedSessions is called in ConnectionPool.GetSessionAsync?)
+				Thread.Sleep(250);
 			}
 		}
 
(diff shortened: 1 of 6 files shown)


bgrainger:
rebuild this

tests/SideBySide/ConnectionPool.cs
```diff
@@ -205,35 +205,6 @@ public void LeakReaders()
 			}
 		}
 
-		[Theory]
-		[InlineData(2u, 3u, true)]
```
bgrainger: Removing this specific test makes `LeakReaders` fail; apparently there's some interference between the two tests 😞

PR #272, opened 2017-06-07
Target netstandard2.0

caleblloyd:
- Opening to preview changes coming in `netstandard2.0`
- Will live in `netstandard2.0` branch until `netcoreapp2.0` and `netstandard2.0` RTM

### DIFF
--- a/src/MySqlConnector/Serialization/MySqlSession.cs
+++ b/src/MySqlConnector/Serialization/MySqlSession.cs
@@ -442,6 +442,7 @@ namespace MySql.Data.Serialization
 					try
 					{
 						tcpClient = new TcpClient(ipAddress.AddressFamily);
+
 						using (cancellationToken.Register(() => tcpClient?.Client?.Dispose()))
 						{
 							try
@@ -455,7 +456,21 @@ namespace MySql.Data.Serialization
 #if NETSTANDARD1_3
 									await tcpClient.ConnectAsync(ipAddress, cs.Port).ConfigureAwait(false);
 #else
-									tcpClient.Connect(ipAddress, cs.Port);
+									if (Utility.IsWindows())
+									{
+										tcpClient.Connect(ipAddress, cs.Port);
+									}
+									else
+									{
+										// non-windows platforms block on synchronous connect, use send/receive timeouts: https://github.com/dotnet/corefx/issues/20954
+										var originalSendTimeout = tcpClient.Client.SendTimeout;
+										var originalReceiveTimeout = tcpClient.Client.ReceiveTimeout;
+										tcpClient.Client.SendTimeout = cs.ConnectionTimeoutMilliseconds;
+										tcpClient.Client.ReceiveTimeout = cs.ConnectionTimeoutMilliseconds;
+										tcpClient.Connect(ipAddress, cs.Port);
+										tcpClient.Client.SendTimeout = originalSendTimeout;
+										tcpClient.Client.ReceiveTimeout = originalReceiveTimeout;
+									}
 #endif
 								}
 							}
--- a/src/MySqlConnector/Serialization/MySqlSession.cs
+++ b/src/MySqlConnector/Serialization/MySqlSession.cs
@@ -203,8 +203,12 @@ namespace MySql.Data.Serialization
 				throw new MySqlException("Unable to connect to any of the specified MySQL hosts.");
 			}
 
-			var socketByteHandler = new SocketByteHandler(m_socket);
-			m_payloadHandler = new StandardPayloadHandler(socketByteHandler);
+#if NETSTANDARD2_0
+ 			var byteHandler = new StreamByteHandler(m_networkStream);
+#else
+			var byteHandler = new SocketByteHandler(m_socket);
+#endif
+			m_payloadHandler = new StandardPayloadHandler(byteHandler);
 
 			var payload = await ReceiveAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
 			var reader = new ByteArrayReader(payload.ArraySegment.Array, payload.ArraySegment.Offset, payload.ArraySegment.Count);
@@ -319,6 +323,9 @@ namespace MySql.Data.Serialization
 				return true;
 			}
 			catch (EndOfStreamException)
+			{
+			}
+			catch (IOException)
 			{
 			}
 			catch (SocketException)
--- a/tests/SideBySide/SideBySide.csproj
+++ b/tests/SideBySide/SideBySide.csproj
@@ -1,11 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
  <PropertyGroup Condition=" '$(Configuration)' != 'Baseline' ">
-    <TargetFrameworks>net46;net462;netcoreapp1.1.2;netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks>net461;netcoreapp1.1.2;netcoreapp2.0</TargetFrameworks>
   </PropertyGroup>
 
  <PropertyGroup Condition=" '$(Configuration)' == 'Baseline' ">
-    <TargetFrameworks>net462</TargetFrameworks>
+    <TargetFrameworks>net461</TargetFrameworks>
     <DefineConstants>BASELINE</DefineConstants>
   </PropertyGroup>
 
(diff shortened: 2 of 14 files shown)


caleblloyd:
There's some new errors on both Windows and Linux under `netcoreapp2.0`

Nothing jumps out at me right away - we should research further to figure out if there is any CLR bugs that need to be opened upstream

caleblloyd:
Also worth noting: `net462` appears to still be loading the `net451` library package, even though it is compatible with the `netstandard2.0` library package :frowning_face:

caleblloyd:
Interesting - in `MySqlSession.ConnectAsync`, using `StreamByteHandler` causes a failure on AppVeyor but `SocketByteHandler` does not under `netstandard2.0`.

[Failed on AppVeyor](https://ci.appveyor.com/project/mysqlnet/mysqlconnector/build/1.0.617):
```
#if NETSTANDARD2_0
 			var byteHandler = new StreamByteHandler(m_networkStream);
#else
			var byteHandler = new SocketByteHandler(m_socket);
#endif
			m_payloadHandler = new StandardPayloadHandler(byteHandler);
```

[Passed on AppVeyor](https://ci.appveyor.com/project/mysqlnet/mysqlconnector/build/1.0.618):
```
 			var byteHandler = new SocketByteHandler(m_socket);
 			m_payloadHandler = new StandardPayloadHandler(byteHandler);
```

This could have implications when SSL is enabled on Windows, since that uses `StreamByteHandler`

src/MySqlConnector/Serialization/MySqlSession.cs
```diff
@@ -203,8 +203,12 @@ public void FinishQuerying()
 				throw new MySqlException("Unable to connect to any of the specified MySQL hosts.");
 			}
 
-			var socketByteHandler = new SocketByteHandler(m_socket);
-			m_payloadHandler = new StandardPayloadHandler(socketByteHandler);
+#if NETSTANDARD2_0
+ 			var byteHandler = new StreamByteHandler(m_networkStream);
```
caleblloyd: The .NET Core 2.0 [performance blog post](https://blogs.msdn.microsoft.com/dotnet/2017/06/07/performance-improvements-in-net-core/) points out that:

> Not only do such improvements accrue to components using sockets directly, but also to using sockets indirectly via higher-level components, and other PRs have resulted in additional performance increases in higher-level components, such as NetworkStream. For example, PR dotnet/corefx #16502 re-implemented Socket’s Task-based SendAsync and ReceiveAsync operations on top of SocketAsyncEventArgs and then allowed those to be used from NetworkStream.Read/WriteAsync, and PR dotnet/corefx #12664 added a specialized CopyToAsync override to support more efficiently reading the data from a NetworkStream and copying it out to some other stream. Those changes have a very measurable impact on NetworkStream throughput and allocations.

Since SSL must already use `NetworkStream` and the benefits from `SocketAwaitable` have been rolled into `NetworkStream`, it'd be easier to just support `NetworkStream`.

bgrainger: It would be easier, but there will still be the overhead of allocating `Task`s even to represent those synchronous operations.

I didn't write up the gains (or finish this implementation) but I was finding a performance increase from removing `ValueTask` because its asynchronous version has to allocate a `Task`: https://github.com/bgrainger/MySqlConnector/commits/rewrite-value-task

I'm very interested in testing my experimental code against the new `Socket` and `NetworkStream` in netstandard2.0.

caleblloyd:
`SideBySide.ConnectSync.ConnectTimeout` test failure on Linux opened upstream: https://github.com/dotnet/corefx/issues/20954

caleblloyd:
I am unsure why `SideBySide.CancelTests.CancelSlowQueryWithTokenAfterExecuteReader` takes so long on TravisCI.  On my Ubuntu 17.04 laptop:

```
CancelSlowQueryWithTokenAfterNextResult passed in 503ms
```

Either Travis is just plain slow, or there's something in Ubuntu 14.04 that's since been fixed that makes it slower.

src/MySqlConnector/Serialization/MySqlSession.cs
```diff
@@ -432,9 +439,17 @@ private void VerifyConnected()
...
 					TcpClient tcpClient = null;
+					int originalSendTimeout, originalReceiveTimeout;
 					try
 					{
 						tcpClient = new TcpClient(ipAddress.AddressFamily);
+
+						// unix platforms block on connect, use send/receive timeouts: https://github.com/dotnet/corefx/issues/20954
+						originalSendTimeout = tcpClient.Client.SendTimeout;
+						originalReceiveTimeout = tcpClient.Client.ReceiveTimeout;
+						tcpClient.Client.SendTimeout = cs.ConnectionTimeoutMilliseconds;
```
caleblloyd: I think get/set socket timeouts call `getsocketopt` and `setsocketopt` syscalls. This commit adds 6 get/set socketopt syscalls. These need to be restricted to non-windows platforms as to not cause unnecessary syscalls.

caleblloyd:
@bgrainger I've pretty much got everything working except the benchmark project.  Not too familiar with that, could you help me out on getting it to compile with these changes?  Then this is ready for review.

bgrainger:
Yes, I can take a look at that.

bgrainger:
I build `Benchmark.sln` with the MSBuild engine from VS2017 because I wanted to use C# extern aliases to reference both MySqlConnector and MySql.Data in the same project (and I [can't find a way](https://stackoverflow.com/questions/35899657/how-to-extern-alias-an-assembly-with-net-core) to do this in .NET Core).

It looks like VS2017 doesn't like `<TargetFrameworks>...;netstandard2.0</TargetFrameworks>` in MySqlConnector.csproj; if I delete that framework then everything builds OK. (Makes sense that it doesn't support it yet, doesn't make sense that it fails by issuing compilation errors when building the `netstandard1.3` target.)

Having the old and new connectors side-by-side was really helpful/efficient when trying to improve the performance of MySqlConnector relative to the old connector. Now that that's done, the benchmark mostly serves to catch regressions (e.g., if someone were to open a PR with major changes). I think long term I should remove MySql.Data from the benchmark, port the csproj to the dotnet format, and use BenchmarkDotNet to compare net46 vs netcoreapp1.1 vs netcoreapp2.0. That should fix this build error.

But for now, on this branch, I don't mind if you just delete lines 53ff from `test.ps1` to remove Benchmark from the netstandard2.0 build. If I fix the Benchmark as described, I could merge that into the netstandard2.0 branch and re-enable it under Appveyor.

bgrainger:
When dotnet/BenchmarkDotNet#466 is implemented, it may be easier to bring back comparison tests against MySql.Data.

bgrainger:
I think I've fixed the Benchmark project for netstandard2.0 here: https://github.com/mysql-net/MySqlConnector/tree/ns2-test

Do you want to squash this all down to one commit on top of `master`?

caleblloyd:
Oh nice! Everything else in the PR look good?

If so, let's squash and merge your `ns2-test` branch to the `netstandard2.0` branch and close this PR. I'd say keep this out of master until netstandard2.0 is GA.

tests/SideBySide/SideBySide.csproj
```diff
@@ -1,11 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
  <PropertyGroup Condition=" '$(Configuration)' != 'Baseline' ">
-    <TargetFrameworks>netcoreapp1.1.1;net462</TargetFrameworks>
```
bgrainger: Any particular reason to change `net462` -> `net461` and `netcoreapp1.1.1` -> `netcoreapp1.1.2` here? (Mostly just curious.)

bgrainger:
Squashed to the [netstandard2.0 branch](https://github.com/mysql-net/MySqlConnector/tree/netstandard2.0).

tests/SideBySide/SideBySide.csproj
caleblloyd: `4.6.1` is the lowest desktop target that `netstandard2.0` supports.  It will use the `net451` library as long as we still build that, but if we ever remove `net451`, `4.6.1` will be the lowest level supported.

Moved to `netcoreapp1.1.2` because it is the most recent shared framework for .NET Core `1.x`.  This will test the `netstandard1.3` package, although any version of `netcoreapp1.x` should work.  I just find it's easiest to bring them up to the latest.

PR #280, opened 2017-06-22
Add CA Certificate validation

### LINKED ISSUE #278: Specify Acceptable Server CA / Expose SslStream?
I'm reading through MySqlSession.cs, and it looks like I have no control over the creation of the underlying SslStream  or the remote certificate validation callback.  I would like to pin the server certificate by thumbprint or public key in my application, but this does not appear to be exposed.  

Is there a way to specify an SSL CA certificate to validate the server's cert with (besides installing it as a trusted CA on the client)?  

If not, then as a feature request, can I have more direct access to the SslStream's remote certificate validation callback?  

I'm not 100% sure how certificate pinning is done with MySql in general, but if you know where to put it in the connector api, I can possibly submit a patch.

caleblloyd:
- Fixes #278 
- Please review for technical issues.  I took most of the implementation from [the two SO answers here](https://stackoverflow.com/questions/7695438/verify-remote-server-x509certificate-using-ca-certificate-file).  If there is a better way to validate certificate chains, comment away.
- I've hand tested; need to write Unit tests for this before it is ready to be merged


src/MySqlConnector/Serialization/ConnectionSettings.cs
```diff
@@ -71,8 +72,9 @@ public ConnectionSettings(MySqlConnectionStringBuilder csb)
 
 		// SSL/TLS Options
 		public MySqlSslMode SslMode { get; }
-		public string CertificateFile { get; }
-		public string CertificatePassword { get; }
+		public string ClientCertificate { get; }
+		public string ClientCertificatePassword { get; }
+		public string CACertificate { get; }
```
bgrainger: Although I wonder if `ClientCertificateFilePath` would be better for clarity.

src/MySqlConnector/Serialization/MySqlSession.cs
```diff
@@ -548,19 +552,43 @@ private void VerifyConnected()
 			Func<object, X509Certificate, X509Chain, SslPolicyErrors, bool> remoteCertificateCb =
 				(rcbSender, rcbCertificate, rcbChain, rcbPolicyErrors) =>
 				{
-					switch (rcbPolicyErrors)
+					if (cs.SslMode == MySqlSslMode.Preferred || cs.SslMode == MySqlSslMode.Required)
+						return true;
+
+					if ((rcbPolicyErrors & SslPolicyErrors.RemoteCertificateChainErrors) > 0 && cs.CACertificate != null)
```
bgrainger: Nitpick: I would write `!= 0` for testing a bitflag. (`> 0` first made me think it was a count of errors, not testing a flag.)

src/MySqlConnector/Serialization/MySqlSession.cs
```diff
@@ -548,19 +552,43 @@ private void VerifyConnected()
...
+						try
+						{
+							caCertificate = new X509Certificate2(cs.CACertificate);
+						}
+						catch (CryptographicException ex)
+						{
+							if (!File.Exists(cs.CACertificate))
+								throw new MySqlException("Cannot find CA Certificate File", ex);
+							throw new MySqlException("The CA Certificate File is invalid", ex);
+						}
```
bgrainger: I wonder if constructing `caCertificate` should be done outside the callback. I also wonder if any `MySqlException` that is thrown gets passed through, or would it end up being wrapped by, say, a `CryptographicException`? (I haven't tested to see.)

src/MySqlConnector/MySqlClient/MySqlConnectionStringBuilder.cs
```diff
@@ -54,16 +54,22 @@ public MySqlSslMode SslMode
 			set => MySqlConnectionStringOption.SslMode.SetValue(this, value);
 		}
 
-		public string CertificateFile
```
caleblloyd: Oh yea, forgot this class was Public.  I'll change it back.

src/MySqlConnector/MySqlClient/MySqlConnectionStringBuilder.cs
```diff
@@ -314,12 +321,16 @@ static MySqlConnectionStringOption()
 				keys: new[] { "SSL Mode", "SslMode" },
 				defaultValue: MySqlSslMode.Preferred));
 
-			AddOption(CertificateFile = new MySqlConnectionStringOption<string>(
-				keys: new[] { "CertificateFile", "Certificate File" },
+			AddOption(ClientCertificate = new MySqlConnectionStringOption<string>(
+				keys: new[] { "ClientCertificate", "Client Certificate", "CertificateFile", "Certificate File" },
```
caleblloyd: We've got 2 things named `Certificate` now, a `Client Certificate` and a `CA Certificate`.  I think an alias could help users qualify which one they're dealing with.

src/MySqlConnector/Serialization/ConnectionSettings.cs
caleblloyd: Yes, adding `FilePath` makes sense

src/MySqlConnector/Serialization/MySqlSession.cs
caleblloyd: It probably should.  There's an extremely rare case where the CA was already validated and this construction gets skipped completely, but in that case they wouldn't need to supply a CA certificate at all.  I'll move it out.

caleblloyd:
I reverted all of the naming changes and am simply calling the new option `CACertificateFile`.  I think it should be clear enough that `CertificateFile` means "client certificate", so on second thought no need for an alias, which could create more confusion.

src/MySqlConnector/Serialization/MySqlSession.cs
```diff
@@ -526,47 +526,81 @@ private void VerifyConnected()
 
 		private async Task InitSslAsync(ProtocolCapabilities serverCapabilities, ConnectionSettings cs, IOBehavior ioBehavior, CancellationToken cancellationToken)
 		{
-			X509CertificateCollection clientCertificates = null;
+			X509CertificateCollection CertificateFiles = null;
```
bgrainger: This still only contains (a single) client certificate, correct? If so, I think the old name of `clientCertificates` is clearer.

src/MySqlConnector/Serialization/MySqlSession.cs
```diff
@@ -526,47 +526,81 @@ private void VerifyConnected()
...
 			{
 				try
 				{
-					var certificate = new X509Certificate2(cs.CertificateFile, cs.CertificatePassword);
-					clientCertificates = new X509CertificateCollection {certificate};
+					X509Certificate2 certificate;
+					if (cs.CertificatePassword == null)
+						certificate = new X509Certificate2(cs.CertificateFile);
+					else
+						certificate = new X509Certificate2(cs.CertificateFile, cs.CertificatePassword);
```
bgrainger: Is this `if`/`else` necessary? It looks like the old code worked to pass a potentially-null `cs.CertificatePassword` as the second argument to the constructor, and I don't see any indication in the documentation that it can't be `null`. In corefx, the [single-argument constructor](https://github.com/dotnet/corefx/blob/master/src/System.Security.Cryptography.X509Certificates/src/System/Security/Cryptography/X509Certificates/X509Certificate.cs#L110) just seems to pass `null` when it delegates to an overload.

So it feels like this could be simplified back to the old code.

src/MySqlConnector/Serialization/MySqlSession.cs
```diff
@@ -526,47 +526,81 @@ private void VerifyConnected()
...
+				{
+					if (!File.Exists(cs.CACertificateFile))
+						throw new MySqlException("Cannot find CA Certificate File", ex);
+					throw new MySqlException("The CA Certificate File is invalid", ex);
 				}
 			}
 
-			Func<object, string, X509CertificateCollection, X509Certificate, string[], X509Certificate> localCertificateCb =
-				(lcbSender, lcbTargetHost, lcbLocalCertificates, lcbRemoteCertificate, lcbAcceptableIssuers) => lcbLocalCertificates[0];
+			X509Certificate LocalCertificateCb(object lcbSender, string lcbTargetHost, X509CertificateCollection lcbLocalCertificates, X509Certificate lcbRemoteCertificate, string[] lcbAcceptableIssuers) => lcbLocalCertificates[0];
```
bgrainger: Local function :+1:

src/MySqlConnector/Serialization/MySqlSession.cs
```diff
@@ -526,47 +526,81 @@ private void VerifyConnected()
...
+
+				if (cs.SslMode == MySqlSslMode.VerifyCA)
+					rcbPolicyErrors &= ~SslPolicyErrors.RemoteCertificateNameMismatch;
+
+				return rcbPolicyErrors == SslPolicyErrors.None;
+			}
 
 			SslStream sslStream;
-			if (clientCertificates == null)
+			if (CertificateFiles == null || CertificateFiles.Count == 0)
```
bgrainger: I'm not sure how `CertificateFiles.Count == 0` would happen.

bgrainger: This always-false condition should be removed, or line 641 should be updated with equivalent logic if it can happen (I lean towards the former).

src/MySqlConnector/Serialization/MySqlSession.cs
caleblloyd: Dang find/replace got me here

src/MySqlConnector/Serialization/MySqlSession.cs
caleblloyd: Right, shouldn't happen. It just doesn't feel right referring to the element [0] without checking the count, but it has worked fine without this check so far. I'll take it out.

src/MySqlConnector/Serialization/MySqlSession.cs
caleblloyd: Thanks for the CoreFX link. I wanted to make sure importing PEM encoded certs would work too and usually those aren't password protected. I'll change it back though since it runs through the same codepath.

caleblloyd:
I'll fix the things highlighted, thanks. I also need to add a handful of unit tests to test the CA validation and also check operation with PEM encoded certs.

caleblloyd:
@bgrainger I've made the changes and added tests.  This should be ready.

caleblloyd:
Also I ran the SSL tests from Windows to a SSL Enabled MySQL server (since Appveyor doesn't have SSL enabled).  They all passed, both our tests and the `BASELINE` tests.

bgrainger:
Shipped in [0.22.0](https://github.com/mysql-net/MySqlConnector/releases/tag/0.22.0).

PR #287, opened 2017-06-28
Test with MySQL Server 5.6.

bgrainger:
Anecdotally, MySQL Server 5.6 still seems fairly popular, so we should routinely test against it to catch compatibility issues.

This adds `ServerFeatures.LargePackets` to control whether large BLOBs should be sent to the server. Although `max_allowed_packet` can be increased for MySQL Server 5.6, inserting a large BLOB will still fail with the message:

> The size of BLOB/TEXT data inserted in one transaction is greater than 10% of redo log size. Increase the redo log size using innodb_log_file_size.

Rather than customizing this option for MySQL Server 5.6, I've simply feature-flagged the affected tests. (They do pass locally with a correctly-configured `mysql:5.6` Docker container.)


bgrainger:
@caleblloyd bump in case you didn't see this; any objections?

Issue #290, opened 2017-06-29
Fix "Aborted connection" server errors


bgrainger:
Run the `NoPassword` tests: `dotnet test tests\SideBySide\SideBySide.csproj -f net462 --filter=FullyQualifiedName~NoPassword`

The MySQL error log will contain output similar to the following:

```
2017-06-28T23:57:45.172394Z 282 [Note] Aborted connection 282 to db: 'unconnected' user: 'mysqltest' host: 'localhost' (Got an error reading communication packets)
2017-06-28T23:57:45.172394Z 281 [Note] Aborted connection 281 to db: 'unconnected' user: 'mysqltest' host: 'localhost' (Got an error reading communication packets)
2017-06-28T23:57:45.172394Z 283 [Note] Aborted connection 283 to db: 'unconnected' user: 'no_password' host: 'localhost' (Got an error reading communication packets)
2017-06-28T23:57:45.172394Z 285 [Note] Aborted connection 285 to db: 'unconnected' user: 'no_password' host: 'localhost' (Got an error reading communication packets)
2017-06-28T23:57:45.172394Z 284 [Note] Aborted connection 284 to db: 'unconnected' user: 'no_password' host: 'localhost' (Got an error reading communication packets)
```

Run the same tests with the baseline library: `dotnet restore tests\SideBySide\SideBySide.csproj /p:Configuration=Baseline && dotnet test tests\SideBySide\SideBySide.csproj -c Baseline --filter=FullyQualifiedName~NoPassword`

There are no errors in the error log.

Find out why the server is logging aborted connections and fix it.

bgrainger:
The lack of these errors in Connector/Net may be due to [clearing pools at exit](https://github.com/mysql/mysql-connector-net/blob/5864e6b21a8b32f5154b53d1610278abb3cb1cee/Source/MySql.Data/MySqlPoolManager.cs#L48-L49).

However, MySqlConnector still logs errors during the test run, not only at the end.

Connector/Net logs during a test run:

```
2017-06-29T02:55:49.243361Z 33 [Note] Access denied for user 'mysqltest'@'localhost' (using password: YES)
2017-06-29T02:55:51.736886Z 56 [Note] Access denied for user 'mysqltest'@'localhost' (using password: YES)
2017-06-29T02:55:55.526099Z 30 [Note] Aborted connection 30 to db: 'mysqltest' user: 'mysqltest' host: 'localhost' (Got packets out of order)
[...]

PR #291, opened 2017-06-29
Fix "Aborted connection" error when connection leaks.

bgrainger:
This fixes most of the errors logged during a test run noted in #290.

It accomplishes this by making `MySqlSession` have a `WeakReference<MySqlConnection>` to its owning connection, so we can return the session to the pool after the connection has been GCed. Reap and ClearPool both now recover sessions before cleaning.

There still appears to be one final "Aborted connection ... (Got an error reading communication packets)" logged when running `SideBySide.ConnectionPool` tests, but not when running `SideBySide.ConnectionPool.LeakConnections` by itself, even though I suspect (from mapping IP source ports back to originating call stacks) that the problem originates in the `LeakConnections` test.

### DIFF
--- a/src/MySqlConnector/MySqlClient/ConnectionPool.cs
+++ b/src/MySqlConnector/MySqlClient/ConnectionPool.cs
@@ -10,7 +10,7 @@ namespace MySql.Data.MySqlClient
 {
 	internal sealed class ConnectionPool
 	{
-		public async Task<MySqlSession> GetSessionAsync(IOBehavior ioBehavior, CancellationToken cancellationToken)
+		public async Task<MySqlSession> GetSessionAsync(MySqlConnection connection, IOBehavior ioBehavior, CancellationToken cancellationToken)
 		{
 			cancellationToken.ThrowIfCancellationRequested();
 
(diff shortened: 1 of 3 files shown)


caleblloyd:
I thought that when GC ran on `MySqlConnection`, it would run the `Dispose` method, which would in-turn call ReturnToPool.  How are these `MySqlConnection` objects getting GC'd without properly going through `Dispose`?

bgrainger:
The .NET GC does not invoke objects' `Dispose` methods; that has to be done manually by the consumer (typically using a `using` block).

The .NET finalizer will run a finalizable object's destructor (i.e., `~MyObject()`) after the object has been collected. In the typical implementation of the dispose pattern, the finalizer will call `Dispose(false)`. (So in that sense "the GC calls dispose", but in general `Dispose` is not called deterministically, unlike say C++ destructors.)

If the client doesn't call `MySqlConnection.Dispose` (explicitly, or implicitly by using a `using` block) then once the last root reference to it (typically a local variable, maybe a field) is gone, then the whole tree of objects referenced by `MySqlConnection` will be GCed; before this PR, this would have included `MySqlSession`, the payload handler, the `Socket`, `NetworkStream`, etc. and the `SafeHandle`s that wrap the native socket. After the `SafeHandle`s are GCed, they'll be registered for finalization (because they all have finalizers) and that finalizer will close the socket. Eventually the previous code detected that `WeakReference<MySqlSession>` referenced a collected object and removed it from the count of sessions that the pool had.

Now, the pool maintains a hard reference to the `MySqlSession` and a weak reference to the `MySqlConnection`. If the client abandons its reference, the `MySqlConnection` will be collected, as before. However, the `MySqlSession` and all the socket-related objects will remain alive. This lets us recover that session and return it to the pool, and prevents the socket from being closed.

PR #292, opened 2017-07-04
GetCodeRootPath breaks things

Nicholi:
There is a bit of a code smell in AppConfig's startup logic to find the "RootPath" of the executing assembly. Even more problematic is that it is hardcoded to look for the solution dir "MySqlConnector". If someone renames their top level checkout it creates a wonderful infinite loop in testing (took me awhile to notice this was why).

CodeRootPath is used in only three places: to create the path for certificate files, to create the path for test data files, and to set the location of config.json relative to the SideBySide directory. The first two can easily just be pulled from new properties in config.json. And the third, calling SetBasePath on ConfigurationBuilder, is totally un-necessary. We just need to make sure config.json is copied to output directory.

I added the new properties to the config files in ".ci/config" to point to the correct paths on checkout, so all the tests still work as is.

One minor addition I made to differeniate the "server" and "local" bulk file insertion paths, was to use a different property to "expand" the paths %TESTDATA% variable. Note the "server" bulk file insertion isn't even used in the config files, only "local" files. The two properties are "TestData" and "RemoteTestData", fairly obvious. Confirmed that the "server" bulk file insertions still work as well.

### DIFF
--- a/.ci/config/config.compression+ssl.json
+++ b/.ci/config/config.compression+ssl.json
@@ -9,6 +9,6 @@
       "MySqlBulkLoaderLocalCsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.CSV",
       "MySqlBulkLoaderLocalTsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.TSV",
 
-      "CertificatesPath": "..\\..\\..\\..\\..\\.ci\\server\\certs"
+      "CertificatesPath": "../../../../../.ci/server/certs"
     }
 }
--- a/.ci/config/config.compression+ssl.json
+++ b/.ci/config/config.compression+ssl.json
@@ -4,7 +4,11 @@
       "PasswordlessUser": "no_password",
       "SecondaryDatabase": "testdb2",
       "SupportedFeatures": "Json,StoredProcedures,Sha256Password,LargePackets",
+	  
+      "TestData": "../../../../TestData",
       "MySqlBulkLoaderLocalCsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.CSV",
-      "MySqlBulkLoaderLocalTsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.TSV"
+      "MySqlBulkLoaderLocalTsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.TSV",
+	  
+	  "CertificatesPath": "..\\..\\..\\..\\..\\.ci\\server\\certs"
     }
 }
--- a/.ci/config/config.ssl.json
+++ b/.ci/config/config.ssl.json
@@ -9,6 +9,6 @@
       "MySqlBulkLoaderLocalCsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.CSV",
       "MySqlBulkLoaderLocalTsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.TSV",
 
-      "CertificatesPath": "..\\..\\..\\..\\..\\.ci\\server\\certs"
+      "CertificatesPath": "../../../../../.ci/server/certs"
     }
 }
--- a/.ci/config/config.ssl.json
+++ b/.ci/config/config.ssl.json
@@ -4,7 +4,11 @@
       "PasswordlessUser": "no_password",
       "SecondaryDatabase": "testdb2",
       "SupportedFeatures": "Json,StoredProcedures,Sha256Password,LargePackets",
+	  
+      "TestData": "../../../../TestData",
       "MySqlBulkLoaderLocalCsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.CSV",
-      "MySqlBulkLoaderLocalTsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.TSV"
+      "MySqlBulkLoaderLocalTsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.TSV",
+	  
+	  "CertificatesPath": "..\\..\\..\\..\\..\\.ci\\server\\certs"
     }
 }
--- a/.ci/config/config.uds+ssl.json
+++ b/.ci/config/config.uds+ssl.json
@@ -7,6 +7,8 @@
 
       "TestData": "../../../../TestData",
       "MySqlBulkLoaderLocalCsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.CSV",
-      "MySqlBulkLoaderLocalTsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.TSV"
+      "MySqlBulkLoaderLocalTsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.TSV",
+
+      "CertificatesPath": "../../../../../.ci/server/certs"
     }
 }
--- a/.ci/config/config.uds+ssl.json
+++ b/.ci/config/config.uds+ssl.json
@@ -4,6 +4,8 @@
       "PasswordlessUser": "no_password",
       "SecondaryDatabase": "testdb2",
       "SupportedFeatures": "Json,StoredProcedures,Sha256Password,LargePackets",
+	  
+	  "TestData": "../../../../TestData",
       "MySqlBulkLoaderLocalCsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.CSV",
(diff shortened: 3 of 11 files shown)


.ci/config/config.compression+ssl.json
```diff
@@ -4,7 +4,11 @@
       "PasswordlessUser": "no_password",
       "SecondaryDatabase": "testdb2",
       "SupportedFeatures": "Json,StoredProcedures,Sha256Password,LargePackets",
+	  
+      "TestData": "../../../../TestData",
       "MySqlBulkLoaderLocalCsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.CSV",
-      "MySqlBulkLoaderLocalTsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.TSV"
+      "MySqlBulkLoaderLocalTsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.TSV",
+	  
+	  "CertificatesPath": "..\\..\\..\\..\\..\\.ci\\server\\certs"
```
bgrainger: Does this need to use `/` to fix the Travis build? (I haven't tested)

Nicholi: Sort of, I figured safer to just put the logic in AppConfig to normalize it no matter what user inputs.

Should be ok now.

tests/SideBySide/AppConfig.cs
```diff
@@ -74,19 +73,8 @@ public static MySqlConnectionStringBuilder CreateSha256ConnectionStringBuilder()
...
-			var directory = new Uri(currentAssembly.CodeBase).LocalPath;
-			while (!string.Equals(Path.GetFileName(directory), "MySqlConnector", StringComparison.OrdinalIgnoreCase))
-				directory = Path.GetDirectoryName(directory);
-			return directory;
-		}
-
-		private static string ExpandVariables(string value) => value?.Replace("%TESTDATA%", TestDataPath);
+		private static string ExpandVariables(string value, string replacement) => value?.Replace("%TESTDATA%", replacement);
+		private static string ExpandLocalVariables(string value) => ExpandVariables(value, TestDataPath);
+		private static string ExpandRemoteVariables(string value) => ExpandVariables(value, RemoteTestDataPath);
```
bgrainger: If you need to use two variables (and I'm not sure why that's necessary), this should be:

```
private static string ExpandVariables(string value) =>
value?.Replace("%TESTDATA%", TestDataPath)?.Replace("%REMOTETESTDATA%", RemoteTestDataPath);
```

I.e., use two variables; don't overload the meaning of one.

bgrainger: Secondly, I'm not sure there's a real win here for using `%TESTDATA%` as a variable when it's no longer automatically determined by the system, but is instead spelt out in the config file. Might as well just fully hard-code it in the config files in the values that need it.

bgrainger:
> And the third, calling SetBasePath on ConfigurationBuilder, is totally un-necessary. We just need to make sure config.json is copied to output directory.

@caleblloyd Do you remember why it wasn't done this way originally? I'm guessing it might have been a limitation of `project.json` but I can't completely recall.

tests/SideBySide/AppConfig.cs
Nicholi: There should be two because of the way mysql is loading the file. The "local" calls are pulling from the client's machine, and the "server" ones are pulling from the server's machine. Which could be a completely different plaform, so using the same variable doesn't really make sense. As in my case I was testing the "server" files from "/var/lib/mysql-files" and the "local" files" directly from the git checkout.

We don't really need to have %TESTDATA% and %REMOTESTDATA% as they'll never cross paths. I can add it if you want.

caleblloyd:
A couple reasons:

1. I couldn't figure out how to copy the config.json back in `project.json` days
2. When you run `dotnet test tests/SideBySide` from the repo root, the cwd is the repo root but when you run `dotnet test` from the `<repo root>/tests/SideBySide` that is the cwd. Since my solution was using cwd, it needed a way to find a common directory.

Changing it to look for `MySqlConnector.sln` would be a simple fix. We need to reference multiple different files, such as certs in the `.ci/server` folder and unix sockets mounted from Docker. We need a good way to determine repo root.

Nicholi:
@caleblloyd I think trying to determine the repo root is just going to be problematic in the end, as you will need something to use as a sentinel value. Using the .sln file would be somewhat better rather than the solution directory.

As I was just in the case of checking out my own fork to MySqlConnector_mine, and that's when all the tests broke because of GetCodeRootPath.

.ci/config/config.ssl.json
```diff
@@ -4,7 +4,11 @@
       "PasswordlessUser": "no_password",
       "SecondaryDatabase": "testdb2",
       "SupportedFeatures": "Json,StoredProcedures,Sha256Password,LargePackets",
+	  
+      "TestData": "../../../../TestData",
       "MySqlBulkLoaderLocalCsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.CSV",
-      "MySqlBulkLoaderLocalTsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.TSV"
+      "MySqlBulkLoaderLocalTsvFile": "%TESTDATA%/LoadData_UTF8_BOM_Unix.TSV",
+	  
+	  "CertificatesPath": "..\\..\\..\\..\\..\\.ci\\server\\certs"
```
caleblloyd: let's stick to `../` for config file, mixing `..\` is confusing 

also nitpick- could you use the same size indent

.ci/config/config.uds+ssl.json
```diff
@@ -4,6 +4,8 @@
       "PasswordlessUser": "no_password",
       "SecondaryDatabase": "testdb2",
       "SupportedFeatures": "Json,StoredProcedures,Sha256Password,LargePackets",
+	  
+	  "TestData": "../../../../TestData",
```
caleblloyd: nitpick- could you use the same size indent (on all of the config files)

tests/SideBySide/AppConfig.cs
```diff
@@ -19,18 +19,17 @@ public static class AppConfig
 				["Data:SupportsJson"] = "false",
 			};
 
-		private static string CodeRootPath = GetCodeRootPath();
+		public static string CertsPath => Config.GetValue<string>("Data:CertificatesPath")
```
caleblloyd: Just call `Path.GetFullPath(Config.GetValue<string>("Data:CertificatesPath"))` and it'll take care of convention

tests/SideBySide/AppConfig.cs
```diff
@@ -19,18 +19,17 @@ public static class AppConfig
...
+			// normalize path to whatever the system prefers
+			.Replace('\\', Path.DirectorySeparatorChar)
+			.Replace('/', Path.DirectorySeparatorChar);
 
-		public static string BasePath = Path.Combine(CodeRootPath, "tests", "SideBySide");
-
-		public static string CertsPath = Path.Combine(CodeRootPath, ".ci", "server", "certs");
-
-		public static string TestDataPath = Path.Combine(CodeRootPath, "tests", "TestData");
+		public static string TestDataPath => Config.GetValue<string>("Data:TestData");
```
caleblloyd: `Path.GetFullPath(`

tests/SideBySide/AppConfig.cs
```diff
@@ -19,18 +19,17 @@ public static class AppConfig
...
+		public static string RemoteTestDataPath => Config.GetValue<string>("Data:RemoteTestData");
```
caleblloyd: `Path.GetFullPath(`

Nicholi:
Ok, all changes should be in now.

tests/SideBySide/AppConfig.cs
Nicholi: Are we sure this handles the directory separator conventions? Looks like the travis build is failing again now, for possibly the same reason as before.

From what I saw only the CertsPath is ever concateting with something else using Path.Combine. TestData doesn't have this problem.

Add the following back?
```
			// normalize path to whatever the system prefers
			.Replace('\\', Path.DirectorySeparatorChar)
			.Replace('/', Path.DirectorySeparatorChar);
```

caleblloyd: I think it handles the conversion of `../` to `..\` on Windows.  Change all of the CertFile references to `../` in the `config.json` files for consistency and this should be good.

caleblloyd: Backslash is a valid filename character on Linux so that wouldn't get converted. I can literally make a file called `..\test` on linux:

```
touch "..\test"
-rw-r--r--  1 caleb caleb     0 Jul  4 18:58 ..\test
```

PR #316, opened 2017-08-26
Check length of auth method switch payload for detecting old protocol.

kobake:
Single 0xfe byte of the payload means it's an Old Authentication Method Switch Request Packet.
See http://imysql.com/mysql-internal-manual/connection-phase-packets.html

When MySqlConnector of current master version received single 0xfe byte response from the server, it throws FormatException("Read past end of buffer looking for NUL.") that doesn't make sense.

This PR provides a message for suggesting to upgrade password hash format in MySQL user record.


bgrainger:
Thanks for the contribution!

kobake:
😄

Issue #208, opened 2017-03-27
EndOfStreamException in ResetConnectionAsync with Amazon Aurora


sherloqholmes:
When running a lot of queries (SELECT, INSERT, UPDATE, DELETE) at the same moment on the same server we are seeing an increasing amount of EndOfStreamExceptions.

We were able to reproduce this issue on a server with ±100 concurrent users.

**What happend**
1) A load test was running on an endpoint that was doing a simple select on the database.
2) Some other requests were coming in that resulted in an insert operation into the database, these operations failed.

**Stacktrace**
```
System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task):38
 System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task):40
 MySql.Data.ValueTaskExtensions+<>c__DisplayClass0_0`2.<ContinueWith>b__0(Task`1 task):0
 System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke():36
 System.Threading.Tasks.Task.Execute():16
 System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task):38
 System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task):40
 MySql.Data.Serialization.MySqlSession.TryAsyncContinuation(Task`1 task):21
 System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke():36
 System.Threading.Tasks.Task.Execute():16
 System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task):38
 System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task):40
 System.Runtime.CompilerServices.TaskAwaiter`1.GetResult():0
 System.Runtime.CompilerServices.ConfiguredValueTaskAwaitable`1+ConfiguredValueTaskAwaiter.GetResult():30
 MySql.Data.Serialization.MySqlSession+<ResetConnectionAsync>d__23.MoveNext():1323
 System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task):38
 System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task):40
[...]

PR #328, opened 2017-09-13
Try/catch ResetConnectionAsync (works around #208)

bnabholz:
When the ConnectionPool is asked for a session, depending on server version, it will attempt to reauthenticate.  On occasion the server will close the connection and the next read from the stream will be 0 bytes, causing the connection to fail.  This PR wraps ResetConnectionAsync in a try/catch, and if it throws it will fall back to instantiating a new MySqlSession.

I'm not sure if we should be trying to catch only certain exceptions (EndOfStreamException?) and let others bubble up as usual.

### DIFF
--- a/src/MySqlConnector/MySqlClient/ConnectionPool.cs
+++ b/src/MySqlConnector/MySqlClient/ConnectionPool.cs
@@ -47,17 +47,33 @@ namespace MySql.Data.MySqlClient
 					}
 					else
 					{
+						bool resetFailed = false;
+
 						// session is valid, reset if supported
 						if (m_connectionSettings.ConnectionReset)
 						{
-							await session.ResetConnectionAsync(m_connectionSettings, ioBehavior, cancellationToken).ConfigureAwait(false);
+							try
+							{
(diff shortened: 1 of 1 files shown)


bgrainger:
> I'm not sure if we should be trying to catch only certain exceptions (EndOfStreamException?) and let others bubble up as usual.

I generally prefer to catch the specific exception I know I'm handling; however, in this case it seems good to fall back to creating a new session if any kind of `IOException` or `SocketException` is thrown.

bgrainger:
I'm going to update the code I just pushed to only catch `EndOfStreamException`, `IOException`, `SocketException` (same as `TryPingAsync`); I don't want other programming errors to be masked by this `catch` handler (as it would dramatically reduce performance by essentially disabling connection pooling).

bnabholz:
Sounds good to me.  Thanks for reviewing and thanks to you and the other contributors for this project!

bgrainger:
Shipped in [0.26.4](https://www.nuget.org/packages/MySqlConnector/0.26.4); thanks for the contribution!

Issue #307, opened 2017-08-08
Implement GetSchemaTable


Topcoder01:
Dim oFilesToProcessDataTable As New DataTable
 Dim oMySqlCommand As New MySqlCommand
 oFilesToProcessDataTable.Load(oMySqlCommand.ExecuteReader)

The above tosses an "command not implemented" or something very similar to that, I forgot :( 

If this isn't supported is their another easy way to populate a DataTable with this connector?

bgrainger:
The stack trace is:

```
NotSupportedException: Specified method is not supported.
   at MySql.Data.MySqlClient.MySqlDataReader.GetSchemaTable()
   at System.Data.ProviderBase.SchemaMapping..ctor(DataAdapter adapter, DataSet dataset, DataTable datatable, DataReaderContainer dataReader, Boolean keyInfo, SchemaType schemaType, String sourceTableName, Boolean gettingData, DataColumn parentChapterColumn, Object parentChapterValue)
   at System.Data.Common.DataAdapter.FillMappingInternal(DataSet dataset, DataTable datatable, String srcTable, DataReaderContainer dataReader, Int32 schemaCount, DataColumn parentChapterColumn, Object parentChapterValue)
   at System.Data.Common.DataAdapter.FillMapping(DataSet dataset, DataTable datatable, String srcTable, DataReaderContainer dataReader, Int32 schemaCount, DataColumn parentChapterColumn, Object parentChapterValue)
   at System.Data.Common.DataAdapter.FillFromReader(DataSet dataset, DataTable datatable, String srcTable, DataReaderContainer dataReader, Int32 startRecord, Int32 maxRecords, DataColumn parentChapterColumn, Object parentChapterValue)
   at System.Data.Common.DataAdapter.Fill(DataTable[] dataTables, IDataReader dataReader, Int32 startRecord, Int32 maxRecords)
   at System.Data.DataTable.Load(IDataReader reader, LoadOption loadOption, FillErrorEventHandler errorHandler)
```

This may actually be part of #183; if not part of that work, it's closely related.

bgrainger:
Fixed in [0.27.0](https://github.com/mysql-net/MySqlConnector/releases/tag/0.27.0).

PR #332, opened 2017-09-22
Add GetSchemaTable API

elemount:
Related with issue #307 Support GetShemaTable API.


src/MySqlConnector/Serialization/CharacterSet.cs
```diff
@@ -273,4 +273,213 @@ internal enum CharacterSet : ushort
 		Utf8Mb4VietnameseUca900AccentSensitiveCaseSensitive = 300,
 		Utf8Mb4JapaneseUca900AccentSensitiveCaseSensitive = 303,
 	}
+
+	internal static partial class Extensions
+	{
+		public static int MaxLength(this CharacterSet character)
```
bgrainger: Can you merge this into the existing `SerializationUtility.GetBytesPerCharacter`?

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -270,6 +279,100 @@ protected override void Dispose(bool disposing)
...
+				if (col.ColumnType.IsDecimal())
+				{
+					schemaRow[precision] = col.ColumnLength - 2 + ((col.ColumnFlags & ColumnFlags.Unsigned) != 0 ? 1 : 0);
+					schemaRow[scale] = col.Decimals;
+				}
+
+				schemaRow[baseSchemaName] = col.SchemaName;
+				schemaRow[baseTableName] = col.PhysicalTable;
+				schemaRow[baseColumnName] = col.PhysicalName;
+				schemaRow[isUnique] = false;
```
bgrainger: Should this be set from `ColumnFlags.UniqueKey`? It doesn't appear to be covered in the tests to verify that this is the same value as Connector/NET.

src/MySqlConnector/Serialization/ColumnDefinitionPayload.cs
```diff
@@ -35,17 +51,90 @@ public static ColumnDefinitionPayload Create(PayloadData payload)
...
+				PhysicalTable = physicalTable,
+				PhysicalName = physicalName,
+				Decimals = decimals
+			};
+
 		}
 
-		private ColumnDefinitionPayload(string name, CharacterSet characterSet, int columnLength, ColumnType columnType, ColumnFlags columnFlags)
+		public Type GetDataType(
+			bool treadTinyAsBoolean = false,
```
bgrainger: Typo: `treatTinyAsBoolean`

tests/SideBySide/QueryTests.cs
```diff
@@ -795,6 +795,32 @@ public void ReturnDerivedTypes()
...
+		[Fact]
+		public void GetSchemaTable()
+		{
+			var csb = AppConfig.CreateConnectionStringBuilder();
+			using (var connection = new MySqlConnection(csb.ConnectionString))
+			{
+				connection.Open();
+				using (MySqlCommand command = connection.CreateCommand())
+				{
+					command.CommandText = "select Host, ssl_cipher, max_connections, plugin, authentication_string from mysql.user limit 1";
```
bgrainger: If this were moved to DataTypesTests, it could test everything set up in `DataTypesFixture` (https://github.com/mysql-net/MySqlConnector/blob/master/tests/SideBySide/DataTypesFixture.cs), which should provide very broad coverage of all the data types.

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
```diff
@@ -270,6 +279,100 @@ protected override void Dispose(bool disposing)
...
+			columns.Add(isHidden);
+			columns.Add(isLong);
+			columns.Add(isReadOnly);
+
+			for (var i = 0; i < colDefinitions.Length; ++i)
+			{
+				var col = colDefinitions[i];
+				var schemaRow = schemaTable.NewRow();
+				schemaRow[columnName] = col.Name;
+				schemaRow[ordinal] = i;
```
bgrainger: This is incompatible with Connector/NET which [uses 1-based ordinals](https://github.com/mysql/mysql-connector-net/blob/5864e6b21a8b32f5154b53d1610278abb3cb1cee/Source/MySql.Data/extensions/NonRT/MySqlDataReader.cs#L1000) for some reason. That seems like a bug to me, and not sure if we should clone that behaviour for backwards compatibility.

src/MySqlConnector/Serialization/ColumnDefinitionPayload.cs
bgrainger: Seems to be copied/duplicated from `ResultSet.GetFieldType`; that method should probably be refactored to delegate to this one.

src/MySqlConnector/Serialization/ColumnDefinitionPayload.cs
```diff
@@ -35,17 +51,90 @@ public static ColumnDefinitionPayload Create(PayloadData payload)
...
+				case ColumnType.TinyBlob:
+				case ColumnType.Blob:
+				case ColumnType.MediumBlob:
+				case ColumnType.LongBlob:
+					if (CharacterSet == CharacterSet.Binary && oldGuids && ColumnLength == 16)
+					{
+						return typeof(Guid);
+					}
+
+					return typeof(string);
```
bgrainger: Seems like this should return `typeof(byte[])` when `CharacterSet == CharacterSet.Binary`? Please increase test coverage.

bgrainger:
Here's the skeleton of some sample tests that could be added to `DataTypes.cs`:

```
#if !NETCOREAPP1_1_2
		[Theory]
		[InlineData("rowid", "datatypes_bools", 11, typeof(int), false, true, false, 0, 0)]
		[InlineData("Boolean", "datatypes_bools", 1, typeof(bool), false, false, true, 0, 0)]
		[InlineData("SByte", "datatypes_integers", 4, typeof(sbyte), false, false, true, 0, 0)]
		[InlineData("Int16", "datatypes_integers", 6, typeof(short), false, false, true, 0, 0)]
		[InlineData("UInt16", "datatypes_integers", 5, typeof(ushort), false, false, true, 0, 0)]
		[InlineData("Double", "datatypes_reals", 22, typeof(double), false, false, true, 0, 31)]
		[InlineData("MediumDecimal", "datatypes_reals", 30, typeof(decimal), false, false, true, 28, 8)]
		[InlineData("utf8", "datatypes_strings", 300, typeof(string), false, false, true, 0, 0)]
		[InlineData("guid", "datatypes_strings", 144, typeof(Guid), false, false, true, 0, 0)]
		[InlineData("guidbin", "datatypes_strings", 144, typeof(Guid), false, false, true, 0, 0)]
		[InlineData("Blob", "datatypes_blobs", 16383, typeof(byte[]), true, false, true, 0, 0)] // 16383 is almost certainly a bug; should be 65535
		[InlineData("Date", "datatypes_times", 10, typeof(DateTime), false, false, true, 0, 0)]
		[InlineData("Time", "datatypes_times", 17, typeof(TimeSpan), false, false, true, 0, 6)]
		public void GetSchemaTable(string column, string table, int columnSize, Type dataType, bool isLong, bool isKey, bool allowDbNull, int precision, int scale)
		{
			using (var command = m_database.Connection.CreateCommand())
			{
				command.CommandText = $"select 1, `{column}` from `{table}`;";
				using (var reader = command.ExecuteReader())
				{
					var schema = reader.GetSchemaTable().Rows[1];
					Assert.Equal(column, schema["ColumnName"]);
#if BASELINE
					int ordinal = 2; // https://bugs.mysql.com/bug.php?id=61477
#else
					int ordinal = 1;
#endif
					Assert.Equal(ordinal, schema["ColumnOrdinal"]);
					Assert.Equal(columnSize, schema["ColumnSize"]);
					Assert.Equal(dataType, schema["DataType"]);
					Assert.Equal(isLong, schema["IsLong"]);
					Assert.Equal(isKey, schema["IsKey"]);
					Assert.Equal(allowDbNull, schema["AllowDBNull"]);
					Assert.Equal(precision, schema["NumericPrecision"]);
					Assert.Equal(scale, schema["NumericScale"]);
					Assert.Equal(m_database.Connection.Database, schema["BaseSchemaName"]);
					Assert.Equal(table, schema["BaseTableName"]);
					Assert.Equal(column, schema["BaseColumnName"]);
					Assert.False((bool) schema["IsUnique"]);
					Assert.False((bool) schema["IsRowVersion"]);
					Assert.False((bool) schema["IsReadOnly"]);
				}
			}
		}
#endif
```

They currently pass against Connector/NET, but not this PR. Each difference would need to be investigated to determine whether this implementation is an improvement (e.g., starting at 0-based columns, which fixes https://bugs.mysql.com/bug.php?id=61477) or a regression.

We need to have a compelling rationale for each incompatibility with Connector/NET.

bgrainger:
(And the tests should be expanded to cover every possible data type, not just the dozen or so I put in there as an example.)

src/MySqlConnector/Serialization/CharacterSet.cs
bgrainger: I've pushed a change to `SerializationUtility.GetBytesPerCharacter` so this should be removed in favour of that: 83c1cea6f22ac4547e157a2eb93fccc85e6850a5

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
elemount: Oracle set UniqueKey to be `false` always, what's your idea on this backwards compatibility?

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
elemount: Let me fix it for backwards compatibility.

src/MySqlConnector/Serialization/CharacterSet.cs
elemount: Sure.

src/MySqlConnector/Serialization/ColumnDefinitionPayload.cs
elemount: Fixed by use `ResultSet.GetFieldType`

src/MySqlConnector/Serialization/ColumnDefinitionPayload.cs
elemount: Fixed by use `ResultSet.GetFieldType`

tests/SideBySide/QueryTests.cs
elemount: Fixed.

src/MySqlConnector/MySqlClient/MySqlDataReader.cs
bgrainger: I'm OK with keeping it hardcoded to `false`; it's easier and backwards-compatible.

tests/SideBySide/DataTypesFixture.cs
```diff
@@ -121,8 +121,8 @@ public DataTypesFixture()
   latin1 varchar(300) character set 'latin1' null,
   latin1bin varchar(300) character set latin1 collate latin1_bin null,
   cp1251 varchar(300) character set 'cp1251' null,
-  guid char(36) null,
-  guidbin char(36) binary null
+  guid char(36) character set 'utf8mb4' null,
+  guidbin char(36) binary character set 'utf8mb4' null
```
bgrainger: Why was this change needed?

bgrainger: OK, I see that the "baseline" code returns 108 for this column length (== 36 * 3, so probably defaulting to `utf8`?). IMO, this is a bug in Oracle's connector; the `ColumnSize` of a `char(36)` column should be reported as 36, no matter the encoding. 

MariaDB's failure is interesting... I wonder if you could work around them both for now by using `character set 'ascii'` (and an expected column size of 36)?

src/MySqlConnector/MySqlClient/MySqlConnectionStringBuilder.cs
```diff
@@ -431,7 +431,7 @@ static MySqlConnectionStringOption()
 
 			AddOption(UseCompression = new MySqlConnectionStringOption<bool>(
 				keys: new[] { "Compress", "Use Compression", "UseCompression" },
-				defaultValue: false));
+				defaultValue: true));
```
bgrainger: ?

tests/SideBySide/AppConfig.cs
```diff
@@ -34,7 +34,7 @@ public static class AppConfig
 
 		public static ServerFeatures SupportedFeatures => (ServerFeatures) Enum.Parse(typeof(ServerFeatures), Config.GetValue<string>("Data:SupportedFeatures"));
 
-		public static bool SupportsJson => SupportedFeatures.HasFlag(ServerFeatures.Json);
+		public static bool SupportsJson => false;
```
bgrainger: ?

src/MySqlConnector/MySqlClient/MySqlConnectionStringBuilder.cs
elemount: false commit, please ignore it.

tests/SideBySide/AppConfig.cs
elemount: False commit, please ignore it.

tests/SideBySide/DataTypesFixture.cs
elemount: I read the code and it is a special case for GUID. I decide to do something special for GUID length which is different with Orcale's connector, what's your idea?

tests/SideBySide/DataTypes.cs
```diff
@@ -624,8 +624,8 @@ public void GetSchemaTable(string column, string table, int columnSize, Type dat
 					Assert.Equal(column, schema["ColumnName"]);
 					int ordinal = 2; // https://bugs.mysql.com/bug.php?id=61477
 					Assert.Equal(ordinal, schema["ColumnOrdinal"]);
-					Assert.Equal(columnSize, schema["ColumnSize"]);
 					Assert.Equal(dataType, schema["DataType"]);
+					Assert.Equal(columnSize, schema["ColumnSize"]);
```
bgrainger: Appveyor runs the tests against this library, then against MySql.Data (to check for backwards compatibility). In this case, it would be OK to add a `#if BASELINE` check for the cases we know Oracle's connector has bugs (GUIDs and BLOBs).

bgrainger:
Don't worry about codacy "failure"; I was interested in seeing what it could do, so I added it to this project (but didn't realise it would analyse every current PR).

Issue #286, opened 2017-06-27
Add server's RSA public key connection string option


bgrainger:
If the client connects with `sha256_password` over an insecure connection, it has to encrypt the user's password with the server's RSA public key. This can be requested from the server [in authentication negotiation](http://web.archive.org/web/20170405002245/https://dev.mysql.com/doc/internals/en/public-key-retrieval.html).

However, as [Protecting MySQL Passwords With the sha256_password Plugin](http://mysqlserverteam.com/protecting-mysql-passwords-with-the-sha256_password-plugin/) says:

> Distributing keys securely can be an operational headache. MySQL Server will supply its own RSA public key upon request from the client, so that the key doesn’t have to be explicitly distributed and configured for each client. But this introduces another security concern – a proxy in the middle may substitute an RSA public key for which it has the private key, decrypt and harvest the plain-text password, then re-encrypting the password with the actual server RSA public key for the connection attempt to continue. For this reason, it’s strongly recommended that clients define a local RSA public key to use instead of request the server RSA key during the handshake.

The `mysql` CLI supports specifying the server's RSA public key with the [`--server-public-key-path`](https://dev.mysql.com/doc/refman/5.7/en/mysql-command-options.html#option_mysql_server-public-key-path) variable.

[Connector/J](https://dev.mysql.com/doc/connectors/en/connector-j-reference-configuration-properties.html) has `allowPublicKeyRetrieval` (default `false`) and `serverRSAPublicKeyFile` properties.

We should add both these connection string options for RSA public key support.

bgrainger:
Setting milestone to 1.0 as adding `AllowPublicKeyRetrieval` (with a sensible default of `false`) will be a breaking change.

bgrainger:
Added in [0.24.0](https://github.com/mysql-net/MySqlConnector/releases/tag/0.24.0).

PR #345, opened 2017-10-03
Implement caching_sha2_password support.

Drake103:
Implemented caching_sha2_password support according to following sources:
1. https://dev.mysql.com/doc/refman/8.0/en/caching-sha2-pluggable-authentication.html
2. https://dev.mysql.com/worklog/task/?id=9591

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,13 @@ env:
     FEATURES=StoredProcedures
   - IMAGE=mysql:5.7
     NAME=mysql
-    FEATURES=Json,StoredProcedures,Sha256Password,LargePackets,CachingSha2Password
+    FEATURES=Json,StoredProcedures,Sha256Password,LargePackets
   - IMAGE=mysql:8.0
     NAME=mysql
     FEATURES=Json,StoredProcedures,Sha256Password,LargePackets,CachingSha2Password
   - IMAGE=percona:5.7
     NAME=percona
-    FEATURES=Json,StoredProcedures,Sha256Password,OpenSsl,LargePackets,CachingSha2Password
+    FEATURES=Json,StoredProcedures,Sha256Password,OpenSsl,LargePackets
   - IMAGE=mariadb:10.2
     NAME=mariadb
     FEATURES=StoredProcedures,OpenSsl,LargePackets
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,13 @@ env:
     FEATURES=StoredProcedures
   - IMAGE=mysql:5.7
     NAME=mysql
-    FEATURES=Json,StoredProcedures,Sha256Password,LargePackets
+    FEATURES=Json,StoredProcedures,Sha256Password,LargePackets,CachingSha2Password
   - IMAGE=mysql:8.0
     NAME=mysql
-    FEATURES=Json,StoredProcedures,Sha256Password,LargePackets
+    FEATURES=Json,StoredProcedures,Sha256Password,LargePackets,CachingSha2Password
   - IMAGE=percona:5.7
     NAME=percona
-    FEATURES=Json,StoredProcedures,Sha256Password,OpenSsl,LargePackets
+    FEATURES=Json,StoredProcedures,Sha256Password,OpenSsl,LargePackets,CachingSha2Password
   - IMAGE=mariadb:10.2
     NAME=mariadb
     FEATURES=StoredProcedures,OpenSsl,LargePackets
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,7 @@ before_test:
 - cmd: |-
     "C:\Program Files\MySQL\MySQL Server 5.7\bin\mysql.exe" -u root --password=Password12! < .ci\server\init.sql
     "C:\Program Files\MySQL\MySQL Server 5.7\bin\mysql.exe" -u root --password=Password12! < .ci\server\init_sha256.sql
+    "C:\Program Files\MySQL\MySQL Server 5.7\bin\mysql.exe" -u root --password=Password12! < .ci\server\init_caching_sha2.sql
 test_script:
 - ps: .\.ci\test.ps1
 after_test:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,7 +7,6 @@ before_test:
 - cmd: |-
     "C:\Program Files\MySQL\MySQL Server 5.7\bin\mysql.exe" -u root --password=Password12! < .ci\server\init.sql
     "C:\Program Files\MySQL\MySQL Server 5.7\bin\mysql.exe" -u root --password=Password12! < .ci\server\init_sha256.sql
-    "C:\Program Files\MySQL\MySQL Server 5.7\bin\mysql.exe" -u root --password=Password12! < .ci\server\init_caching_sha2.sql
 test_script:
 - ps: .\.ci\test.ps1
 after_test:
--- a/src/MySqlConnector/Serialization/MySqlSession.cs
+++ b/src/MySqlConnector/Serialization/MySqlSession.cs
@@ -216,7 +216,7 @@ namespace MySql.Data.Serialization
 				authPluginName = initialHandshake.AuthPluginName;
 			else
 				authPluginName = (initialHandshake.ProtocolCapabilities & ProtocolCapabilities.SecureConnection) == 0 ? "mysql_old_password" : "mysql_native_password";
-			if (authPluginName != "mysql_native_password" && authPluginName != "sha256_password")
+			if (authPluginName != "mysql_native_password" && authPluginName != "sha256_password" && authPluginName != "caching_sha2_password")
 				throw new NotSupportedException("Authentication method '{0}' is not supported.".FormatInvariant(initialHandshake.AuthPluginName));
 
 			ServerVersion = new ServerVersion(Encoding.ASCII.GetString(initialHandshake.ServerVersion));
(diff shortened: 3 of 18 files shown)


src/MySqlConnector/Serialization/MySqlSession.cs
```diff
@@ -397,6 +370,112 @@ public void FinishQuerying()
...
+				// request the RSA public key
+				await SendReplyAsync(new PayloadData(new ArraySegment<byte>(new byte[] { 0x02 }, 0, 1)), ioBehavior,
+					cancellationToken).ConfigureAwait(false);
+				var payload = await ReceiveReplyAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
+				var publicKeyPayload = AuthenticationMoreDataPayload.Create(payload);
+				return Encoding.ASCII.GetString(publicKeyPayload.Data);
+			}
+
+			throw new MySqlException(
+				"Authentication method '{0}' failed. Either use a secure connection, specify the server's RSA public key with ServerRSAPublicKeyFile, or set AllowPublicKeyRetrieval=True."
```
bgrainger: Does using `ServerRSAPublicKeyFile` work for `caching_sha256_password`? This method doesn't appear to load from the file specified by that connection string setting, but the exception message still mentions it.

bgrainger: (Does MySQL Server use the same RSA public key for `sha256_password` and `caching_sha2_password` or can it support a different public/private keypair for each authentication method?)

Drake103: It can support different key-pair for each method:
- `caching_sha2_password_private_key_path`, `caching_sha2_password_public_key_path` for `caching_sha2_password`;
- `sha256_password_private_key_path`, `sha256_password_public_key_path` for `sha256_password`.

Drake103: >Does using ServerRSAPublicKeyFile work for caching_sha2_password? This method doesn't appear to load from the file specified by that connection string setting, but the exception message still mentions it.

I removed support of ServerRSAPublicKeyFile, but forgot to fix exception text. Should I add another specific option for the file or fix exception message?

bgrainger: Please remove it from the exception message for now.

Long-term, it seems like another connection string option should be provided, but it could be good to wait to see what other connector libraries do (for consistency). For example, Connector/J doesn't seem to have a different option in [its configuration properties](https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-configuration-properties.html) yet.

bgrainger: Actually (after thinking about it more), from the client side there doesn't seem to be any reason we couldn't use `ServerRSAPublicKeyFile` for both `sha256_password` and `caching_sha2_password`. Presumably when the user creates the connection string (with the user ID and password) they know which auth plugin the server is using, and which RSA private key it's set up with. For any individual connection, there's no need to specify both the `sha256_password` public key *and* the `caching_sha2_password` public key. Thus, the same connection string option can be used for both. More specifically, `ServerRSAPublicKeyFile` can continue to be used in this new code path to specify the `caching_sha2_password` public key file.

.travis.yml
```diff
@@ -8,13 +8,13 @@ env:
     FEATURES=StoredProcedures
   - IMAGE=mysql:5.7
     NAME=mysql
-    FEATURES=Json,StoredProcedures,Sha256Password,LargePackets
+    FEATURES=Json,StoredProcedures,Sha256Password,LargePackets,CachingSha2Password
```
bgrainger: This seems incorrect, since `CachingSha2Password` is new in MySQL 8.0.

.travis.yml
```diff
@@ -8,13 +8,13 @@ env:
...
   - IMAGE=mysql:8.0
     NAME=mysql
-    FEATURES=Json,StoredProcedures,Sha256Password,LargePackets
+    FEATURES=Json,StoredProcedures,Sha256Password,LargePackets,CachingSha2Password
   - IMAGE=percona:5.7
     NAME=percona
-    FEATURES=Json,StoredProcedures,Sha256Password,OpenSsl,LargePackets
+    FEATURES=Json,StoredProcedures,Sha256Password,OpenSsl,LargePackets,CachingSha2Password
```
bgrainger: Also seems wrong.

appveyor.yml
```diff
@@ -7,6 +7,7 @@ before_test:
 - cmd: |-
     "C:\Program Files\MySQL\MySQL Server 5.7\bin\mysql.exe" -u root --password=Password12! < .ci\server\init.sql
     "C:\Program Files\MySQL\MySQL Server 5.7\bin\mysql.exe" -u root --password=Password12! < .ci\server\init_sha256.sql
+    "C:\Program Files\MySQL\MySQL Server 5.7\bin\mysql.exe" -u root --password=Password12! < .ci\server\init_caching_sha2.sql
```
bgrainger: AppVeyor is running MySQL Server 5.7, which won't support `caching_sha2_password`.

tests/SideBySide/ConnectAsync.cs
```diff
@@ -258,7 +258,7 @@ public ConnectAsync(DatabaseFixture database)
 			using (var connection = new MySqlConnection(csb.ConnectionString))
 			{
 #if BASELINE || NET45
-				await Assert.ThrowsAsync<NotImplementedException>(() => connection.OpenAsync());
+				await Assert.ThrowsAsync<MySqlException>(() => connection.OpenAsync());
```
bgrainger: I would expect the "baseline" library (Oracle's Connector/NET) to continue to throw `NotImplementedException` here. 

I wonder if the `|| NET45` check is out of date, because the SideBySide project builds against `net452`, so it's probably never getting into this block.

src/MySqlConnector/Serialization/MySqlSession.cs
Drake103: I think in `caching_sha2_password` scenario there are 2 cases:
1. Server has RSA keys, and always can provide them (does not require OpenSSL);
2. Server has not any RSA keys, so there is no point in specifying public key at client-side.

bgrainger: The point of specifying them client-side is to prevent a MITM attack; see #286.

Drake103: Thank you. I will fix it by restoring `ServerRSAPublicKeyFile` support.

src/MySqlConnector/Serialization/MySqlSession.cs
```diff
@@ -397,6 +370,125 @@ public void FinishQuerying()
...
+
+			using (rsa)
+			{
+				// XOR the password bytes with the challenge
+				AuthPluginData = Utility.TrimZeroByte(switchRequest.Data);
+				for (var i = 0; i < passwordBytes.Length; i++)
+					passwordBytes[i] ^= AuthPluginData[i % AuthPluginData.Length];
+
+				// encrypt with RSA public key
+				var encryptedPassword = rsa.Encrypt(passwordBytes, RSAEncryptionPadding.Pkcs1);
```
Drake103: I have changed RSA padding from OaepSHA1 to Pkcs1, it may have caused this failed test `SideBySide.ConnectAsync.Sha256WithoutSecureConnection [FAIL]` (https://travis-ci.org/mysql-net/MySqlConnector/jobs/283268492).

bgrainger: The [MySQL documentation for `sha256_password`](https://web.archive.org/web/20170404163410/https://dev.mysql.com/doc/internals/en/sha256.html) says:

> RSA encrypts that with the server's public key using `RSA_PKCS1_OAEP_PADDING` as padding

I found that I needed to use `RSAEncryptionPadding.OaepSHA1` to be compatible.

Drake103: It didn't work with `caching_sha2_password` so I have switched it and all tests passed locally.

bgrainger: As an experiment, I changed the padding to be different for each auth method and the tests passed: fb884aa5080c248e33f840f455a45c52d96279b1

However, I'm unsettled that it would need to be different for each case, and I'd like to find proof from the documentation (or source code) before committing to that.

bgrainger: It is actually different for the two authentication methods.

`sha256_password` uses `RSA_PKCS1_OAEP_PADDING`: https://github.com/mysql/mysql-server/blob/020025d9f277fd3283b8c75024f590857f58fbbd/sql-common/client_authentication.cc#L267

`caching_sha2_password` uses `RSA_PKCS1_PADDING`: https://github.com/mysql/mysql-server/blob/020025d9f277fd3283b8c75024f590857f58fbbd/sql-common/client_authentication.cc?utf8=%E2%9C%93#L497

bgrainger: Or on the server-side, https://github.com/mysql/mysql-server/blob/020025d9f277fd3283b8c75024f590857f58fbbd/sql/auth/sql_authentication.cc?utf8=%E2%9C%93#L3245 vs https://github.com/mysql/mysql-server/blob/020025d9f277fd3283b8c75024f590857f58fbbd/sql/auth/sha2_password.cc?utf8=%E2%9C%93#L1081

Drake103: That was unexpected. Should I push the fix?

bgrainger: Yes please; I think that's the only thing blocking this PR.

Issue #334, opened 2017-09-22
Information disclosure vulnerability with LOAD DATA LOCAL INFILE


bgrainger:
A malicious server could respond with `0xFB C:\local\filename.ext`  to any `COM_QUERY` request, and MySqlConnector would transmit that file to the server. See https://dev.mysql.com/doc/refman/5.7/en/load-data-local.html for a full description.

The simplest mitigation would be to obfuscate all source file names in `MySqlBulkLoader.LoadAsync`; that is, if a file name is specified, open it for reading then treat it as a `SourceStream`. In `ResultSet.ReadResultSetHeaderAsync`, only allow "file names" that begin with `MySqlBulkLoader.StreamPrefix`. This would prohibit arbitrary filesystem paths from being requested.

This would break clients who manually construct a `LOAD DATA LOCAL INFILE` SQL statement and execute it. However [this seems to be rare](https://github.com/search?l=C%23&p=5&q=%22load+data+local+infile%22&type=Code&utf8=%E2%9C%93), and has a fairly simple workaround: switch to `MySqlBulkLoader`. We could also relax the "path must start with `MySqlBulkLoader.StreamPrefix`" requirement if `SslMode` is VerifyCA (or higher), as that implies we can trust the server; if we did that, another workaround would be to use a SSL certificate to connect to the server.

bgrainger:
Fixed in [0.29.0](https://github.com/mysql-net/MySqlConnector/releases/tag/0.29.0).

PR #357, opened 2017-10-11
Treat File as SourceStream in case of LOAD DATA LOCAL INFILE

Drake103:
I have implemented some logic according to #334:
 - In case of Local File create `FileStream` for the file and use it as `SourceStream`.
 - If the host is not trusted and `localInfile` does not start with StreamPrefix, throw `NotSupportedException`.

### DIFF
--- a/src/MySqlConnector/MySqlClient/MySqlBulkLoader.cs
+++ b/src/MySqlConnector/MySqlClient/MySqlBulkLoader.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Data;
 using System.IO;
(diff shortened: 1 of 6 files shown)


src/MySqlConnector/MySqlClient/MySqlBulkLoader.cs
```diff
@@ -196,7 +203,24 @@ private string BuildSqlCommand()
             }
         }
 
-        internal const string StreamPrefix = ":STREAM:";
+		private Stream CreateFileStream(string fileName)
+		{
+			try
+			{
+				return new FileStream(fileName, FileMode.Open);
```
bgrainger: Use `File.OpenRead` because it specifies good defaults for `FileAccess` and `FileShare`.

src/MySqlConnector/MySqlClient/MySqlBulkLoader.cs
```diff
@@ -153,15 +153,22 @@ private string BuildSqlCommand()
             if (!string.IsNullOrWhiteSpace(FileName) && SourceStream != null)
                 throw new InvalidOperationException("Cannot set both FileName and SourceStream");
 
-            if (string.IsNullOrWhiteSpace(FileName) && SourceStream != null)
-            {
-                if (!Local)
-                    throw new InvalidOperationException("Cannot use SourceStream when Local is not true.");
+			// LOCAL INFILE case
+			if (Local && !string.IsNullOrWhiteSpace(FileName) && SourceStream == null)
```
bgrainger: We know that `SourceStream == null` due to the test on line 153; it's redundant to check it again here.

Drake103:
Some jobs have failed:
 - https://travis-ci.org/mysql-net/MySqlConnector/jobs/286737226
 - https://travis-ci.org/mysql-net/MySqlConnector/builds/286731935
Looks like `SideBySide.CancelTests.CancelSlowQueryWithTokenAfterExecuteReader` sometimes is slower than expected.

bgrainger:
I've restarted the spurious failures; looks like Appveyor might have a legitimate problem.

tests/SideBySide/LoadDataInfileAsync.cs
```diff
@@ -42,17 +42,35 @@ public LoadDataInfileAsync(DatabaseFixture database)
...
+			if (m_database.Connection.State != ConnectionState.Open)
+				await m_database.Connection.OpenAsync();
+
 			int rowCount = await command.ExecuteNonQueryAsync();
+
 			m_database.Connection.Close();
 			Assert.Equal(20, rowCount);
 		}
 
+		[BulkLoaderLocalCsvFileFact(TrustedHost = false)]
```
bgrainger: You can put `#if !BASELINE` around this whole test, since this is a deliberate breaking change with the "baseline" connector (i.e., Oracle's Connector/NET) so we're not worried about testing backwards compatibility.

Drake103: I have missed that. Thank you!

bgrainger:
Thank you!

Looks like the tests will pass once I restart that failing Travis job.

Do you want to squash these commits down or should I take them as-is?

Drake103:
I will squash them, one moment.

bgrainger:
I don't know why that "Cancel" test has decided to start intermittently failing just for this PR. (It seems to have been fine for work I've done recently.) Restarting tests.

PR #358, opened 2017-10-13
Rewrite test skipping attributes

bgrainger:
Refactor the test attribute classes down to two common implementations: `[SkippableFact]` and `[SkippableTheory]`.

Reduce duplicated code and also eliminate some `#if BASELINE` blocks.


bgrainger:
@caleblloyd You wrote the first versions of the test attributes; thoughts on the redesign?

caleblloyd:
This is a nice way of doing things! Cuts down on the number of attributes drastically

PR #367, opened 2017-10-20
Remove not implemented

Naragato:
This removes not implemented exceptions being thrown when using certain properties.  The reason being OrmLite from servicestack sets some of these properties, although it does not use them.

It also contained a fix (14b38e342eafd3fd98bbfa20189ce465468ebdc4 [14b38e3]) and a test for issues when reading column defs, but this has been merged away and was fixed in the commit (6b7b3f8dc2b5d4aeb144df2f4b6fa9f60218d49a [6b7b3f8]).

I have left the changes to the test to catch any issues surrounding this thou :).

### DIFF
--- a/.ci/config/config.buffer.json
+++ b/.ci/config/config.buffer.json
@@ -1,6 +1,6 @@
 {
     "Data": {
-      "ConnectionString": "server=127.0.0.1;user id=mysqltest;password='test;key=\"val';port=3306;database=mysqltest;ssl mode=none;Use Affected Rows=true;BufferResultSets=true",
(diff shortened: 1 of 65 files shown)


bgrainger:
This PR was failing so I reimplemented the fix here: ebd35aa7042ccb5b1472160dd6194a412bebf562

I also saw that you added support for MySqlConnector in OrmLite; thanks!

And then I saw this hack, which I wanted to help you get rid of :smile: 
https://github.com/ServiceStack/ServiceStack.OrmLite/blob/1263792/src/ServiceStack.OrmLite/Expressions/SqlExpression.cs#L2420-L2423

bgrainger:
Fixed in [0.29.1](https://github.com/mysql-net/MySqlConnector/releases/tag/0.29.1).

Naragato:
Good stuff :)

PR #373, opened 2017-10-22
Update docs to .NET Core 2.0

dustinsoftware:
Disclaimer: I did not build the whole sample app with these changes, only enough to get a sample app working and attempting to open a connection.


dustinsoftware:
A few notes:

- It seemed like a lot of the documentation was copying the scaffolding from generating a new project. That didn't seem necessary to me, but I can add it back if it is needed
- It doesn't look like it is possible to access the config directly from a service anymore, so I am now injecting the configuration in when registering the service as a transient

PR #391, opened 2017-11-05
Implement server load balancing

### LINKED ISSUE #226: Behavior of Server connection string option
What strategy should the connector use when the `Server` connection string option contains multiple comma-delimited host names?

* Ordered: always try the first, fall back to the second, then the third
* Random: pick one at random
* Round-robin: use a global synchronized index (that wraps around) and cycle through them in order

The behaviour doesn't seem to be specified in the MySQL documentation, nor described in our docs. I believe we currently do "Ordered". Do we want to provide a guarantee; if so, which behavior is most appropriate?

(I haven't checked the Connector/NET source code to see what it might have done back when it supported comma-delimited host names.)

Related to #225 

CC @caleblloyd

bgrainger:
Fixes #226.

Implements a `LoadBalance` connection string option with four settings:

* `RoundRobin` (default) - each server is tried in turn across all calls to `Open`
* `InOrder` - servers are tried in order, starting from the first, for each call to `Open`
* `Random` - servers are tried randomly
* `FewestConnections` - a connection is opened to the server with the fewest open connections in this connection pool (requires `Pooling=true`)

### DIFF
--- a/src/MySqlConnector/MySqlClient/ConnectionPool.cs
+++ b/src/MySqlConnector/MySqlClient/ConnectionPool.cs
@@ -318,11 +318,11 @@ namespace MySql.Data.MySqlClient
 			if (cs.LoadBalance == MySqlLoadBalance.FewestConnections)
 			{
 				m_hostSessions = new Dictionary<string, int>();
-				foreach (var hostName in cs.Hostnames)
+				foreach (var hostName in cs.HostNames)
 					m_hostSessions[hostName] = 0;
 			}
 			m_loadBalancer = cs.ConnectionType != ConnectionType.Tcp ? null :
-				cs.Hostnames.Count == 1 || cs.LoadBalance == MySqlLoadBalance.InOrder ? InOrderLoadBalancer.Instance :
+				cs.HostNames.Count == 1 || cs.LoadBalance == MySqlLoadBalance.InOrder ? InOrderLoadBalancer.Instance :
 				cs.LoadBalance == MySqlLoadBalance.Random ? RandomLoadBalancer.Instance :
 				cs.LoadBalance == MySqlLoadBalance.FewestConnections ? new FewestConnectionsLoadBalancer(this) :
 				(ILoadBalancer) new RoundRobinLoadBalancer();
--- a/src/MySqlConnector/MySqlClient/ConnectionPool.cs
+++ b/src/MySqlConnector/MySqlClient/ConnectionPool.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using MySql.Data.Protocol.Serialization;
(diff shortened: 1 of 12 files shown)


src/MySqlConnector/MySqlClient/ConnectionPool.cs
```diff
@@ -306,6 +315,35 @@ private ConnectionPool(ConnectionSettings cs)
...
 			m_sessions = new LinkedList<MySqlSession>();
 			m_leasedSessions = new Dictionary<int, MySqlSession>();
+			if (cs.LoadBalance == MySqlLoadBalance.FewestConnections)
+			{
+				m_hostSessions = new Dictionary<string, int>();
+				foreach (var hostName in cs.Hostnames)
+					m_hostSessions[hostName] = 0;
+			}
+			m_loadBalancer = cs.ConnectionType != ConnectionType.Tcp ? null :
+				cs.Hostnames.Count == 1 || cs.LoadBalance == MySqlLoadBalance.InOrder ? InOrderLoadBalancer.Instance	:
```
ejball: Tab should be space.

src/MySqlConnector/MySqlClient/ConnectionPool.cs
```diff
@@ -306,6 +315,35 @@ private ConnectionPool(ConnectionSettings cs)
...
+				cs.LoadBalance == MySqlLoadBalance.FewestConnections ? new FewestConnectionsLoadBalancer(this) :
+				(ILoadBalancer) new RoundRobinLoadBalancer();
+		}
+
+		private void AdjustHostConnectionCount(MySqlSession session, int delta)
+		{
+			if (m_hostSessions != null)
+			{
+				lock (m_hostSessions)
+					m_hostSessions[session.HostName] = m_hostSessions[session.HostName] + delta;
```
ejball: FWIW, `m_hostSessions[session.HostName] += delta` works fine.

src/MySqlConnector/Serialization/ILoadBalancer.cs
```diff
@@ -0,0 +1,74 @@
+using System;
+using System.Collections.Generic;
+
+namespace MySql.Data.Serialization
+{
+    internal interface ILoadBalancer
```
ejball: Indentation.

src/MySqlConnector/Serialization/ILoadBalancer.cs
```diff
@@ -0,0 +1,74 @@
...
+		public static ILoadBalancer Instance { get; } = new InOrderLoadBalancer();
+
+		public IReadOnlyList<string> LoadBalance(IReadOnlyList<string> hosts) => hosts;
+	}
+
+	internal sealed class RandomLoadBalancer : ILoadBalancer
+	{
+		public static ILoadBalancer Instance { get; } = new RandomLoadBalancer();
+
+		public RandomLoadBalancer()
```
ejball: Could be `private`.

PR #401, opened 2017-11-21
Make Enum params behave like official client.

Naragato:
Enums in the official client are serialized to string values.

https://github.com/mysql/mysql-connector-net/blob/5864e6b21a8b32f5154b53d1610278abb3cb1cee/Source/MySql.Data/extensions/NonRT/MySqlParameter.cs

https://dev.mysql.com/doc/dev/connector-net/6.10/html/T_MySql_Data_MySqlClient_MySqlDbType.htm

### DIFF
--- a/src/MySqlConnector/MySql.Data.MySqlClient/MySqlParameter.cs
+++ b/src/MySqlConnector/MySql.Data.MySqlClient/MySqlParameter.cs
@@ -226,7 +226,7 @@ namespace MySql.Data.MySqlClient
 			}
 			else if (Value is Enum)
 			{
-				writer.WriteUtf8("{0:d}".FormatInvariant(Value));
+				writer.WriteUtf8("'{0:G}'".FormatInvariant(Value));
 			}
 			else
 			{


Naragato:
Sorry I didn't adjust the unit tests, I will do at the weekend.

bgrainger:
> Enums in the official client are serialized to string values.

I don't believe this is true, and the file you linked to doesn't confirm this claim. The tests added in 813a61e032e01a655bfe8f8f34974563600e47ce run against Connector/NET and produce the same behaviour as this library.

Do you have a packet capture (from Connector/NET) that shows them being serialised as strings on the wire?

bgrainger:
This code in Connector/NET maps `Enum` types to `Int32` (which causes bug [#84701](https://bugs.mysql.com/bug.php?id=84701)): https://github.com/mysql/mysql-connector-net/blob/5864e6b21a8b32f5154b53d1610278abb3cb1cee/Source/MySql.Data/parameter.cs#L311-L316

Closing this PR as it would introduce a breaking change from Connector/NET without a compelling reason.

Naragato:
@bgrainger you are partially write, sorry for the rushed pull request.  I won't be able to look in full detail until the weekend,  I'm kind of doing this by proxy at the moment:.

The reason I assumed it was a connector serialization issue was when using the connector with OrmLite, the output query differs.  But the end result is that using `ExecuteNonQuery(this IDbConnection dbConn, string sql)` in OrmLite behaves differently between the clients.

PR #413, opened 2017-12-21
fix columnDefinitionPayload, skip all 0 bytes in tail 

Seamas:
adapt Alibaba Analytic DB


bgrainger:
I assume this is related to https://www.alibabacloud.com/product/analytic-db ?

Is there any documentation related to these extra bytes, or do you have a packet capture showing them?

bgrainger:
I found a bug in the code you opened a PR against, so I fixed it: f2bacdaae6c3aa0f3ac610401a80c643bae3dffe

At the same time, I relaxed this restriction (about the packet length), which should avoid any exception if AnalyticDB is sending extra bytes.

bgrainger:
Fixed in [0.34.0](https://github.com/mysql-net/MySqlConnector/releases/tag/0.34.0).

PR #416, opened 2017-12-26
Moved m_cachedProcedures to static dictionary

pritchin:
To reduce querying to `information_schema.parameters`

### DIFF
--- a/src/MySqlConnector/MySql.Data.MySqlClient/MySqlConnection.cs
+++ b/src/MySqlConnector/MySql.Data.MySqlClient/MySqlConnection.cs
@@ -1,5 +1,5 @@
 using System;
-using System.Collections.Generic;
+using System.Collections.Concurrent;
 using System.Data;
 using System.Data.Common;
 using System.Net.Sockets;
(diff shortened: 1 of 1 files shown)


src/MySqlConnector/MySql.Data.MySqlClient/MySqlConnection.cs
```diff
@@ -302,16 +302,10 @@ internal void Cancel(MySqlCommand command)
...
-
 			var normalized = NormalizedSchema.MustNormalize(name, Database);
-			if (!m_cachedProcedures.TryGetValue(normalized.FullyQualified, out var cachedProcedure))
-			{
-				cachedProcedure = await CachedProcedure.FillAsync(ioBehavior, this, normalized.Schema, normalized.Component, cancellationToken).ConfigureAwait(false);
-				m_cachedProcedures[normalized.FullyQualified] = cachedProcedure;
-			}
-			return cachedProcedure;
+
+			return await m_cachedProcedures.GetOrAdd(normalized.FullyQualified,
```
bgrainger: I suspect this is unsafe because `normalized.FullyQualified` isn't globally unique across all servers and databases that an arbitrary .NET process might connect to.

src/MySqlConnector/MySql.Data.MySqlClient/MySqlConnection.cs
```diff
@@ -302,16 +302,10 @@ internal void Cancel(MySqlCommand command)
...
+				(key) => CachedProcedure.FillAsync(ioBehavior, this, normalized.Schema, normalized.Component, cancellationToken));
```
bgrainger: Capturing `this` will leak the connection and prevent the regular clean-up logic from detecting whether the user has abandoned the `MySqlConnection` object without closing it.

bgrainger: If the `cancellationToken` is cancelled, a Canceled Task will be cached, preventing any future client from succeeding.

bgrainger: Missing `.ConfigureAwait(false)`; this can lead to deadlock.

bgrainger:
Solved this in an alternate way: bc145ab3cae57a96e5a9f4ee2191cff21cb89d12

PR #437, opened 2018-02-15
Check for Private Key in CertificateFile

### LINKED ISSUE #436: CertificateFile without Bundled Key does not use Mutual Authentication
Both a Client Certificate and a Client Key are needed to perform [Mutual Authentication](https://en.wikipedia.org/wiki/Mutual_authentication)

`X509Certificate2` allows loading a PEM encoded certificate from CertificateFile, but the [X509Certificate2.HasPrivateKey](https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509certificate2.hasprivatekey(v=vs.110).aspx) property will be false.  The SSL connection will still succeed, but will not be mutually authenticated - [SslStream.IsMutuallyAuthenticated](https://msdn.microsoft.com/en-us/library/system.net.security.sslstream.ismutuallyauthenticated(v=vs.110).aspx) will be `false`.  To the user, this is unexpected behavior; when they provide a Client Certificate and the connection succeeds, they would assume the connection is using Mutual Authentication.

Loading separate Certificate and Key files has spotty support - it is possible in `netstandard2.0` but not elegant: https://github.com/dotnet/corefx/issues/17892.  If we wanted to support it, it'd probably be better to find a third-party library (BouncyCastle?) that handles loading and validating certificates in different formats.

I think that the best path forward that does not involve adding dependencies would be to only allow pkcs12 (.pfx) certificates with a bundled Certificate and Private Key in the `CertificateFile`.

caleblloyd:
- Fixes #436


src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -772,6 +772,13 @@ private void VerifyConnected()
 				try
 				{
 					var certificate = new X509Certificate2(cs.CertificateFile, cs.CertificatePassword);
+					if (!certificate.HasPrivateKey)
+					{
+						m_logArguments[1] = cs.CertificateFile;
+						Log.Error("{0} no private key included with certificate '{1}'", m_logArguments);
+						throw new MySqlException("CertificateFile does not contain a private key. "+
+						                         "CertificateFile should in PKCS #12 (.pfx) format and contain both a Certificate and Private Key");
```
bgrainger: Nitpick: `should be in`

PR #458, opened 2018-03-16
Use the best TLS version by default

bgrainger:
Specifying `SslProtocols.None` allows the .NET Framework to choose the best TLS version supported by the OS (on .NET 4.7 and later).

However, negotiating TLS 1.2 with a Windows Schannel client against a yaSSL-based MySQL Server will fail (see #101); automatically try with a lower TLS version if it's possible we're in this scenario.

This replaces #450, which moved specifying the TLS version into a connection string setting. In this PR, the library tries to do the best thing possible, at the cost (for Windows clients only) of extra TLS connection time for MySQL < 8.0 and/or .NET Framework < 4.7.

### DIFF
--- a/src/MySqlConnector/Core/ServerSession.cs
+++ b/src/MySqlConnector/Core/ServerSession.cs
@@ -225,60 +225,93 @@ namespace MySqlConnector.Core
 					VerifyState(State.Created);
 					m_state = State.Connecting;
 				}
-				var connected = false;
-				if (cs.ConnectionType == ConnectionType.Tcp)
-					connected = await OpenTcpSocketAsync(cs, loadBalancer, ioBehavior, cancellationToken).ConfigureAwait(false);
-				else if (cs.ConnectionType == ConnectionType.Unix)
-					connected = await OpenUnixSocketAsync(cs, ioBehavior, cancellationToken).ConfigureAwait(false);
-				if (!connected)
+
+				// TLS negotiation should automatically fall back to the best version supported by client and server. However,
+				// Windows Schannel clients will fail to connect to a yaSSL-based MySQL Server if TLS 1.2 is requested and
+				// have to use only TLS 1.1: https://github.com/mysql-net/MySqlConnector/pull/101
+				// In order to use the best protocol possible (i.e., not always default to TLS 1.1), we try the OS-default protocol
+				// (which is SslProtocols.None; see https://docs.microsoft.com/en-us/dotnet/framework/network-programming/tls),
+				// then fall back to SslProtocols.Tls11 if that fails and it's possible that the cause is a yaSSL server.
+				bool shouldRetrySsl;
+				var sslProtocols = Pool?.SslProtocols ?? Utility.GetDefaultSslProtocols();
+				PayloadData payload;
+				InitialHandshakePayload initialHandshake;
+				do
 				{
-					lock (m_lock)
-						m_state = State.Failed;
-					Log.Error("{0} connecting failed", m_logArguments);
-					throw new MySqlException("Unable to connect to any of the specified MySQL hosts.");
-				}
+					shouldRetrySsl = (sslProtocols == SslProtocols.None || (sslProtocols & SslProtocols.Tls12) == SslProtocols.Tls12) && Utility.IsWindows();
 
-				var byteHandler = new SocketByteHandler(m_socket);
-				m_payloadHandler = new StandardPayloadHandler(byteHandler);
-
-				var payload = await ReceiveAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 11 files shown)


src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -225,60 +225,109 @@ public void FinishQuerying()
...
+					{
+						if (!serverSupportsSsl)
+						{
+							Log.Error("{0} requires SSL but server doesn't support it", m_logArguments);
+							throw new MySqlException("Server does not support SSL");
+						}
+
+						// MySQL Server 8.0 and MariaDB are compiled with OpenSSL, which negotiates TLS 1.2 correctly; a failure should
+						// not cause an automatic retry with an older/weaker TLS version.
+						var serverHasOpenSsl = shouldRetrySsl && (ServerVersion.Version.Major >= 8 || ServerVersion.OriginalString.IndexOf("MariaDB", StringComparison.OrdinalIgnoreCase) != -1);
```
vaintroub: MariaDB is not compiled with openssl on Windows, not yet at least. MariaDB (currently only in 10.2.6+ , and in the future in other versions) fixes YASSL's  TLS version negotiation, so that it downgrades to TLS1.2 that schannel sends, to TLS 1.1 without fatal error.

vaintroub: MySQL is compiled with OpenSSL almost everywhere, too. But not on Windows prior to 8.0, not in tar.gz packages, and maybe in 1-2 other obscure cases.

bgrainger: Thanks for the correction; you already described the MariaDB/yaSSL situation in #450 but I didn't follow it carefully enough. I need to fix the name/description of this variable at a minimum, or maybe rework this section.

RE: MySQL

> MySQL Community Edition binary distributions are compiled using OpenSSL. (Prior to MySQL 8.0.4, MySQL Community Edition binary distributions are compiled using yaSSL.)

https://dev.mysql.com/doc/refman/8.0/en/openssl-versus-yassl.html

I don't know if it's "obscure" but the DockerHub `mysql:5.7` image appears to be built with yaSSL.

src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -225,60 +225,109 @@ public void FinishQuerying()
...
+							// Prior to .NET Framework 4.7, SslProtocols.None is not a valid argument to AuthenticateAsClientAsync.
+							// If the NET46 build is loaded by an application that targets. NET 4.7, SslProtocols.None will work; however,
+							// if the application targets .NET 4.6.2 or earlier, then it will fail at runtime; we catch the exception and retry.
+							Log.Warn(ex, "{0} doesn't support SslProtocols.None; retrying with SslProtocols.Tls12", m_logArguments);
+							sslProtocols = SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12;
+							if (Pool != null)
+								Pool.SslProtocols = sslProtocols;
+						}
+#endif
+						catch (Exception ex) when (shouldRetrySsl && serverHasOpenSsl && ((ex is MySqlException && ex.InnerException is IOException) || ex is IOException))
```
bgrainger: This logic is reversed; it was supposed to be `!serverHasOpenSsl`.

This code fails when tested against a local instance of MySQL Server 5.7.21 for Windows (but apparently not on Appveyor?).

src/MySqlConnector/Core/ServerSession.cs
vaintroub: Let me clarify the OpenSSL statement.  Oracle MySQL, or MariaDB, both provide tar.gz or zip/MSI for downloads, yet outside Windows, those packages are not used much.

Instead, Linux/BSD distribution packagers would build MySQL or MariaDB the way they think is correct, and this normally means openssl, since  packagers prefer better known libraries over less known, and shared libraries over compiled-in static ones.

As for docker, you can make a good guess how it was compiled by using "Show session status like 'Ssl_cipher_list' " as described in https://dev.mysql.com/doc/refman/5.7/en/encrypted-connection-protocols-ciphers.html . Yassl list is short, openssl is much larger.

bgrainger:
I'm reverting the `ServerVersion`-based check for now. It was intended to avoid a TLS downgrade (and extra server roundtrip) for servers that the client believes should negotiate TLS 1.2 correctly without an exception. However, it's not clear that the check covers enough situations to be useful. Additionally, it seems like an attacker who can modify traffic to force a TLS downgrade could modify the server's initial handshake to report a lower `ServerVersion` and bypass this check.

bgrainger:
Added in [0.37.0](https://github.com/mysql-net/MySqlConnector/releases/tag/0.37.0).

Issue #459, opened 2018-03-19
MySqlDbType.String not respected when assigning an enum to MySqlParameter Value.


Naragato:
[https://github.com/Naragato/OrmLiteMySQLClientTests](url)

Using the official oracle client with and setting a MySQL param as MySqlDbType.String, and using an enum value results in the string value of the enum inserted into the database. However using mysql-net, int is inserted.

bgrainger:
Fixed in [0.37.1](https://github.com/mysql-net/MySqlConnector/releases/tag/0.37.1).

PR #460, opened 2018-03-19
Add string check for enums

Naragato:
#459


Naragato:
Actually slight error in my PR will correct now :)

bgrainger:
Please add a test that verifies the new behaviour.

Naragato:
Done. : >

bgrainger:
Sorry, I wasn't clear: the test needs to be in the SideBySide project to verify that the behaviour is the same in MySqlConnector and Connector/NET (since this is trying to clone existing Connector behaviour).

Please also revert the changes to the key files.

Naragato:
Reverted the certs (not sure what actually changed them :/ I left the original test in place as it makes sense to be there. However I added an additional test into the sidebyside project.

src/MySqlConnector/MySql.Data.MySqlClient/MySqlParameter.cs
```diff
@@ -303,6 +303,10 @@ internal void AppendSqlString(BinaryWriter writer, StatementPreparerOptions opti
 			{
 				writer.WriteUtf8("{0}".FormatInvariant((ulong) Value));
 			}
+			else if (MySqlDbType == MySqlDbType.String || MySqlDbType == MySqlDbType.VarChar && HasSetDbType && Value is Enum)
```
bgrainger: Precedence requires parentheses around `(MySqlDbType == MySqlDbType.String || MySqlDbType == MySqlDbType.VarChar)`.

bgrainger:
Thanks for the bug report and PR!

Issue #315, opened 2017-08-25
Add logging with LibLog


bgrainger:
[LibLog](https://github.com/damianh/LibLog) ([NuGet](https://www.nuget.org/packages/LibLog/)) is "a single file for you to either copy/paste or install via nuget, into your library/framework/application to provide a logging abstraction".

> A source code package designed primarily for library and framework authors who want dependency free logging support in their component.

bgrainger:
After getting the LibLog source code, I see that it uses `dynamic` in a few places, which is potentially bad for performance and brings in unwanted dependencies.

Additionally, it's not clear that the library was developed with performance as a primary concern; ideally, the library should be (close to) zero overhead if logging is disabled. damianh/LibLog#129 discusses this and other issues.

Npgsql had a custom logging implementation, then switched to `Microsoft.Extensions.Logging`, then reverted that: npgsql/npgsql#1438. They are currently looking at LibLog, and may switch when performance is improved.

For now, I plan to follow their approach and define a custom interface. We could also publish additional NuGet packages for common logging frameworks that automatically hook them up.

PR #464, opened 2018-03-23
Add Serilog logging adapter. Fixes #463.

### LINKED ISSUE #463: Serilog logging integration
Need an IMySqlConnectorLoggerProvider for Serilog. Willing to build/contribute if there isn't one hanging about out there.

marcrocny:
Adds "light" structuring to the messages, transforming them into Serilog templates by concatenating the logger name and argument index, and lets Serilog use the arguments as Properties.

**This is not ideal.** This method loses the identity of a Property between classes, so it undercuts filtering or relating events between classes by the common Property. E.g., Session ID will have the same value but different names across `TextCommandExecutor`, `ServerSession` and `ConnectionPool`.

This could be overcome by hand-creating a map from name-index to Property-name. That seems pretty fragile, though.

Is the trade-off worth it? Are the argument lists relatively static? Could there be a way to "signal" changes so the Serilog adapter could be automatically update? Any better ideas?

Also, there's no automated tests for this. However, that's the same as the other adapters and it is very straightforward.

### DIFF
--- a/src/MySqlConnector/Core/ConnectionPool.cs
+++ b/src/MySqlConnector/Core/ConnectionPool.cs
@@ -29,7 +29,7 @@ namespace MySqlConnector.Core
 			// on the lock in RecoverLeakedSessions in high-concurrency situations
 			if (m_sessionSemaphore.CurrentCount == 0 && unchecked(((uint) Environment.TickCount) - m_lastRecoveryTime) >= 1000u)
 			{
-				Log.Warn("{0} is empty; recovering leaked sessions", m_logArguments);
+				Log.Warn("Pool{0} is empty; recovering leaked sessions", m_logArguments);
 				RecoverLeakedSessions();
 			}
 
@@ -37,7 +37,7 @@ namespace MySqlConnector.Core
 				await CreateMinimumPooledSessions(ioBehavior, cancellationToken).ConfigureAwait(false);
 
 			// wait for an open slot (until the cancellationToken is cancelled, which is typically due to timeout)
-			Log.Debug("{0} waiting for an available session", m_logArguments);
+			Log.Debug("Pool{0} waiting for an available session", m_logArguments);
 			if (ioBehavior == IOBehavior.Asynchronous)
 				await m_sessionSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
 			else
@@ -57,12 +57,12 @@ namespace MySqlConnector.Core
 				}
 				if (session != null)
 				{
-					Log.Debug("{0} found an existing session; checking it for validity", m_logArguments);
+					Log.Debug("Pool{0} found an existing session; checking it for validity", m_logArguments);
 					bool reuseSession;
 
 					if (session.PoolGeneration != m_generation)
 					{
-						Log.Debug("{0} discarding session due to wrong generation", m_logArguments);
+						Log.Debug("Pool{0} discarding session due to wrong generation", m_logArguments);
 						reuseSession = false;
 					}
 					else
@@ -84,7 +84,7 @@ namespace MySqlConnector.Core
 					if (!reuseSession)
 					{
 						// session is either old or cannot communicate with the server
-						Log.Warn("{0} Session{1} is unusable; destroying it", m_logArguments[0], session.Id);
+						Log.Warn("Pool{0} Session{1} is unusable; destroying it", m_logArguments[0], session.Id);
 						AdjustHostConnectionCount(session, -1);
 						await session.DisposeAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
 					}
@@ -99,7 +99,7 @@ namespace MySqlConnector.Core
 							leasedSessionsCountPooled = m_leasedSessions.Count;
 						}
 						if (Log.IsDebugEnabled())
-							Log.Debug("{0} returning pooled Session{1} to caller; m_leasedSessions.Count={2}", m_logArguments[0], session.Id, leasedSessionsCountPooled);
+							Log.Debug("Pool{0} returning pooled Session{1} to caller; LeasedSessionsCount={2}", m_logArguments[0], session.Id, leasedSessionsCountPooled);
 						return session;
 					}
 				}
@@ -107,7 +107,7 @@ namespace MySqlConnector.Core
 				// create a new session
 				session = new ServerSession(this, m_generation, Interlocked.Increment(ref m_lastSessionId));
 				if (Log.IsInfoEnabled())
-					Log.Info("{0} no pooled session available; created new Session{1}", m_logArguments[0], session.Id);
+					Log.Info("Pool{0} no pooled session available; created new Session{1}", m_logArguments[0], session.Id);
 				await session.ConnectAsync(ConnectionSettings, m_loadBalancer, ioBehavior, cancellationToken).ConfigureAwait(false);
 				AdjustHostConnectionCount(session, 1);
 				session.OwningConnection = new WeakReference<MySqlConnection>(connection);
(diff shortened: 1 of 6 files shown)


marcrocny:
Hang on, this needs some cleanup.  Plus, I have an idea about property names.

marcrocny:
Okay, I couldn't simple or robust naming to work because of the variety in the existing log messages.

@bgrainger, some structured-logging options to run by you:

1. Reformatting the existing log messages (~100 calls) (I'd be gentle!). Keep them working OOTB but make them "machine readable" for structured logging.
3. Pursue the aforementioned hand-rolled map (which upon further investigation may not be workable).
2. Ditch current logging abstraction entirely and use [LibLog](https://github.com/damianh/LibLog).
3. Just call the approach in this PR "good enough" as is.

bgrainger:
1. Can you give a couple of examples of what "Reformatting the existing log messages" would look like?
2. I'm not sure I fully understand this one, but agree that it sounds brittle.
3. I previously considered LibLog but decided against it here: https://github.com/mysql-net/MySqlConnector/issues/315#issuecomment-341946525
4. I'm OK with this, but I'm not a Serilog user; is the current PR truly "good enough" for your needs, or does it really need improvement to work for people who have chosen to use Serilog?

marcrocny:
1. Below.
2. A hard-coded map of "arg X on logger Y -> property Z". Brittle and also changes from message to message. 
3. No LibLog. That's fair.
4. The behavior would be surprising, possibly frustrating to a Serilog user.

Reformatting: Make property references in messages conform to `PropName{n}` or `PropName {n}` or `PropName '{n}'`. `PropName` would be strict across loggers/messages.  This would allow a transform of `{n}` into `{PropName}` to transform the message into a Serilog [template](https://github.com/serilog/serilog/wiki/Writing-Log-Events).

For example, in `ConnectionPool`, `m_logArguments[0]` would be changed from `"Pool{0}".FormatInvariant(Id)` to just `"{0}".FormatInvariant(Id)`. All log messages would refer to `"Pool{0}..."`. In this class most references to `"Session{n}"` would already be correctly formatted.

bgrainger:
So, for example, https://github.com/mysql-net/MySqlConnector/blob/8f29217d32534f311cb8c2a12edb81ceca67bbe0/src/MySqlConnector/Core/ConnectionPool.cs#L87 would become:

    Log.Warn("Pool{0} Session{1} is unusable; destroying it", m_logArguments[0], session.Id);

and https://github.com/mysql-net/MySqlConnector/blob/8f29217d32534f311cb8c2a12edb81ceca67bbe0/src/MySqlConnector/Core/ServerSession.cs#L64 would become:

     Log.Debug("Session{0} returning to Pool{1}", m_logArguments);

and it's OK that across different classes you might have `Pool{0}`, `Pool{1}`, etc.? This change seems completely reasonable; there's no particularly good reason that `Pool` was concatenated as part of `m_logArguments[0]` other than it saved a bit of typing in each log message.

How would you handle arbitrary logging parameters, such as:

```
m_logArguments[1] = authPluginName;
Log.Debug("{0} server sent auth_plugin_name '{1}'", m_logArguments);
...
Log.Info("{0} made connection; ServerVersion={1}; ConnectionId={2}; Flags: {3}{4}{5}{6}", m_logArguments[0], ServerVersion.OriginalString, ConnectionId,
	m_useCompression ? "Cmp " : "", m_supportsConnectionAttributes ? "Attr " : "", m_supportsDeprecateEof ? "" : "Eof ", serverSupportsSsl ? "Ssl " : "");
```

(from https://github.com/mysql-net/MySqlConnector/blob/8f29217d32534f311cb8c2a12edb81ceca67bbe0/src/MySqlConnector/Core/ServerSession.cs#L269-L286)

marcrocny:
For
```c#
m_logArguments[1] = authPluginName;
Log.Debug("{0} server sent auth_plugin_name '{1}'", m_logArguments);
...
Log.Info("{0} made connection; ServerVersion={1}; ConnectionId={2}; Flags: {3}{4}{5}{6}",
    m_logArguments[0], ServerVersion.OriginalString, ConnectionId,
    m_useCompression ? "Cmp " : "", m_supportsConnectionAttributes ? "Attr " : "", 
    m_supportsDeprecateEof ? "" : "Eof ", serverSupportsSsl ? "Ssl " : "");
```

First,
https://github.com/mysql-net/MySqlConnector/blob/8f29217d32534f311cb8c2a12edb81ceca67bbe0/src/MySqlConnector/Core/ServerSession.cs#L41

would become `m_logArguments = new object[] { "{0}".FormatInvariant(Id), null };` 
and the above would be changed to

```c#
m_logArguments[1] = authPluginName;
Log.Debug("Session{0} server sent auth_plugin_name '{1}'", m_logArguments);
...
Log.Info("Session{0} made connection; ServerVersion={1}; ConnectionId={2}; Compression={3}; Attributes={4}; DeprecateEof={5}; Ssl={6}", 
    m_logArguments[0], ServerVersion.OriginalString, ConnectionId, 
    m_useCompression, m_supportsConnectionAttributes, m_supportsDeprecateEof, serverSupportsSsl);
```
The templates for these would be `"Session{Session} server sent auth_plugin_name '{auth_plugin_name}'"` and `Session{Session} made connection; ServerVersion={ServerVersion}; ConnectionId={ConnectionId}; Compression={Compression}; Attributes={Attributes}; DeprecateEof={DeprecateEof}; Ssl={Ssl}"`.

The latter makes it less terse, but also communicates more--the message directly shows what isn't set. I'm assuming any changes to message formats aren't considered "breaking" as long as info isn't lost.

bgrainger:
:+1: sounds good.

I assume these templates reside in the MySqlConnector.Logging.Serilog project and get substituted in at runtime?

Thanks for the help!

src/MySqlConnector/Core/ConnectionPool.cs
```diff
@@ -438,9 +438,9 @@ private ConnectionPool(ConnectionSettings cs)
 				(ILoadBalancer) new RoundRobinLoadBalancer();
 
 			Id = Interlocked.Increment(ref s_poolId);
-			m_logArguments = new object[] { "Pool{0}".FormatInvariant(Id) };
+			m_logArguments = new object[] { "{0}".FormatInvariant(Id) };
 			if (Log.IsInfoEnabled())
-				Log.Info("{0} creating new connection pool for {1}", m_logArguments[0], cs.ConnectionStringBuilder.GetConnectionString(includePassword: false));
+				Log.Info("Pool{0} creating new connection pool for ConnectionString{1}", m_logArguments[0], cs.ConnectionStringBuilder.GetConnectionString(includePassword: false));
```
bgrainger: Would `ConnectionString=` or `ConnectionString:` work here?

src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -159,7 +159,7 @@ public void FinishQuerying()
 				// KILL QUERY will kill a subsequent query if the command it was intended to cancel has already completed.
 				// In order to handle this case, we issue a dummy query that will consume the pending cancellation.
 				// See https://bugs.mysql.com/bug.php?id=45679
-				Log.Info("{0} sending 'DO SLEEP(0)' command to clear pending cancellation", m_logArguments);
+				Log.Info("Session{0} sending 'DO SLEEP(0)' command to clear pending cancellation", m_logArguments[0]);
```
bgrainger: This should remain as `m_logArguments` to avoid allocating a `params object[]` array when calling the logging method.

src/MySqlConnector/Core/ConnectionPool.cs
marcrocny: Yes; more below.

marcrocny:
The transform and message changes in this PR allow for using the word preceding a `string.Format()`-style token as a Property name for structured logging. I built some flexibility in based on existing message styles using a Regex:
https://github.com/mysql-net/MySqlConnector/blob/ccfa5d102eea263c11867f049bea4a013920d00a/src/MySqlConnector.Logging.Serilog/SerilogLoggerProvider.cs#L60

Some examples of property name styles it will recognize:

* `PropName{0}` 
* `PropName {0}` 
* `PropName: {0}` 
* `PropName={0}` 
* `PropName = {0}` 
* `PropName '{0}'` 
* `PropName : '{0}'` 

Because Serilog works best with the property names declared within message template strings, this log adapter works by directly transforming the message strings. It also generates them with a prefix of "MySql" so they can be differentiated from other property sources.

PR #470, opened 2018-04-02
NLogLoggerProvider - NLog 4.5.0 with NetStandard Support

snakefoot:
NLog 4.5.0 supports structured logging like Serilog, but not sure if it is a good idea to repeat its search-replace logic for nicer property-names.

Right now it works just like the log4net provider.


bgrainger:
Thanks for the contribution!

bgrainger:
Published [at NuGet](https://www.nuget.org/packages/MySqlConnector.Logging.NLog/).

PR #484, opened 2018-04-18
Use DateTimeKind for all returned dates and date/times.

ejball:


bgrainger:
Cherry-picked these changes in 4460be865b5f9ed05ac4eb62418702a3cd60962b (and added a fix for MariaDB).

bgrainger:
From [MariaDB docs](https://mariadb.com/kb/en/library/microseconds-in-mariadb/):

> Note: When you convert a temporal value to a value with a smaller precision, it will be truncated, not rounded. This is done to guarantee that the date part is not changed. For example:
> `SELECT CAST('2009-12-31 23:59:59.998877' as DATETIME(3));`
> `-> 2009-12-31 23:59:59.998`

bgrainger:
Fixed in [0.40.1](https://github.com/mysql-net/MySqlConnector/releases/tag/0.40.1).

PR #505, opened 2018-05-30
Support GuidFormat connection option.

ejball:


tests/SideBySide/DataTypes.cs
```diff
@@ -495,6 +495,121 @@ public void QueryBinaryGuid(bool oldGuids)
...
+			bool isLittleEndianBinary16 = guidFormat == MySqlGuidFormat.LittleEndianBinary16 || (guidFormat == MySqlGuidFormat.Default && oldGuids);
+
+			Guid guid = new Guid("00112233-4455-6677-8899-AABBCCDDEEFF");
+			string guidChar36 = "00112233-4455-6677-8899-AABBCCDDEEFF";
+			string guidChar32 = "00112233445566778899AABBCCDDEEFF";
+			byte[] guidBinary16 = { 0x00, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77, 0x88, 0x99, 0xAA, 0xBB, 0xCC, 0xDD, 0xEE, 0xFF };
+			byte[] guidTimeSwapBinary16 = { 0x66, 0x77, 0x44, 0x55, 0x00, 0x11, 0x22, 0x33, 0x88, 0x99, 0xAA, 0xBB, 0xCC, 0xDD, 0xEE, 0xFF };
+			byte[] guidLittleEndianBinary16 = { 0x33, 0x22, 0x11, 0x00, 0x55, 0x44, 0x77, 0x66, 0x88, 0x99, 0xAA, 0xBB, 0xCC, 0xDD, 0xEE, 0xFF };
+
+			const bool mysql8 = false;
```
ejball: Is there any way to detect MySQL 8+?

bgrainger: I can think of several:

* Extend `ServerFeatures` enum with `UuidToBin` member, turn that off on other DBs
* Check `ServerVersion` at runtime for >= 8 (not ideal because MariaDB 10.3 may or may not support this function)
* Execute `SELECT UUID_TO_BIN` during test setup and set the flag based on whether an exception occurs

Overall I'd lean towards `ServerFeatures` as that's the standard way the test suite indicates whether a server feature can be expected to be present.

tests/SideBySide/DataTypes.cs
```diff
@@ -495,6 +495,121 @@ public void QueryBinaryGuid(bool oldGuids)
...
+			bool isBinary16 = guidFormat == MySqlGuidFormat.Binary16;
+			bool isTimeSwapBinary16 = guidFormat == MySqlGuidFormat.TimeSwapBinary16;
+			bool isLittleEndianBinary16 = guidFormat == MySqlGuidFormat.LittleEndianBinary16 || (guidFormat == MySqlGuidFormat.Default && oldGuids);
+
+			Guid guid = new Guid("00112233-4455-6677-8899-AABBCCDDEEFF");
+			string guidChar36 = "00112233-4455-6677-8899-AABBCCDDEEFF";
+			string guidChar32 = "00112233445566778899AABBCCDDEEFF";
+			byte[] guidBinary16 = { 0x00, 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77, 0x88, 0x99, 0xAA, 0xBB, 0xCC, 0xDD, 0xEE, 0xFF };
+			byte[] guidTimeSwapBinary16 = { 0x66, 0x77, 0x44, 0x55, 0x00, 0x11, 0x22, 0x33, 0x88, 0x99, 0xAA, 0xBB, 0xCC, 0xDD, 0xEE, 0xFF };
+			byte[] guidLittleEndianBinary16 = { 0x33, 0x22, 0x11, 0x00, 0x55, 0x44, 0x77, 0x66, 0x88, 0x99, 0xAA, 0xBB, 0xCC, 0xDD, 0xEE, 0xFF };
```
bgrainger: It's a bit confusing that these variables are all prefixed with `guid` yet none of them is a `Guid`.

src/MySqlConnector/MySql.Data.MySqlClient/MySqlParameter.cs
```diff
@@ -262,17 +262,42 @@ internal void AppendSqlString(BinaryWriter writer, StatementPreparerOptions opti
...
 							writer.Write((byte) 0x5C);
 						writer.Write(by);
 					}
 					writer.Write((byte) '\'');
 				}
+				else if (guidOptions == StatementPreparerOptions.GuidFormatChar32)
+				{
+					writer.WriteUtf8("'{0:N}'".FormatInvariant(guidValue));
+				}
 				else
```
bgrainger: Should this throw an exception if you try to save a `Guid` when you've explicitly opted in to `GuidFormat.None`?

tests/SideBySide/DataTypes.cs
ejball: That's fair. `guidAsChar32` perhaps.

src/MySqlConnector/MySql.Data.MySqlClient/MySqlParameter.cs
ejball: Yeah, I wondered that, too. I decided that rendering a `Guid` in a `UUID()`-compatible format seemed reasonable even if we're not doing any automatic column conversion on the way out.

tests/SideBySide/DataTypes.cs
ejball: Okay, that doesn't look too hard.

src/MySqlConnector/MySql.Data.MySqlClient/MySqlParameter.cs
bgrainger: Implicit conversions are rare in this library, but that does sound reasonable.

ejball: Oh, that reminds me, `Row.GetGuid` should respect the binary format if one is set... I'll make that change too.

PR #516, opened 2018-06-13
Update performance tests to netcoreapp2.1

caleblloyd:
I will update the [Concurrency Benchmark Results](https://github.com/mysql-net/MySqlConnector/wiki/Concurrency-Benchmark-Results) with the newest versions of `MySqlConnector` and `MySql.Data` on .NET Core 2.1 after merging


caleblloyd:
Added a few more changes to bring this "in-line" with the way we copy configurations to the output directory, and the way that ASP.NET MVC is recommending to build WebHost objects these days

PR #517, opened 2018-06-13
Change MySqlConnector.Performance SDK to Microsoft.NET.Sdk.Web

caleblloyd:
Here is another fun issue in .NET Core 2.1: https://github.com/aspnet/Mvc/issues/7882#issuecomment-395880248

Apparently since this now references `Microsoft.AspNetCore.All`, we are supposed to use `<Project Sdk="Microsoft.NET.Sdk.Web">`.  However, the following MSBuild Task does not work in `<Project Sdk="Microsoft.NET.Sdk.Web">`:

```
  <ItemGroup>
    <None Update="appsettings.json">
      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
    </None>
    <None Update="config.json">
      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
    </None>
  </ItemGroup>
```

This MSBuild Task _does_ work in `<Project Sdk="Microsoft.NET.Sdk">`, which is why I initially changed the project SDK.  Our `MySqlConnector.Performance` project seems to work fine with either Project SDK, but a similar project in Pomelo had it's routes stopped working with `<Project Sdk="Microsoft.NET.Sdk">` and required the change to `<Project Sdk="Microsoft.NET.Sdk.Web">`

In the interest of following Microsoft's migration notes, I've changed `MySqlConnector.Performance` to use `<Project Sdk="Microsoft.NET.Sdk.Web">`, and I've added a different msbuild task that accomplishes the same output as the `<CopyToOutputDirectory>` task.

### DIFF
--- a/tests/MySqlConnector.Performance/MySqlConnector.Performance.csproj
+++ b/tests/MySqlConnector.Performance/MySqlConnector.Performance.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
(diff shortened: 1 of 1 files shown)


caleblloyd:
As an added bonus, this msbuild task will also recursively copy directories, which can be useful for web projects  that need to copy a static asset directory.  Not needed in this project, but good boilerplate code for other projects...

bgrainger:
Also, the mixed capitalisation in `Microsoft.NET.Sdk` drives me insane every time I see it, but I’m learning to let go…

(And of course it’s [case-sensitive](https://github.com/Microsoft/msbuild/issues/2510).)

caleblloyd:
I am also new to the concept, I didn't even notice there was a difference before.

I did remember reading about this change in .NET Core 2.0 when `Microsoft.AspNetCore.All` came along though:
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/metapackage?view=aspnetcore-2.1

> When you use the Microsoft.AspNetCore.All metapackage, no assets from the referenced ASP.NET Core NuGet packages are deployed with the application — the .NET Core Runtime Store contains these assets.

I just now learned from the https://github.com/aspnet/Mvc/issues/7882#issuecomment-395880248 that using `Microsoft.AspNetCore.All` requires using the `Microsoft.NET.Sdk.Web` SDK though.  Maybe this is how it knows to pull the packages from the runtime store instead of from NuGet?

And yes... also learned from copying and pasting the invalid case example from that comment that it is, in fact, case sensitive :smile:

Issue #375, opened 2017-10-22
Support more schema collections in GetSchema


bgrainger:
Connector/NET supports the following schema collections in `MySqlConnection.GetSchema`. MySqlConnector only supports the bolded entries.

* **MetaDataCollections**
* DataSourceInformation
* **DataTypes**
* Restrictions
* ReservedWords
* Databases
* Tables
* Columns
* Users
* Foreign Keys
* IndexColumns
* Indexes
* Foreign Key Columns
* UDF
* Views
* ViewColumns
* Procedure Parameters
* **Procedures**
* Triggers

Add support for additional schema collections that weren't implemented in #361.

phillipc:
Look at my PR: I have implemented Foreign Keys-, IndexColumns-, Indexes- and Foreign Key Columns-Support

eaexo:
Hi @bgrainger

Are there any plans to fix this issue? Missing schema collections make the mysqlconnector.net kind of useless with NHibernate (even if one could discuss if it make sense to use automatic scheme update)

bgrainger:
`Foreign Keys`, `Indexes`, and `Index Columns` support were added by @eaexo in #1419.

eaexo:
Hello @bgrainger 
when can we expect to have at least a beta released resp. published to the nuget.org?

bgrainger:
Foreign Keys, Indexes, Index Columns added in [2.3.2](https://github.com/mysql-net/MySqlConnector/releases/tag/2.3.2).

eaexo:
Hi @bgrainger 

unfortunatelly in https://github.com/mysql-net/MySqlConnector/commit/8748a725e6de168c01094f10b6206a8d6a6fae70#diff-b53ea1b55d852cbcb0fe21dab331dcf093424734bd32164bc03f4df0d6624b3cL843 (probably) you broke it again.

At least NHibernate passes 5 restrictions for determining index columns, see  `NHibernate.Dialect.Schema.AbstractDataBaseSchema`.

```

		public virtual DataTable GetIndexColumns(string catalog, string schemaPattern, string tableName, string indexName)
		{
			if (UseDialectQualifyInsteadOfTableName)
			{
				var actualTableName = GetActualTableName(catalog, schemaPattern, tableName);
				return Connection.GetSchema("IndexColumns", new[] {null, null, actualTableName, indexName, null});
			}

			var restrictions = new[] {catalog, schemaPattern, tableName, indexName, null};
[...]

PR #558, opened 2018-09-12
Added support of ADO.NET "ReservedWords" collection schema (used for …

fedesasso:
Hi there. Hi added support of "ReservedWords" ADO.NET collection schema (see issue #375 "Support more schema collections in GetSchema").
Since I didn't find much in the official docs, I gathering the full list from what returned by the ODBC provider.
I needed that to make MySqlConnector work with NHibernate.


bgrainger:
The hard-coded list of reserved words for MySQL 5.7 is here: https://dev.mysql.com/doc/refman/5.7/en/keywords.html

This list includes entries, such as `ABSOLUTE` and `ADA` that aren't on that list.

For MySQL 8.0, the [`INFORMATION_SCHEMA.KEYWORDS`](https://dev.mysql.com/doc/refman/8.0/en/keywords-table.html) table could be used to load the list at runtime.

bgrainger:
I dumped the results of `select word from information_schema.keywords where reserved = 1;` on MySQL Server 8.0.12 in https://gist.github.com/bgrainger/54e7f3200772734855b08970792f0175.

This should be the most up-to-date list of reserved words that can be hard-coded in the connector.

bgrainger:
There's a bug (in 8.0.12 and earlier) that could make it impractical to query `INFORMATION_SCHEMA.KEYWORDS` at runtime: https://bugs.mysql.com/bug.php?id=90160

fedesasso:
Were it not for the annoying bug, I'd SELECT VERSION(); parse somehow the result string and if it were )> 8.0 I'd query the information_schema_keywords table, otherwise use a hard coded list.
Since the mysql bug there, the hard coded list seems the viable solution, regardless the db version (even if different versions would likely have different lists).

Do you want me to change the code to use the list you provided, and then make a new pull request, or do find it quicker to change the code on your own, since I guess you already are on it?
(sorry, I'm new to github workflows)

bgrainger:
> I'd SELECT VERSION(); parse somehow the result string and if it were )> 8.0

The problem with that is that MariaDB 10.0 is only compatible with MySQL 5.6 (but has a higher version number).

> Do you want me to change the code to use the list you provided, and then make a new pull request

Yes; just push a new commit to this PR (instead of opening a new PR).

PR #577, opened 2018-10-31
Support auth_gssapi authentication.

vaintroub:
This patch adds support for MariaDB's gssapi authentication plugin, passwordless authentication on Windows (in both domain and standalone machine environments), as well as on Linux, in Kerberos realm environment.

More information on authentication plugin https://mariadb.com/kb/en/library/authentication-plugin-gssapi

The patch uses .NET NegotiateStream to authenticate. It uses an auxiliary stream to convert NegotiateStream framing for SPNEGO/NTLM payloads into MySQL packet framing with the same payloads.

Things to note

* Currently,  the client "trusts" service principal name that server sends, and passes it as targetName in NegotiateStream.AuthenticateAsClientAsync.  Ideally , to support mutual authentication between client and server , service principal name should be an optional connection parameter.  If set, driver would ignore the string sent by the server. I did not add this into the patch,  but, if we agree it is a good idea, I can be easily added later.

*  On Linux, NegotiateStream implementation currently only works correctly with  service principle names (that's the dotnet implementation of it) . Most importantly , it  does not understand UPNs user principle names , like machine$@domain.com, which are typically in use by MariaDB servers running inside Windows AD.

Thus by default, connection from a Linux client to Windows server would fail (.NET runtime turns machine$@domain.com into funny name  machine$/domain.com, which KDC won't find).
 
There is however a  simple workaround on the server side, to allow Linux client to connect  Windows server, would be to set  server parameter --gssapi-principal-name=HOST/fqdn@DOMAIN.COM (host SPN, usually added whenever machine joins the domain).

### DIFF
--- a/src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
+++ b/src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
@@ -2,6 +2,7 @@ using System;
 using System.IO;
 using System.Net;
 using System.Net.Security;
+using System.Security.Authentication;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -216,13 +217,13 @@ namespace MySqlConnector.Protocol.Serialization
 			var reader = new ByteArrayReader(switchRequest.AsSpan());
 			return Encoding.UTF8.GetString(reader.ReadNullOrEofTerminatedByteString());
 		}
-		public static async Task<PayloadData> AuthenticateAsync(byte[] switchRequestPayloadData,
+		public static async Task<PayloadData> AuthenticateAsync(ConnectionSettings cs, byte[] switchRequestPayloadData,
 			ServerSession session, IOBehavior ioBehavior, CancellationToken cancellationToken)
 		{
 			using (var innerStream = new NegotiateToMySqlConverterStream(session, ioBehavior, cancellationToken))
 			using (var negotiateStream = new NegotiateStream(innerStream))
 			{
-				var targetName = GetServicePrincipalName(switchRequestPayloadData);
+				var targetName =cs.ServerSPN ?? GetServicePrincipalName(switchRequestPayloadData);
 #if NETSTANDARD1_3
 				await negotiateStream.AuthenticateAsClientAsync(CredentialCache.DefaultNetworkCredentials, targetName).ConfigureAwait(false);
 #else
@@ -235,6 +236,13 @@ namespace MySqlConnector.Protocol.Serialization
 					await negotiateStream.AuthenticateAsClientAsync(CredentialCache.DefaultNetworkCredentials, targetName).ConfigureAwait(false);
 				}
 #endif
+				if (cs.ServerSPN != null && !negotiateStream.IsMutuallyAuthenticated)
+				{
+					// Negotiate used NTLM fallback, server name cannot be verified.
+					throw new AuthenticationException(String.Format(
+						"GSSAPI : Unable to verify server principal name using authentication type {0}",
+						negotiateStream.RemoteIdentity?.AuthenticationType));
+				}
 				if (innerStream.MySQLProtocolPayload.ArraySegment.Array != null)
 					// return already pre-read OK packet.
 					return innerStream.MySQLProtocolPayload;
--- a/src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
+++ b/src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
@@ -79,7 +79,7 @@ namespace MySqlConnector.Protocol.Serialization
 					return NegotiateStreamConstants.HeaderLength;
 				}
 				// Read and cache packet from server.
-				var payload = await m_serverSession.ReceiveReplyAsync(m_ioBehavior, cancellationToken);
+				var payload = await m_serverSession.ReceiveReplyAsync(m_ioBehavior, cancellationToken).ConfigureAwait(false);
 				var segment = payload.ArraySegment;
 
 				if (segment.Count > NegotiateStreamConstants.MaxPayloadLength)
@@ -118,7 +118,7 @@ namespace MySqlConnector.Protocol.Serialization
 			return bytesRead;
 		}
 
-		public override int Read(byte[] buffer, int offset, int count) => ReadAsync(buffer, offset, count, m_cancellationToken).Result;
+		public override int Read(byte[] buffer, int offset, int count) => ReadAsync(buffer, offset, count, m_cancellationToken).GetAwaiter().GetResult();
 
 		public override async Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
 		{
@@ -183,12 +183,12 @@ namespace MySqlConnector.Protocol.Serialization
 				// full payload provided
 				payload = new PayloadData(new ArraySegment<byte>(buffer, offset, m_writePayloadLength));
 			}
-			await m_serverSession.SendReplyAsync(payload, m_ioBehavior, cancellationToken);
+			await m_serverSession.SendReplyAsync(payload, m_ioBehavior, cancellationToken).ConfigureAwait(false);
 			// Need to parse NegotiateStream header next time
 			m_writePayloadLength = 0;
 		}
 
-		public override void Write(byte[] buffer, int offset, int count) => WriteAsync(buffer, offset, count, m_cancellationToken).Wait();
+		public override void Write(byte[] buffer, int offset, int count) => WriteAsync(buffer, offset, count, m_cancellationToken).GetAwaiter().GetResult();
 
 		public override bool CanRead => true;
 
@@ -223,14 +223,24 @@ namespace MySqlConnector.Protocol.Serialization
 			using (var negotiateStream = new NegotiateStream(innerStream))
 			{
 				var targetName = GetServicePrincipalName(switchRequestPayloadData);
-				await negotiateStream.AuthenticateAsClientAsync(CredentialCache.DefaultNetworkCredentials, targetName);
-
+#if NETSTANDARD1_3
+				await negotiateStream.AuthenticateAsClientAsync(CredentialCache.DefaultNetworkCredentials, targetName).ConfigureAwait(false);
+#else
+				if (ioBehavior == IOBehavior.Synchronous)
+				{
+					negotiateStream.AuthenticateAsClient(CredentialCache.DefaultNetworkCredentials, targetName);
+				}
+				else
+				{
+					await negotiateStream.AuthenticateAsClientAsync(CredentialCache.DefaultNetworkCredentials, targetName).ConfigureAwait(false);
+				}
+#endif
 				if (innerStream.MySQLProtocolPayload.ArraySegment.Array != null)
 					// return already pre-read OK packet.
 					return innerStream.MySQLProtocolPayload;
 
 				// Read final OK packet from server
-				return await session.ReceiveReplyAsync(ioBehavior, cancellationToken);
+				return await session.ReceiveReplyAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
 			}
 		}
 	}
--- a/src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
+++ b/src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
@@ -1,0 +1,237 @@
+using System;
+using System.IO;
+using System.Net;
+using System.Net.Security;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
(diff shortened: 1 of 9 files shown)


src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
```diff
@@ -0,0 +1,237 @@
...
+
+				if (m_clientHandshakeDone)
+				{
+					// NegotiateStream protocol expects server to send "handshake done"
+					// empty message at the end of handshake.
+					CreateNegotiateStreamMessageHeader(buffer, offset, NegotiateStreamConstants.HandshakeDone, 0);
+					return NegotiateStreamConstants.HeaderLength;
+				}
+				// Read and cache packet from server.
+				var payload = await m_serverSession.ReceiveReplyAsync(m_ioBehavior, cancellationToken);
```
bgrainger:
```suggestion
				var payload = await m_serverSession.ReceiveReplyAsync(m_ioBehavior, cancellationToken).ConfigureAwait(false);
```

Should always use `ConfigureAwait(false)` to avoid potential deadlock; cf. https://twitter.com/bgrainger/status/938167570540007424

src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
```diff
@@ -0,0 +1,237 @@
...
+				var segment = payload.ArraySegment;
```
bgrainger: Be aware that the next `ReceiveReply` call can overwrite the contents of this buffer, so it can be dangerous to save it (e.g., in `m_readBuffer`). However, I think the usage I've seen in this class is safe.

src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
```diff
@@ -0,0 +1,237 @@
...
+			}
+			if (count > 0)
+			{
+				// Return cached data.
+				bytesRead += m_readBuffer.Read(buffer, offset, count);
+			}
+			return bytesRead;
+		}
+
+		public override int Read(byte[] buffer, int offset, int count) => ReadAsync(buffer, offset, count, m_cancellationToken).Result;
```
bgrainger:
```suggestion
		public override int Read(byte[] buffer, int offset, int count) => ReadAsync(buffer, offset, count, m_cancellationToken).GetAwaiter().GetResult();
```

`.Result` will throw any exception as an `AggregateException` which makes it harder for the caller to handle errors.

src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
```diff
@@ -0,0 +1,237 @@
...
+				var payloadBytes = m_writeBuffer.ToArray();
+				payload = new PayloadData(new ArraySegment<byte>(payloadBytes, 0, (int) m_writeBuffer.Length));
+				m_writeBuffer.SetLength(0);
+			}
+			else
+			{
+				// full payload provided
+				payload = new PayloadData(new ArraySegment<byte>(buffer, offset, m_writePayloadLength));
+			}
+			await m_serverSession.SendReplyAsync(payload, m_ioBehavior, cancellationToken);
```
bgrainger:
```suggestion
			await m_serverSession.SendReplyAsync(payload, m_ioBehavior, cancellationToken).ConfigureAwait(false);
```

src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
```diff
@@ -0,0 +1,237 @@
...
+			return Encoding.UTF8.GetString(reader.ReadNullOrEofTerminatedByteString());
+		}
+		public static async Task<PayloadData> AuthenticateAsync(byte[] switchRequestPayloadData,
+			ServerSession session, IOBehavior ioBehavior, CancellationToken cancellationToken)
+		{
+			using (var innerStream = new NegotiateToMySqlConverterStream(session, ioBehavior, cancellationToken))
+			using (var negotiateStream = new NegotiateStream(innerStream))
+			{
+				var targetName = GetServicePrincipalName(switchRequestPayloadData);
+				await negotiateStream.AuthenticateAsClientAsync(CredentialCache.DefaultNetworkCredentials, targetName);
```
bgrainger:
```suggestion
				await negotiateStream.AuthenticateAsClientAsync(CredentialCache.DefaultNetworkCredentials, targetName).ConfigureAwait(false);
```

src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
```diff
@@ -0,0 +1,237 @@
...
+
+				if (innerStream.MySQLProtocolPayload.ArraySegment.Array != null)
+					// return already pre-read OK packet.
+					return innerStream.MySQLProtocolPayload;
+
+				// Read final OK packet from server
+				return await session.ReceiveReplyAsync(ioBehavior, cancellationToken);
```
bgrainger:
```suggestion
				return await session.ReceiveReplyAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
```

src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
bgrainger: If `ioBehavior == IOBehavior.Synchronous`, it would be preferable to use the synchronous `negotiateStream.AuthenticateAsClient` method. This can prevent delays or deadlocks caused by thread pool exhaustion.

src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
```diff
@@ -0,0 +1,237 @@
...
+			{
+				// full payload provided
+				payload = new PayloadData(new ArraySegment<byte>(buffer, offset, m_writePayloadLength));
+			}
+			await m_serverSession.SendReplyAsync(payload, m_ioBehavior, cancellationToken);
+			// Need to parse NegotiateStream header next time
+			m_writePayloadLength = 0;
+		}
+
+		public override void Write(byte[] buffer, int offset, int count) => WriteAsync(buffer, offset, count, m_cancellationToken).Wait();
```
bgrainger: I think this can be `.GetAwaiter().GetResult()` too.

src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
vaintroub: Yes, the intention is just to have data of the last ReceiveReply

src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
vaintroub: OK, applied the fix in new commit.

src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
vaintroub: I also think it is safe, we only need data from the last ReceiveReply in m_readBuffer.

src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
vaintroub: Agree, fixed in last commit.

src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
vaintroub: Agree, fixed in last commit

src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
vaintroub: Agree, fixed in last commit

src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
vaintroub: Fixed.

src/MySqlConnector/Protocol/Serialization/AuthGSSAPI.cs
vaintroub: Fixed that too.

vaintroub:
> I assume you've tested this locally (by setting GSSAPIUser in config.json for the tests)? I think it should be possible to get it automatically tested as part of the Travis CI builds (for the mariadb Docker image); I can take a look at that after this is merged.

> 
Yes, I tested it this way, but just from Windows. On Linux, when I tried to run SideBySide tests, it complained about missing .NET Frameworks, and I ended up testing manually with a tiny console application, that did a connect and disconnect.

I believe the easiest way to automatically test is on Windows, at least for smoke tests, providing the server and client run on the same machine.  If both server and client run under domain user accounts, it will test Kerberos, if the machine is standalone, it would use NTLM.   I'm not very familiar with appveyor,  but our Java Connector uses it, for test, including automated MSI install https://github.com/MariaDB/mariadb-connector-j/blob/master/appveyor.yml , so maybe it is an option, too.

My own Kerberos setup is rather involved,  my KDC is  Windows Active Directory domain controller, with a Linux and Windows machines joined into the domain,  within in a Virtual Box host-only subnet. (I wanted to test same platform or cross-plattform combinations, therefore I made it complicated).

The single-machine Linux testing, with KDC, server and  client on the same box is something I did not do this time. But it looks like other people have useful scripts for that , e.g https://github.com/dotnet/corefx/blob/master/src/System.Net.Security/tests/Scripts/Unix/setup-kdc.sh

> Would GSS Service Principal Name (GssServicePrincipalName) be an appropriate connection string option? I see the command-line client uses gssapi-principal-name; I don't know if there's any "prior art" in other connector libraries. I have no objection to your updating this PR to include that setting if you think it would be good to support in the initial implementation.

I'd prefer ServerSPN (Server SPN) , because it is how SQLServer's ODBC and JDBC call it.
Let me submit another PR for this,  I need to figure how to  test it first;)

> (Is the verification implementation simply ensuring that AuthGSSAPI.GetServicePrincipalName() returns the same value as was specified in the connection string?)

I thought of something else than comparing strings. I think what is envisioned by Kerberos creators is that client knows the SPN without the help of the server. 

So what I would like to use instead, is to ignore whatever server has sent, use ServerSPN for targetName and throw exception if _NegotiateStream.IsMutuallyAuthenticated_ == false after AuthenticateAsClient.

Which would make ServerSPN unusable for NTLM-only on Windows, but I think it is fine if we document that ServerSPN would fail on NTLM(where the targetName parameter for AuthenticateAsClient can be anything anyway)

tests/MySqlConnector.Tests/MySqlConnectionStringBuilderTests.cs
```diff
@@ -58,6 +58,7 @@ public void Defaults()
 #if !BASELINE
 			Assert.Null(csb.ServerRsaPublicKeyFile);
 #endif
+			Assert.Null(csb.ServerSPN);
```
bgrainger: These tests also run against Oracle's Connector/NET (to verify equivalent behaviour) so this line will need to be moved inside the `#if !BASELINE` block to avoid a compiler error with MySqlData.

tests/MySqlConnector.Tests/MySqlConnectionStringBuilderTests.cs
```diff
@@ -121,6 +122,7 @@ public void ParseConnectionString()
 					"Port=1234;" +
 					"protocol=pipe;" +
 					"pwd=Pass1234;" +
+					"server spn=mariadb/host.example.com@EXAMPLE.COM;" +
```
bgrainger: Move up to line 111 (to be inside `#if !BASELINE`).

tests/MySqlConnector.Tests/MySqlConnectionStringBuilderTests.cs
```diff
@@ -168,6 +170,7 @@ public void ParseConnectionString()
 			Assert.False(csb.Pooling);
 			Assert.Equal(1234u, csb.Port);
 			Assert.Equal("db-server", csb.Server);
+			Assert.Equal("mariadb/host.example.com@EXAMPLE.COM", csb.ServerSPN);
```
bgrainger: Move up to line 159 (to be inside `#if !BASELINE`).

vaintroub:
I see you fixed the failing appveyor in https://github.com/mysql-net/MySqlConnector/pull/579 . Thanks for explanation, now I see what  #if BASELINE was for.

bgrainger:
Merged via #579.

vaintroub:
Thank you!

bgrainger:
Shipped in [0.47.0](https://github.com/mysql-net/MySqlConnector/releases/tag/0.47.0).

Issue #374, opened 2017-10-22
Support restrictions for GetSchema


bgrainger:
The `MySqlConnection.GetSchema(string, string[])` overload should support a user-specified list of [restrictions](https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/schema-restrictions) on the schema being requested.

`GetSchema("Restrictions")` should return a `DataTable` containing all the available restrictions.

Not implemented as part of #361.

dechristo:
Hi @bgrainger! Can I work on this one? and also on [#494](https://github.com/mysql-net/MySqlConnector/issues/494)

bgrainger:
Yes. Thanks!

I think the implementation should be fairly straightforward in `SchemaProvider.cs`, with a little change in `MySqlConnection.cs` to pass the restrictions through.

dechristo:
Hi @bgrainger !
I opened this [PR](https://github.com/mysql-net/MySqlConnector/pull/616) already because I think it will be easier for you to point my mistakes and also for me to solve my doubts instead of message ping pong.

Thanks for your attention so far!

phillipc:
Hello everybody, we have implemented the restrictions in https://github.com/mysql-net/MySqlConnector/pull/1188

bgrainger:
Implemented in https://github.com/mysql-net/MySqlConnector/pull/1190.

PR #616, opened 2019-03-07
Implement #374 - Support restrictions for GetSchema

dechristo:


src/MySqlConnector/Core/SchemaProvider.cs
```diff
@@ -33,6 +34,35 @@ public DataTable GetSchema(string collectionName)
...
+		public DataTable GetSchema(string collectionName, string[] restrictions)
+		{
+			Action close = null;
+			if (m_connection.State != ConnectionState.Open)
+			{
+				m_connection.Open();
+				close = m_connection.Close;
+			}
+
+			var dataTable = m_connection.GetSchema(collectionName, restrictions);
```
bgrainger: I think this will introduce infinite recursion between this method and `MySqlConnection.GetSchema`.

Please add tests that verify the correct information is being returned, for example like https://github.com/mysql-net/MySqlConnector/blob/5758207a81fad9c77494995316fcfc4687f04bb7/tests/SideBySide/QueryTests.cs#L1147-L1148

src/MySqlConnector/Core/SchemaProvider.cs
```diff
@@ -33,6 +34,35 @@ public DataTable GetSchema(string collectionName)
 			return dataTable;
 		}
 
+		public DataTable GetSchema(string collectionName, string[] restrictions)
```
bgrainger: I don't think introducing a new method should be necessary; `string[] restrictions` could be added as a new parameter to the existing method.

src/MySqlConnector/Core/SchemaProvider.cs
```diff
@@ -33,6 +34,35 @@ public DataTable GetSchema(string collectionName)
...
+		{
+			Action close = null;
+			if (m_connection.State != ConnectionState.Open)
+			{
+				m_connection.Open();
+				close = m_connection.Close;
+			}
```
bgrainger: I think this should just rely on the connection already being open, instead of opening and closing the connection (unless this is required for compatibility with other drivers).

src/MySqlConnector/Core/SchemaProvider.cs
dechristo: Agreed. Can I make the default value `null` then?

src/MySqlConnector/Core/SchemaProvider.cs
dechristo: Ok! Will change it.

src/MySqlConnector/Core/SchemaProvider.cs
dechristo: Yeap! I will add all the tests once I get things right. I am just confused on how to get the restrictions now if this will lead to a infinite recursion. Could you shed some light, please?

src/MySqlConnector/Core/SchemaProvider.cs
bgrainger: Yes, that sounds fine.

src/MySqlConnector/Core/SchemaProvider.cs
bgrainger: `MySqlConnection.GetSchema` calls this method: https://github.com/mysql-net/MySqlConnector/blob/a6b5d69c6640d5de0b15051f6e734bb4ab75c962/src/MySqlConnector/MySql.Data.MySqlClient/MySqlConnection.cs#L311

(Or it will do, once the overload has changed.)

So this method shouldn't call back to `MySqlConnection.GetSchema`; it should just do the work.

src/MySqlConnector/Core/SchemaProvider.cs
```diff
@@ -33,6 +34,21 @@ public DataTable GetSchema(string collectionName)
...
+					new DataColumn("ParameterName", typeof(string)),
+					new DataColumn("RestrictionDefault", typeof(string)),
+					new DataColumn("RestrictionNumber", typeof(int))
+				});
+
+				dataTable.Rows.Add("Catalog", "@Catalog", "TABLE_CATALOG", 1);
+				dataTable.Rows.Add("Owner", "@Owner", "TABLE_SCHEMA", 2);
+				dataTable.Rows.Add("Table", "@Name", "TABLE_NAME", 3);
+				dataTable.Rows.Add("TableType", "@TableType", "TABLE_TYPE", 4);
+		}
```
bgrainger: The `Restrictions` schema should have four columns: `CollectionName`, `RestrictionName`, `RestrictionDefault`, `RestrictionNumber`: https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/common-schema-collections#restrictions

This table is filled with the names of restrictions that are supported by _other_ schema tables.

For example, to indicate three restrictions on the `Tables` schema, you would add these rows:



CollectionName | RestrictionName | RestrictionDefault | RestrictionNumber
-- | -- | -- | --
Tables | Schema |   | 1
Tables | Table |   | 2
Tables | TableType |   | 3

While it's not OK to bring code from Connector/NET into this project, it is totally fine to perform "black box" testing and see what (for example) `connection.GetSchema("Restrictions")` returns when executed against that library. This code should return a subset of what it supports.

bgrainger:
To fix the DCO errors, please attest that you agree to the [Developer Certificate of Origin](https://developercertificate.org/) (that is, you have the right to submit this code) by adding the line `Signed-off-by: Your Name <youremail@example.com>` to the end of your commit message.

More details here: https://probot.github.io/apps/dco/

dechristo:
Hi @bgrainger. I am still lost on how to proceed with the restrictions array.

bgrainger:
@dechristo I would lean towards making this all data-driven. Some ideas:

* `m_schemaCollections` becomes `Dictionary<string, SchemaInfo>`.
* `SchemaInfo` is a new class with `Action<DataTable, string[]> FillSchema` and `string[] RestrictionNames` properties.
* `RestrictionNames` is a simple array of the restriction names supported by a collection.
* `FillSchema` takes the restriction values (as a second parameter) and applies them appropriately. This will probably be adding a `WHERE` clause with parameters in most cases.
* The "Restrictions" schema is filled by enumerating `m_schemaCollections` and filling the table with its `RestrictionNames`.

KristianWedberg:
@bgrainger Hi, is the restrictions overload actually working now, returning correctly filtered results, or is this a won't-fix?

bgrainger:
@KristianWedberg The issue is still open: https://github.com/mysql-net/MySqlConnector/issues/374

I'm closing out this three-year-old PR as the author seems to have abandoned it, and it would need some work to fix all the merge conflicts with `master`.

PR #706, opened 2019-09-25
Explicitly cast double values down to float in Row.GetFloat().

lauxjpn:
Support casting to float for Pomelo.EntityFramworkCore.MySql for MySQL < 8.0.17.
See https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/827#issuecomment-534745205

Signed-off-by: Laurents Meyer <laucomm@gmail.com>

### DIFF
--- a/src/MySqlConnector/Core/Row.cs
+++ b/src/MySqlConnector/Core/Row.cs
@@ -353,9 +353,15 @@ namespace MySqlConnector.Core
 		public float GetFloat(int ordinal)
 		{
 			var value = GetValue(ordinal);
-			return value is double doubleValue ? (float) doubleValue :
-				value is decimal decimalValue ? (float) decimalValue :
-				(float) value;
+
+			// Loss of precision is expected, significant loss of information is not.
+			// Use explicit range checks to guard against that.
+			return value switch
+			{
+				double doubleValue => (doubleValue >= float.MinValue && doubleValue <= float.MaxValue ? (float) doubleValue : throw new InvalidCastException("The value cannot be safely cast to Single.")),
+				decimal decimalValue => (float) decimalValue,
+				_ => (float) value
+			};
 		}
 
 		public MySqlDateTime GetMySqlDateTime(int ordinal)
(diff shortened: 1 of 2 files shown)


lauxjpn:
If you are concerned, that the explicit guard against information loss could lead to an `InvalidCastException` in cases, where the double value is even only a tiny bit larger/smaller than the `float.MinValue`/`float.MaxValue`, it could be dropped entirely.
But it would also mean, that a way higher or lower double value would silently be converted to a `float.MinValue`/`float.MaxValue`, which might be unexpected.

src/MySqlConnector/Core/Row.cs
```diff
@@ -353,8 +353,15 @@ public double GetDouble(int ordinal)
...
 		{
 			var value = GetValue(ordinal);
-			return value is decimal decimalValue ? (float) decimalValue :
-				(float) value;
+
+			// Loss of precision is expected, significant loss of information is not.
+			// Use explicit range checks to guard against that.
+			return value switch
+			{
+				double doubleValue => (doubleValue >= float.MinValue && doubleValue <= float.MaxValue ? (float) doubleValue : throw new InvalidCastException("The value cannot be safely cast to Single.")),
```
bgrainger: This test could be if `doubleValue` is contained in any of the following ranges: `[float.MinValue, -float.Epsilon], [0, 0], [float.Epsilon, float.MaxValue]`; what do you think?

lauxjpn: So the difference would be, that double values between -/+ `float.Epsilon` and `0` would throw.

The question is, if it is expected that a `double` between `float.Epsilon` and `0` being converted to `0`. I think it is, because it's just a result of precision-loss.

But there could probably made a case against it as well, which would expect those values to end up as `float.Epsilon`. I think that is imprecise though, because then it would need be rounded to either `float.Epsilon` or `0`, whatever is closer. And I would assume, that the C#/ the processor is doing exactly that in the first place.

I would not throw in any of these cases, because this would be unexpected and a bit random.

bgrainger:
Fixed in [0.59.0](https://github.com/mysql-net/MySqlConnector/releases/tag/0.59.0).

PR #707, opened 2019-09-30
GetBoolean might be called on a decimal

lauxjpn:
The following test (and 1 more) currently fails in Pomelo.MySql:
https://github.com/aspnet/EntityFrameworkCore/blob/817aaea6ca1a24a29e8e9f3a7b2dd3748956d942/test/EFCore.Specification.Tests/Query/GearsOfWarQueryTestBase.cs#L1503-L1519

It gets translated to:
```sql
SELECT COALESCE(`w0`.`IsAutomatic`, FALSE)
FROM `Weapons` AS `w`
LEFT JOIN `Weapons` AS `w0` ON `w`.`SynergyWithId` = `w0`.`Id`
ORDER BY `w`.`Id`
```
It seems that `COALESCE` is returning a `decimal` for `IsAutomatic`, though the column is actually of type `bit(1)`. The code expects a `bool` and calls `GetBoolean`, which then throws.

This fix allows the explicit conversion from `decimal` to `bool` within `Row.GetBoolean()`.

This is not critical, so we don't need a quick release for this.

### DIFF
--- a/src/MySqlConnector/Core/Row.cs
+++ b/src/MySqlConnector/Core/Row.cs
@@ -61,6 +61,8 @@ namespace MySqlConnector.Core
 				return (long) value != 0;
 			if (value is ulong)
 				return (ulong) value != 0;
+			if (value is decimal)
+				return (decimal) value != 0;
 			return (bool) value;
 		}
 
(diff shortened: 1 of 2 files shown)


bgrainger:
FWIW, the promotion of the `BIT(n)` data type to `DECIMAL` when using `COALESCE` is another reason why I would strongly recommend `BOOL` (aka `TINYINT(1)`) as the preferred MySQL data type for Boolean data, not `BIT(1)`.

The PR looks good, I will merge it.

PR #711, opened 2019-10-01
Code cleanup

brunao05:


src/MySqlConnector/Core/CachedProcedure.cs
```diff
@@ -23,13 +23,13 @@ internal sealed class CachedProcedure
 			{
 				try
 				{
-					using var cmd = connection.CreateCommand();
+					await using var cmd = connection.CreateCommand();
```
bgrainger: This isn't supported before `netstandard2.1`, and isn't useful because `MySqlCommand` doesn't use asynchronous I/O when disposing.

src/MySqlConnector/Core/ResultSet.cs
```diff
@@ -55,7 +52,7 @@ public async Task ReadResultSetHeaderAsync(IOBehavior ioBehavior)
 						RecordsAffected = (RecordsAffected ?? 0) + ok.AffectedRowCount;
 						LastInsertId = unchecked((long) ok.LastInsertId);
 						WarningCount = ok.WarningCount;
-						if (ok.NewSchema is object)
+						if (ok.NewSchema != null)
```
bgrainger: `is object` is preferred as a non-nullability test.

https://twitter.com/jaredpar/status/1115019017297596416

src/MySqlConnector/Core/TypeMapper.cs
```diff
@@ -120,19 +120,19 @@ private TypeMapper()
 
 		public MySqlDbType GetMySqlDbTypeForDbType(DbType dbType)
 		{
-			foreach (var pair in m_mySqlDbTypeToColumnTypeMetadata)
+			foreach (var pair in m_mySqlDbTypeToColumnTypeMetadata.Where(pair => pair.Value.DbTypeMapping.DbTypes.Contains(dbType)))
```
bgrainger: LINQ is less efficient than a simple `foreach` / `if` test; please avoid.

bgrainger:
It looks like you might have just run some automatic code formatter over the code. This just unnecessarily obfuscates `git blame` and doesn't provide any actual enhancement. In some cases, it's made the code less efficient.

PR #718, opened 2019-10-15
CI: Use windows MySQL docker image

caleblloyd:
I created a Windows Docker Image for Pomelo for CI.

Opening this in the chance you want to use it as well.  Only real benefit is that it's an immutable image coming from DockerHub as opposed to a download link from oracle.

Windows Docker Images do come with a downside.  Azure Build Agents only support process isolation (not hyper-v Isolation) so images [compatibility](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility) is tied to the Host OS.  This specific image works with Windows Server 2019.  The Docker Image is open source at [PomeloFoundation/MySql-Windows-Docker](https://github.com/PomeloFoundation/MySql-Windows-Docker) and it will be easy to update for new versions of Windows Server.

Base layers from `mcr.microsoft.com/windows/servercore:ltsc2019` are [cached](https://github.com/microsoft/azure-pipelines-image-generation/blob/master/images/win/Vs2019-Server2019-Readme.md#docker-images) on the Azure DevOps Windows 2019 build server already, so it will not incur the ~2GB download penalty for that base image.  It will just download the layers that contain MySQL and configuration.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -13,7 +13,7 @@ jobs:
 - job: windows_build
   displayName: 'Windows Build'
   pool:
-    vmimage: 'vs2017-win2016'
+    vmimage: 'windows-2019'
   steps:
   - template: '.ci/build-steps.yml'
   - task: DotNetCoreCLI@2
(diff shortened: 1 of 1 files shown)


caleblloyd:
Looks like it's slower - `3m24s` with the container vs `1m58s` without it

Also looks like there are some issues with the Baseline tests against the container.  May be `my.ini` settings or something else, not entirely sure.

I'll abandon for now, maybe revisit in the future if I have the resources to test locally/tune.

bgrainger:
Thanks for getting the ball rolling on this. The performance regression is disappointing, but I might use this as a template when working on this in the future.

(It's currently installing an old version of MySQL Server, but upgrading it is a pain. Docker would be much nicer.)

PR #725, opened 2019-10-21
Optimise GetInt32

bgrainger:
Proof-of-concept to see if there's value in providing an efficient implementation of various `GetX` methods.


bgrainger:
Implemented in [0.62.0](https://github.com/mysql-net/MySqlConnector/releases/tag/0.62.0).

PR #727, opened 2019-10-21
Add unit of measure and default to ConnectionTimeout

JosephAmalfitanoSSA:
The summary is what is shown by consumers  when using autocomplete.  This change allows consumers to easily see the unit of measure (seconds) as well as the default value when using autocomplete/code completion. 

I'm proposing this change because it wasn't clear to me if the timeout was in seconds or milliseconds. Also, i had no idea what the default is/was until I added a breakpoint in my code that was consuming this.


JosephAmalfitanoSSA:
This picture shows the current state of the summary/tooltip. My proposed change will make it clear that the unit of measure is in seconds and the default timeout is 15 seconds. 

image

bgrainger:
Thanks for the contribution!

FYI, several of these properties are initialised from the connection string (or its defaults), and even though they don't have XML documentation, they are all described here: https://mysqlconnector.net/connection-options/

bgrainger:
Added in [0.60.2](https://github.com/mysql-net/MySqlConnector/releases/tag/0.60.2).

PR #756, opened 2020-01-07
SqlBulkCopy throws exception when output byte buffer is too small

laurensknoll:
This pull requests resolves an issue I encountered when bulk-copying data: Output byte buffer is too small when writing string values.

The code used [Encoding.GetBytes()](https://github.com/mysql-net/MySqlConnector/blob/8e9ea2739d2a119825c7afd90d45f693561c3c83/src/MySqlConnector/MySql.Data.MySqlClient/MySqlBulkCopy.cs#L378) which throws this exception when the output buffer is unable to hold the encoded value-bytes.

In this PR I added a GetByteCount call to prevent this Exception.

Note: I also considered rewriting the code to [Encoding.GetEncoder().Convert()](https://docs.microsoft.com/en-us/dotnet/api/system.text.encoder.convert?view=netcore-3.1#System_Text_Encoder_Convert_System_ReadOnlySpan_System_Char__System_Span_System_Byte__System_Boolean_System_Int32__System_Int32__System_Boolean__). I've chosen to stick to the GetBytes method to prevent regression issues, despite the introduced performance decrease.

### DIFF
--- a/src/MySqlConnector/MySql.Data.MySqlClient/MySqlBulkCopy.cs
+++ b/src/MySqlConnector/MySql.Data.MySqlClient/MySqlBulkCopy.cs
@@ -350,48 +350,6 @@ namespace MySql.Data.MySqlClient
 				}
 			}
 
-			static bool WriteString(string value, Span<byte> output, out int bytesWritten)
-			{
-				var index = 0;
-				bytesWritten = 0;
-				while (index < value.Length)
-				{
-					if (Array.IndexOf(s_specialCharacters, value[index]) != -1)
-					{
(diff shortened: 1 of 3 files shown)


bgrainger:
Thanks for the submission. I believe I already fixed this (in exactly the same way) here: ebc2c2fe362359747261300ac07933872c9831e8 (it just hasn't shipped yet).

bgrainger:
Fix shipped in [0.62.0-beta6](https://www.nuget.org/packages/MySqlConnector/0.62.0-beta6).

PR #768, opened 2020-02-05
Update list of ORMs

exileDev:
Added LINQ to DB ORM


bgrainger:
Thanks for the addition!

Issue #721, opened 2019-10-20
Implement optimised MySqlDataReader.GetX methods


bgrainger:
`MySqlDataReader.GetInt32(int ordinal)` is essentially implemented as `return (int) GetValue(ordinal);`

This boxes the `int` as an object then unboxes it. It would be more efficient to return the `int` value directly and would eliminate allocations. (Coupled with 00199a0248bf436dd77e3f517d12a8fde2678d63 this would also optimise `GetFieldValue<int>`.)

For the text protocol, it might be possible to implement this without significant code duplication if each `GetX` method just needs to check for an integral column type, parse the text, and check for overflow.

A quick-and-dirty implementation yielded the following benchmark results:

### .NET Core 3.0:

|        Method |     Mean |     Error |    StdDev |   Median |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------- |---------:|----------:|----------:|---------:|-------:|------:|------:|----------:|
|      GetInt32 | 21.21 ns | 0.4360 ns | 0.6112 ns | 21.06 ns |      - |     - |     - |         - |
|      GetValue | 30.96 ns | 0.7450 ns | 1.4878 ns | 30.56 ns | 0.0051 |     - |     - |      24 B |
| GetFieldValue | 26.04 ns | 0.5446 ns | 1.0228 ns | 25.57 ns |      - |     - |     - |         - |

Before, `GetInt32` and `GetFieldValue<int>` were 31-37ns and also allocating.

### .NET 4.8

|        Method |     Mean |     Error |    StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|-------------- |---------:|----------:|----------:|-------:|-------:|------:|----------:|
|      GetInt32 | 54.19 ns | 1.0760 ns | 1.1513 ns |      - |      - |     - |         - |
|      GetValue | 57.13 ns | 1.1467 ns | 1.6076 ns | 0.0050 | 0.0001 |     - |      24 B |

The improvement is less pronounced here, with the method taking ~59ns before the change.

bgrainger:
Optimising `GetInt32` specifically would be useful for the TechEmpower Framework Benchmarks: https://github.com/TechEmpower/FrameworkBenchmarks/blob/a94705893dd6a0e7e8312c79d0e4dff7ba8eb990/frameworks/CSharp/aspnetcore/PlatformBenchmarks/Data/RawDb.cs#L49-L50

bgrainger:
[...]

PR #783, opened 2020-03-22
Allow reading `bool` as integer, like `MySql.Data` allows (Fixes #782)

### LINKED ISSUE #782: Can't convert Bool to Int32 (TINYINT)
With Oracle's `MySql.Data`, when you read a TINYINT column - you can read it as a `bool` or as an `Int32` - either way works.

And it only makes sense as `TINYINT` is not explicitly a boolean, it's just so tiny that it can only hold 0 or 1... And it's called TINY**INT**.

Seems like an easy fix.

It currently throws: `Can't convert Bool to Int32`
From: `MySqlConnector\Core\Row.cs:line 224`

For me - this was an unexpected breaking change when trying the transition to this library.

danielgindi:


src/MySqlConnector/Core/Row.cs
```diff
@@ -207,16 +209,15 @@ public int GetInt32(int ordinal)
...
 				columnDefinition.ColumnType != ColumnType.Short &&
 				columnDefinition.ColumnType != ColumnType.Int24 &&
 				columnDefinition.ColumnType != ColumnType.Long &&
 				columnDefinition.ColumnType != ColumnType.Longlong &&
 				columnDefinition.ColumnType != ColumnType.Bit &&
 				columnDefinition.ColumnType != ColumnType.Year &&
 				columnDefinition.ColumnType != ColumnType.Decimal &&
-				columnDefinition.ColumnType != ColumnType.NewDecimal) ||
-				(columnDefinition.ColumnType == ColumnType.Tiny && Connection.TreatTinyAsBoolean && columnDefinition.ColumnLength == 1 && (columnDefinition.ColumnFlags & ColumnFlags.Unsigned) == 0))
+				columnDefinition.ColumnType != ColumnType.NewDecimal)
```
bgrainger: I suspect this fails to coerce "bool" values to 0 or 1.

bgrainger: Fixed: 693b8e0b54fad79ff076b85ae44f34b9d1042428

danielgindi:
Great!

Btw how do you explain why GetInt32 is so different the implementation of other int getters?
It seems like it’s duplicating the code of GetValue, adding a lot of checks and manually parsing the raw data.

bgrainger:
See https://github.com/mysql-net/MySqlConnector/issues/721 for the rationale.

Issue #664, opened 2019-06-28
GetDouble and GetFloat less type auto conversion


dczuo:
simple example:
using (var reader = new MySqlCommand("select 1 / 3 from tablename", conn).ExecuteReader())
{
	reader.Read();
	reader.GetDouble(0);    // InvalidCastException
	reader.GetFloat(0);     // InvalidCastException
}

bgrainger:
MySqlConnector currently only performs a small number of known "safe" conversions; see #481 and #406. 

Note that a `decimal` column can be retrieved as an int (#54) so it may make sense to allow implicit conversion to `float`/`double` (which would be less lossy).

Like similar issues, this needs testing in AdoNetApiTest first.

bgrainger:
Added to AdoNetApiTest: https://github.com/mysql-net/AdoNetApiTest/commit/2a48acc59148b1f299d16cbb19c440fa068e9527

Conversions from a decimal column (or implicit decimal result) will be handled automatically by `GetDouble` and `GetFloat`.

bgrainger:
Fix shipped in [0.57.0-rc1](https://www.nuget.org/packages/MySqlConnector/0.57.0-rc1).

dczuo:
thanks

PR #785, opened 2020-03-24
Have the same casts in GetDecimal as in GetDouble

danielgindi:
Support casts to `decimal` from `float`/`double`.

Rationale:
It is already supported for the other way.  
Casting from `decimal` to `float`/`double` could result in data loss.  
If that is supported, so why not the other way, where there's no data loss?

### DIFF
--- a/src/MySqlConnector/Core/Row.cs
+++ b/src/MySqlConnector/Core/Row.cs
@@ -350,7 +350,13 @@ namespace MySqlConnector.Core
 
 		public string GetString(int ordinal) => (string) GetValue(ordinal);
 
-		public decimal GetDecimal(int ordinal) => (decimal) GetValue(ordinal);
+		public decimal GetDecimal(int ordinal)
+		{
+			var value = GetValue(ordinal);
+			return value is float floatValue ? (decimal) floatValue :
+				value is double decimalValue ? (decimal) decimalValue :
+				(decimal) value;
+		}
 
 		public double GetDouble(int ordinal)
 		{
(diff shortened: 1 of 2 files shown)


bgrainger:
Seems reasonable.

The conversions the other way were added to fix #664.

(Note that `decimal` has a much smaller range than a `double` or even a `float`, so some conversions might result in an `OverflowException` for extremely large or small values, but that would be expected.)

bgrainger:
I'm not sure why the Netlify checks are failing, but they don't need to block this PR.

Thanks for the fix!

danielgindi:
> (Note that `decimal` has a much smaller range than a `double` or even a `float`, so some conversions might result in an `OverflowException` for extremely large or small values, but that would be expected.)

It is true and untrue at the same time :-)
`decimal` has precision of up to 128 bits, while `double` is 64 bits and `float` is 32 bits.  
So `decimal` stores more information, up to more digits before/after the decimal point.
But for math calculations - `decimal` will lose precision because it contains the actual digits, not the parts that make the actual number. So fractions in `float`/`double` could continue forever (i.e 1.333333333333333333333) while in decimal they have a fixed count.

In cases where the user cares about "money" precision, they can safely cast to `decimal`, and keep all the information. Casting back from `decimal` to `float` in this case will have a different number than the original `float`, but that is true when casting directly from `decimal` to `float`.

So there's potential data-loss in casting both ways, and it always sums up to what you intend to do with the number.

bgrainger:
> So fractions in float/double could continue forever 

Nitpick: this isn't actually true; `float` and `double` always have a finite decimal expansion (accessible using the [`"R"` roundtrip format](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings?view=netcoreapp-3.1#the-round-trip-r-format-specifier)) and need at most 9 or 17 decimal digits (respectively). They do not store rational numbers precisely.

I think `decimal` always has more precision than a `double` (this is not something I've looked into), but it doesn't have as large a range: `decimal` can go up to 7.9228 x 10²⁸, while `float` can go to 3.4 x 10³⁸ (and `double` even greater).

danielgindi:
@bgrainger Let's agree that they weirdly mismatch in all categories :-)

bgrainger:
Fixed in [0.63.0](https://github.com/mysql-net/MySqlConnector/releases/tag/0.63.0); thanks for the PR!

PR #791, opened 2020-03-27
Allows the use of snapshot isolation level on transactions

battyejp:
Allows the use of using the isolation level snapshot for transactions, as per https://dev.mysql.com/doc/refman/5.7/en/commit.html

### DIFF
--- a/src/MySqlConnector/Core/StandardEnlistedTransaction.cs
+++ b/src/MySqlConnector/Core/StandardEnlistedTransaction.cs
@@ -29,9 +29,9 @@ namespace MySqlConnector.Core
 				_ => "repeatable read",
 			};
 
-			var cosistentSnapshotText = Transaction.IsolationLevel == IsolationLevel.Snapshot ? " with consistent snapshot" : string.Empty;
-			using var cmd = new MySqlCommand($"set transaction isolation level {isolationLevel}; start transation{cosistentSnapshotText}", Connection);
-			cmd.ExecuteNonQuery();
+			var consistentSnapshotText = Transaction.IsolationLevel == IsolationLevel.Snapshot ? " with consistent snapshot" : string.Empty;
+			using (var cmd = new MySqlCommand($"set transaction isolation level {isolationLevel}; start transaction{consistentSnapshotText};", Connection))
+				cmd.ExecuteNonQuery();
 		}
 
 		protected override void OnPrepare(PreparingEnlistment enlistment)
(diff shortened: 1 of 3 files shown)


tests/SideBySide/Transaction.cs
```diff
@@ -50,6 +52,50 @@ public void DbConnectionCommit()
...
+		[InlineData(IsolationLevel.ReadCommitted, "read committed")]
+		[InlineData(IsolationLevel.RepeatableRead, "repeatable read")]
+		[InlineData(IsolationLevel.Serializable, "serializable")]
+		[InlineData(IsolationLevel.Snapshot, "repeatable read")]
+		[InlineData(IsolationLevel.Unspecified, "repeatable read")]
+		public void DbConnectionIsolationLevel(IsolationLevel inputIsolationLevel, string expectedTransactionIsolationLevel)
+		{
+			DbConnection connection = m_connection;
+			m_connection.Execute(@"set global log_output = 'table';");
+			m_connection.Execute(@"set global general_log = 1;");
```
bgrainger: Thanks for adding tests for the new behaviour!

Since multiple tests can be run in parallel, is it possible that `set global general_log` is capturing log statements from other tests running simultaneously on other connections? You might be able to filter by `thread_id` to this connection's log output (using `MySqlConnection.ServerThread`).

src/MySqlConnector/Core/StandardEnlistedTransaction.cs
```diff
@@ -21,16 +21,17 @@ protected override void OnStart()
...
 
 				// "In terms of the SQL:1992 transaction isolation levels, the default InnoDB level is REPEATABLE READ." - http://dev.mysql.com/doc/refman/5.7/en/innodb-transaction-model.html
 				IsolationLevel.Unspecified => "repeatable read",
 				_ => "repeatable read",
 			};
 
-			using var cmd = new MySqlCommand("set transaction isolation level " + isolationLevel + "; start transaction;", Connection);
-			cmd.ExecuteNonQuery();
+			var consistentSnapshotText = Transaction.IsolationLevel == IsolationLevel.Snapshot ? " with consistent snapshot" : string.Empty;
+			using (var cmd = new MySqlCommand($"set transaction isolation level {isolationLevel}; start transaction{consistentSnapshotText};", Connection))
```
bgrainger: No need to revert the C# 8 using declaration here.

tests/SideBySide/Transaction.cs
battyejp: Thanks for the tip, will try it out

battyejp: Are you happy with this now the tests are passing?

bgrainger: Sorry, I didn't see that the PR was updated and passing. Thanks for pinging me; I'll check it out soon.

bgrainger:
Thanks for the contribution!

bgrainger:
Added in [0.63.2](https://github.com/mysql-net/MySqlConnector/releases/tag/0.63.2).

PR #795, opened 2020-04-05
Add support for DataSourceInformation Schema

vdaron:
Some progress on the #375 issue

### DIFF
--- a/src/MySqlConnector/Core/SchemaProvider.cs
+++ b/src/MySqlConnector/Core/SchemaProvider.cs
@@ -4,8 +4,6 @@ using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
 using System.Linq;
-using System.Text;
-using System.Text.RegularExpressions;
 using MySql.Data.MySqlClient;
 
 namespace MySqlConnector.Core
--- a/src/MySqlConnector/Core/SchemaProvider.cs
+++ b/src/MySqlConnector/Core/SchemaProvider.cs
@@ -84,9 +84,9 @@ namespace MySqlConnector.Core
 
 			var row = dataTable.NewRow();
 			row["CompositeIdentifierSeparatorPattern"] = @"\.";
-			row["DataSourceProductName"] = "MysqlConnector";
+			row["DataSourceProductName"] = "MySQL";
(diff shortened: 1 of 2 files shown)


src/MySqlConnector/Core/SchemaProvider.cs
```diff
@@ -58,6 +60,54 @@ public DataTable GetSchema(string collectionName)
...
+				new DataColumn("QuotedIdentifierCase", typeof(IdentifierCase)),
+				new DataColumn("ParameterNamePattern", typeof(string)),
+				new DataColumn("StatementSeparatorPattern", typeof(string)),
+				new DataColumn("StringLiteralPattern", typeof(string)),
+				new DataColumn("SupportedJoinOperators", typeof(SupportedJoinOperators))
+			});
+
+			var row = dataTable.NewRow();
+			row["CompositeIdentifierSeparatorPattern"] = @"\.";
+			row["DataSourceProductName"] = "MysqlConnector";
```
bgrainger, line 87: `DataSourceProductName` is "The name of the product accessed by the provider, such as 'Oracle' or 'SQLServer'." (https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/common-schema-collections)

I don't know if we can accurately determine the remote database (MySQL, Percona, MariaDB, Amazon, etc.) so it might be best to just return `"MySQL"`.

bgrainger:
Thanks for the contribution! I'll look over it shortly.

src/MySqlConnector/Core/SchemaProvider.cs
```diff
@@ -58,6 +60,54 @@ public DataTable GetSchema(string collectionName)
...
+			row["DataSourceProductVersion"] = m_connection.ServerVersion;
+			row["DataSourceProductVersionNormalized"] = m_connection.Session.ServerVersion.Version.ToString();
```
bgrainger, line 89: > A normalized version for the data source, such that it can be compared with `String.Compare()`.

There would need to be some padding here (e.g., `nn.nn.nnnn`) to make sure 10 doesn't compare between 1 and 2.

bgrainger:
Added in [0.63.2](https://github.com/mysql-net/MySqlConnector/releases/tag/0.63.2).

PR #807, opened 2020-04-28
Get COLUMN_TYPE with FillDataTable

mitchydeath:
Changed to get COLUMN_TYPE with GetSchema("COLUMNS").
Because get the signed or unsigned.


bgrainger:
I removed this in 7d34b3b917793d46e4afe69c7043073a7c4fafd0 just to sync up with the list of columns returned by Connector/NET, but have no objection to including it if it's useful.

mitchydeath:
I see.
I think this is the best way to get this value(signed or unsigned).
So I would like to know if there is another good method.

bgrainger:
Actually, it looks like Connector/NET provides `COLUMN_TYPE`, and removing it was an error.

I don't know of a better way to detect signed/unsigned.

bgrainger:
Fixed in [0.64.2](https://github.com/mysql-net/MySqlConnector/releases/tag/0.64.2); thanks for the contribution!

Issue #70, opened 2016-10-01
Implement spatial data types


bgrainger:
Implement support for spatial data types: `GEOMETRY`, `POINT`, `LINESTRING`, `POLYGON`, `MULTIPOINT`, `MULTILINESTRING`, `MULTIPOLYGON`, `GEOMETRYCOLLECTION`.

http://dev.mysql.com/doc/refman/5.7/en/spatial-datatypes.html

Required for #4.

caleblloyd:
If we implement these types, I think we'll have a lot of maintenance on upkeep of the types, standardizing against MySql.Data, compatibility with third party libraries, etc.

What are your thoughts on re-introducing reflection-based conversion for datatypes that don't have hard 1:1 mappings to CLR types:

[Pomelo.EntityFrameworkCore.MySql/Storage/SynchronizedMySqlDataReader.cs#L179](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/src/Pomelo.EntityFrameworkCore.MySql/Storage/SynchronizedMySqlDataReader.cs#L179
)

This would allow for implementations to define their types using normal objects or generics:

	public class Geometry {
		public static implicit operator Geometr(byte[] data){
			// convert data
		}
	}

OR:

	public class Geometry<T> {
		public static implicit operator Geometry<T>(byte[] data){
			if (T == typeof(double)){
				// convert to double
			}
			else if (T == typeof(decimal)){
				// convert to decimal
			}
		}
	}

This would also work nicely for JSON conversion, as it does for the Pomelo library now.

This would allow for callers to implement their own types, or download libraries with pre-built types that are compatible, like [Pomelo.JsonObject](https://github.com/PomeloFoundation/Pomelo.JsonObject)

Reflection would be tried as a last resort if all other type conversions failed, so it would not be a performance hit on the existing library.

bgrainger:
> If we implement these types, I think we'll have a lot of maintenance on upkeep of the types, standardizing against MySql.Data, compatibility with third party libraries, etc.

I'm not sure I follow. The surface area of `MySqlGeometry` looks pretty small, and [it doesn't change frequently](https://github.com/mysql/mysql-connector-net/commits/6.9/Source/MySql.Data/Types/MySqlGeometry.cs).

By reimplementing `MySqlGeometry` we get compatibility with any third-party libraries that use it.

> What are your thoughts on re-introducing reflection-based conversion for datatypes

[...]

PR #829, opened 2020-06-08
Fix EF Core issue when using MySqlGeometry.Value

lauxjpn:
EF Core throws an exception, when using `MySqlGeometry`, because there is some very general code in `Microsoft.EntityFrameworkCore.Storage.Internal.DbParameterCollectionExtensions.FormatParameterValue(StringBuilder builder, object parameterValue)`, that is unable to handle the `ReadOnlySpan<byte>` type of `MySqlGeometry.Value`.

This PR changes the type to `byte[]` instead. Technically, this is a breaking change for anyone who used the `Value` property before. But because of the implicit cast between `byte[]` and `ReadOnlySpan<byte>`, this should be less of an issue.

In any case, if there is an issue in someones code, this change will lead to a compiler error, that can be easily fixed by replacing `Value` with either `Value.AsSpan()` or `new ReadOnlySpan<byte>(Value)`.

### DIFF
--- a/src/MySqlConnector/MySql.Data.Types/MySqlGeometry.cs
+++ b/src/MySqlConnector/MySql.Data.Types/MySqlGeometry.cs
@@ -38,12 +38,12 @@ namespace MySql.Data.Types
 		/// <summary>
 		/// The Well-known Binary serialization of this geometry.
 		/// </summary>
-		public ReadOnlySpan<byte> WKB => Value.Slice(4);
+		public ReadOnlySpan<byte> WKB => new ReadOnlySpan<byte>(Value).Slice(4);
 
 		/// <summary>
 		/// The internal MySQL form of this geometry.
 		/// </summary>
-		public ReadOnlySpan<byte> Value => m_bytes;
+		public byte[] Value => m_bytes;
 
 		internal MySqlGeometry(byte[] bytes) => m_bytes = bytes;
 


bgrainger:
> Technically, this is a breaking change for anyone who used the Value property before. 

I don't think `MySqlGeometry` gets a lot of use (see this comment: https://github.com/mysql-net/MySqlConnector/issues/70#issuecomment-257742366), so I'm not too worried about a breaking change.

This change does make the `MySqlGeometry` API more compatible with Connector/NET (which does return a `byte[]`) but I'm not happy with returning a mutable `byte[]` that directly exposes the private data; this was why `ReadOnlySpan<byte>` was used in the first place.

And shouldn't you submit a PR to fix EFCore, adding `ReadOnlySpan<byte>` support in https://github.com/dotnet/efcore/blob/252dee5c84eb5d7d1ef2254974f1100f3cb9cc1a/src/EFCore.Relational/Storage/Internal/DbParameterCollectionExtensions.cs#L142-L192? 😀

bgrainger:
> I'm not happy with returning a mutable `byte[]` that directly exposes the private data

Fixed in afa4bf53d4d4a3f896e8b42f606b60f605721611.

bgrainger:
Fixed in [0.69.0](https://github.com/mysql-net/MySqlConnector/releases/tag/0.69.0).

lauxjpn:
> I'm not happy with returning a mutable byte[] that directly exposes the private data; this was why ReadOnlySpan<byte> was used in the first place.

Good point!

> Fixed in 0.69.0.

Thanks, that was quick!

> And shouldn't you submit a PR to fix EFCore, adding ReadOnlySpan<byte> support in https://github.com/dotnet/efcore/blob/252dee5c84eb5d7d1ef2254974f1100f3cb9cc1a/src/EFCore.Relational/Storage/Internal/DbParameterCollectionExtensions.cs#L142-L192?

I am thinking about it. Might be worth it for EF Core 5.

PR #855, opened 2020-07-21
Fix some broken links in docs

yyjdelete:


bgrainger:
Thanks!

(At some point, relative paths stopped working properly, e.g., see https://github.com/mysql-net/MySqlConnector/commit/f41fb8fb32ca56e8d3257cfee45ce4f0b8c896f9, but I didn't look into why that was or if there's a better way to fix it.)

PR #863, opened 2020-07-27
Allows for the retrievel of more than 2^31-1 rows in a single query.

akamor:
Prior to this fix the internal counter for the sequence number was stored as a signed 32-bit integer which was causing overflow issues when we hit Int32.MaxValue.

Signed-off-by: Your Name <adam@tonic.ai>

### DIFF
--- a/src/MySqlConnector/Protocol/Serialization/StandardPayloadHandler.cs
+++ b/src/MySqlConnector/Protocol/Serialization/StandardPayloadHandler.cs
@@ -47,6 +47,6 @@ namespace MySqlConnector.Protocol.Serialization
 		readonly Func<int> m_getNextSequenceNumber;
 		IByteHandler? m_byteHandler;
 		BufferedByteReader? m_bufferedByteReader;
-		int m_sequenceNumber;
+		byte m_sequenceNumber;
 	}
 }


bgrainger:
Thanks for the PR! One minor (legal) thing: can you please amend the commit message so the sign-off has the form `Signed-off-by: Your Real Name <adam@tonic.ai>` with your actual name? I do need to make sure I have the rights to use all contributions to the repository.

You can edit the commit message with `git commit --amend` (or GUI tools provide various ways to do it).

akamor:
Done

image

akamor:
FYI, I'm attempting to spin up a 2B+ row table to confirm this is working as expected.  Would be good to know there aren't other places in code that are gonna have overflow issues.  I should have results in a few hours.

akamor:
@bgrainger Just finished a success test with 2B+ rows (a bit more than 2^31-1 actually) and there were no issues.

bgrainger:
Excellent, thanks!

akamor:
Updated my comment above right after you replied so just making sure you see my update.

bgrainger:
MySQL Server sends one row per (MySQL) packet: https://dev.mysql.com/doc/internals/en/com-query-response.html#packet-ProtocolText::ResultsetRow

(These may be bundled together into one network/TCP packet but they're logically separate. So the counter will overflow after 2.1 billion rows.)

bgrainger:
Fixed in [1.0.1](https://github.com/mysql-net/MySqlConnector/releases/tag/1.0.1).

PR #866, opened 2020-08-06
Implement "IgnoreHasActiveReader" to fix #867

### LINKED ISSUE #867: It is not possible to create two commands in parallel if they are assigned to one connection
````c#
 MySqlConnection connection = new MySqlConnection(/* ... */);
connection.Open(); ;

var cmd1 = connection.CreateCommand();
cmd1.CommandText = "SELECT 1";
using (var reader = cmd1.ExecuteReader())
{
    var cmd2 = connection.CreateCommand();
    cmd2.CommandText = "SELECT 1";  //should not fail

    //cmd2.Execute[..]() <- should fail
}
````

BoBiene:
I need to disable the "HasActiveReader" check because it is validating even if ther is no open reader.

fix #867


bgrainger:
The change in https://github.com/mysql-net/MySqlConnector/commit/67b1f46c75f81b18b615dd0e551d91468a764c7a removed the calls to `HasActiveReader`, so I don't think this PR is necessary anymore.

PR #887, opened 2020-10-01
Cancel query when CommandTimeout expires

### LINKED ISSUE #455: Improve command timeout handling
The query timeout handling in MySqlConnection can be simplified a lot, if it could use server support for query timeouts.  E.g in MariaDB 10.1+ you can using per-statement max_query_time , described in https://mariadb.com/kb/en/library/aborting-statements/ .   There is a max_statement_time session variable (double) for max time, measured in seconds 

MySQL supports this functionality but chose to implement it slightly differently, and in a limited way (variable name is max_execution_time, and it does only work for SELECTs).

Traditional way to implement statement timeouts in other connectors (preceding server support) , was to start a timer task, right before execution of the query. If the task fires, i would create issue "KILL QUERY <connection id>" from another connection. If query finishes before the timeout , the timer task would be canceled. For MySqlConnector,  I feel that even this handling would simplify IO code, which currently relies on socket or stream timeouts( yet still has timer queues for async)

bgrainger:
Fixes #455.

### DIFF
--- a/docs/content/connection-options.md
+++ b/docs/content/connection-options.md
@@ -256,6 +256,12 @@ These are the other options that MySqlConnector supports. They are set to sensib
     <td>Sets the <c>program_name</c> connection attribute passed to MySQL Server. This value may be displayed by diagnostic tools,
     e.g., as the “Program” column in “Client Connections” in <a href="https://www.mysql.com/products/workbench/">MySQL Workbench</a>.</td>
   </tr>
+  <tr id="CancellationTimeout">
+    <td>CancellationTimeout, Cancellation Timeout</td>
+    <td>2</td>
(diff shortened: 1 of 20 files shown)


firephantomassasin:
hi @bgrainger do you have a documentation on the fina behaviour of command timeout? Also, which version should i use to include this fix? Thank you so much!

bgrainger:
The implemented behaviour matches the proposal here: https://github.com/mysql-net/MySqlConnector/issues/455#issuecomment-702424012

You can get it in [1.1.0-beta.1 on NuGet](https://www.nuget.org/packages/MySqlConnector/1.1.0-beta.1); I'm very interested to hear any feedback you have before it's shipped with a non-beta version. (No changes are planned for 1.1.0 release; I just added the beta label so it could get some additional testing.)

PR #892, opened 2020-10-04
Made DisposeAsync fully async. Fixes #876

### LINKED ISSUE #876: Make DisposeAsync fully async in failure case
From https://github.com/mysql-net/MySqlConnector/issues/872#issuecomment-676715354:

The following call stack is possible:

```
MySqlConnection.DisposeAsync
MySqlConnection.CloseAsync
ServerSession.ReturnToPool
ConnectionPool.Return
ServerSession.DisposeAsync(IOBehavior.Synchronous, CancellationToken.None).GetAwaiter().GetResult();
```

The two non-`Async` methods eventually cause sync-over-async in the failure path.

dlemstra:
Fixes #876.

### DIFF
--- a/src/MySqlConnector/Core/ConnectionPool.cs
+++ b/src/MySqlConnector/Core/ConnectionPool.cs
@@ -29,7 +29,7 @@ namespace MySqlConnector.Core
 			if (IsEmpty && unchecked(((uint) Environment.TickCount) - m_lastRecoveryTime) >= 1000u)
 			{
 				Log.Info("Pool{0} is empty; recovering leaked sessions", m_logArguments);
-				RecoverLeakedSessions();
+				await RecoverLeakedSessionsAsync(ioBehavior).ConfigureAwait(false);
 			}
 
 			if (ConnectionSettings.MinimumPoolSize > 0)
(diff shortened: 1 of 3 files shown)


bgrainger:
Thanks for the contribution!

dlemstra:
Thanks for merging this. This is probably not the place to ask but would you be okay with me refectoring this statement to three `if` statements to improve the readability: https://github.com/mysql-net/MySqlConnector/blob/f75ba3ebd10ffe787432e79e47cd57615336e217/src/MySqlConnector/MySqlDataReader.cs#L127.

bgrainger:
Well... I guess I think that refactoring is unnecessary, otherwise I would have written it that way in the first place. 😁

If I were rewriting it right now, I'd _probably_ use C# 9 pattern matching:

```csharp
throw (mySqlException?.ErrorCode, cancellationToken.IsCancellationRequested) switch
{
	(MySqlErrorCode.QueryInterrupted, true) => new OperationCanceledException(mySqlException!.Message, mySqlException, cancellationToken),
	(MySqlErrorCode error, _) => new MySqlException(error, mySqlException!.SqlState, mySqlException.Message, mySqlException),
	_ => new MySqlException("Failed to read the result set.", m_resultSet.ReadResultSetHeaderException),
};
``` 

But that won't actually compile due to some frameworks not supporting `ValueTuple` (even though the generated IL doesn't even use it!). 

No objection if you want to open a PR with some `if` statements.

PR #896, opened 2020-10-15
Support conversion from byte[1] to byte

lauxjpn:
Pomelo translates constant byte values to an explicit byte notation (e.g. `18` will be translated to `0x12`), so that the `~` operator works as expected.

A side effect is however, that constant byte values are returned as `byte[1]` instead of `byte`.

### DIFF
--- a/src/MySqlConnector/Core/Row.cs
+++ b/src/MySqlConnector/Core/Row.cs
@@ -62,6 +62,8 @@ namespace MySqlConnector.Core
 				return (ulong) value != 0;
 			if (value is decimal)
 				return (decimal) value != 0;
+			if (value is byte[] byteArrayValue && byteArrayValue.Length == 1)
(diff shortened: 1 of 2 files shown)


bgrainger:
Can you explain more? I don't see a difference between `SELECT ~0x12;` and `SELECT ~18;`; both are promoted to 64-bit integers.

bgrainger:
Also TIL that `SELECT 18;` and `SELECT 0x12;` return different results.

lauxjpn:
Let me recheck this. I did the change in the process fixing the statement in general, which ended up just translating statements like this to `CAST(~18 as signed)` for a result of `-19` instead of `18446744073709551597`. It might not be needed at all.

lauxjpn:
Thanks for looking into that. I can confirm that this is not needed at all for Pomelo anymore.

PR #905, opened 2020-11-12
Allow cipher suite policy to be configured

iamcarbon:
.NET 5.0's default cipher policy is too restrictive and breaks Amazon's Aurora.

This PR extends the default ciphers when connecting to an Aurora endpoint over Tls1.2.


iamcarbon:
Currently forked and running these changes in production.

src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -1321,11 +1322,35 @@ bool ValidateRemoteCertificate(object rcbSender, X509Certificate? rcbCertificate
 			{
 				if (ioBehavior == IOBehavior.Asynchronous)
 				{
+#if NET5_0
```
bgrainger, line 1325: Use the pattern I described here to opt into old behaviour for old frameworks and have the `#else` have the newest behaviour; that way, the code is future-proof by default: https://github.com/dotnet/sdk/issues/13377#issuecomment-708530179

Specifically, copy a (verbose) line like this one, and swap the order of the if/else: https://github.com/mysql-net/MySqlConnector/blob/902f179a86f254c86ec1b92fa54bcd01500c0287/src/MySqlConnector/MySqlTransaction.cs#L26

iamcarbon: done

iamcarbon:
OK, I think I've just broke my brain keeping track of all these framework targets (and inverted logic). Starting a fresh PR.

iamcarbon:
Superseded by #906

PR #913, opened 2020-12-11
Extract key from between PEM header/footer, instead of just removing the PEM header/footer

### LINKED ISSUE #912: Key from PEM section in SslKey file not extracted correctly in some cases
When reading commonly generated `client-key.pem` files by MySQL, the files can contain an explicit terminating null character (`\0`) at the end of the file (verified on multiple servers).

Because of the way that MySqlConnector extracts the private key from the file, which is by replacing the header and footer with an empty string, the terminating null character stays part of the remaining key.

Finally, when `System.Convert.FromBase64String(key)` is called, the terminating null character is not handled gracefully and an exception is being thrown.

I'll push a PR with a fix.

lauxjpn:
Actually extracts the private/public key from between the PEM header/footer.
This ensures, that anything above or below the PEM section (like an explicit terminating null character at the end of the file) will influence the handling of the key.

Fixes #912

Signed-off-by: Laurents Meyer <laucomm@gmail.com>


bgrainger:
Thanks!

PR #944, opened 2021-02-13
Use hardware-accelerated Adler32 computation

### LINKED ISSUE #865: Use hardware-accelerated Adler32 computation
https://github.com/SixLabors/ImageSharp/blob/master/src/ImageSharp/Formats/Png/Zlib/Adler32.cs is an Apache 2.0-licensed file that implements a hardware-accelerated implementation of Adler32 in C#.

novak-as:
Fixes #865

Signed-off-by: Oleksandr Novak <novak.as@i.ua>


bgrainger:
Thanks!

(I thought I disabled the non-functional SonarCloud scan for PRs, but looks like it's still running; you can ignore that.)

Issue #961, opened 2021-03-22
CTE not working


billkiddo:
Version 1.3.1, trying to execute a query like this: **";WITH cte0 as ( SELECT 1 as `xxx`)  SELECT 2 as `xxx`"**, throws **"Query was empty"** at:  _System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at MySqlConnector.Core.ResultSet.<ReadResultSetHeaderAsync>d__2.MoveNext() in /_/src/MySqlConnector/Core/ResultSet.cs:line 50
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at MySqlConnector.MySqlDataReader.ActivateResultSet(CancellationToken cancellationToken) in /_/src/MySqlConnector/MySqlDataReader.cs:line 137
   at MySqlConnector.MySqlDataReader.<CreateAsync>d__97.MoveNext() in /_/src/MySqlConnector/MySqlDataReader.cs:line 445_

billkiddo:
It works just fine with MySql.Data.

bgrainger:
I get a different error:

```
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ';WITH cte0 as ( SELECT 1 as xxx) SELECT 2 as xxx' at line 1
--
   at MySqlConnector.Core.ResultSet.ReadResultSetHeaderAsync(IOBehavior ioBehavior) in F:\Code\Projects\MySqlConnector\src\MySqlConnector\Core\ResultSet.cs:line 50
   at MySqlConnector.MySqlDataReader.ActivateResultSet(CancellationToken cancellationToken) in F:\Code\Projects\MySqlConnector\src\MySqlConnector\MySqlDataReader.cs:line 135
   at MySqlConnector.MySqlDataReader.CreateAsync(CommandListPosition commandListPosition, ICommandPayloadCreator payloadCreator, IDictionary`2 cachedProcedures, IMySqlCommand command, CommandBehavior behavior, IOBehavior ioBehavior, CancellationToken cancellationToken) in F:\Code\Projects\MySqlConnector\src\MySqlConnector\MySqlDataReader.cs:line 444
[...]

PR #962, opened 2021-03-23
Trim SQL statements

bgrainger:
Related to https://github.com/mysql-net/MySqlConnector/issues/961.

This trims the SQL (from `MySqlCommand.CommandText`) for _non-prepared_ commands before sending it to the server for execution.

This matches the behaviour of `MySqlCommand.Prepare()` better, as well as matching Connector/NET's behaviour in some trivial cases.

It does add some increased risk that unusual SQL will be transformed/truncated by this new code and thus introduce new bugs.

### DIFF
--- a/src/MySqlConnector/Core/StatementPreparer.cs
+++ b/src/MySqlConnector/Core/StatementPreparer.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Data;
 using MySqlConnector.Protocol;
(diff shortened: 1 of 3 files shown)


bgrainger:
> It does add some increased risk that unusual SQL will be transformed/truncated by this new code and thus introduce new bugs.

Indeed, the ADO.NET conformance tests found a problem: executing the command `-- TODO: write SQL` used to work, but now fails.

Issue #979, opened 2021-04-29
Connection pool discrepancies


huraken:
Hello,

We discovered last few weeks some connection pool issues in our production environment and so we decide to dig a bit more with some load tests. For that we compare MySqlConnector and MySql.Data libraries

Giving some context about the load tests we ran :
 
**Libraries used**
MySqlConnector version : 1.3.2
MySql.Data (Oracle) : 0.8.24

**Infrastructure :**
Applications and Databases are hosted in AWS cloud
For databases, we’re using AWS RDS with Aurora MySQL engine 5.6.10, we have a DB cluster that contains 5 Db instances (1 writer and 4 readers)
Applications are containerized with docker (ECS) and we have 6 connections strings defined (3 of them point to the AWS DB cluster reader endpoint and 3 of them point to the AWS DB cluster writer endpoint)
**Load test**
I ran a load test that simulate 1K http requests / second, each request is calling a simple api that is making 5 DB calls to the database. The test will ran during 20 minutes : it will take 10 minutes to reach 1K calls and it will continue with this rate 10 more minutes. Those 5 DB calls are using only one connection string to get the relevant data.
To simplify this test, I only had one instance of the application (so 1 container)

**Connection string**
`Server=***.us-east-1.rds.amazonaws.com;Port=3306;Database=***;Uid=***;Pwd=***;AllowUserVariables=True;ConnectionLifeTime=60;ConnectionTimeout=5;SslMode=None;UseAffectedRows=false;ConnectionIdleTimeout=60;MaximumPoolsize=200;`

**Example of code used for a call to database**

```
public async Task<TestData> GetDataAsync(Guid field1)
        {
            TestData testData = null;

            var query = @"SELECT `field1`, `field2`
                            FROM `table1` 
                            WHERE `field1` = @field1";

            await using (var sqlConnection = new MySqlConnection(connectionString))
            await using (var sqlCommand = new MySqlCommand(query, sqlConnection))
            {

                sqlCommand.Parameters.Add(new MySqlParameter
                {
                    ParameterName = "@field1",
                    MySqlDbType = MySqlDbType.Binary,
                    Value = field1.ToByteArray()
                });

                await sqlConnection.OpenAsync();

                await using (var sqlDataReader = await sqlCommand.ExecuteReaderAsync())
                {
                    if (await sqlDataReader.ReadAsync())
                    {
                        testData = new TestData();

                        if (sqlDataReader.IsDBNull(0))
                        {
                            return null;
[...]

PR #982, opened 2021-04-30
Don't serialize returning connections to pool in background

bgrainger:
Greatly improves latency for #979.

### DIFF
--- a/src/MySqlConnector/Core/BackgroundConnectionResetHelper.cs
+++ b/src/MySqlConnector/Core/BackgroundConnectionResetHelper.cs
@@ -11,12 +11,12 @@ namespace MySqlConnector.Core
 	{
 		public static void AddSession(ServerSession session, MySqlConnection? owningConnection)
 		{
-			var resetTask = session.TryResetConnectionAsync(session.Pool!.ConnectionSettings, IOBehavior.Asynchronous, default);
+			var resetTask = session.TryResetConnectionAsync(session.Pool!.ConnectionSettings, owningConnection, true, IOBehavior.Asynchronous, default);
 			lock (s_lock)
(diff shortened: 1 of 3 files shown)


bgrainger:
A test was run with `wrk -t4 -c100 -d60s -R2000 --latency 'http://localhost:5000/...'`

Legend: `CR` = `ConnectionReset`; `DCR` = `DeferConnectionReset`. Columns (left to right):

* `ConnectionReset=false` should be the fastest, as it eliminates the reset work entirely
* `DeferConnectionReset=true` is the behaviour of 1.2.x; should be slower due to extra work (resetting) when opening a connection
* `DeferConnectionReset=false` is default behaviour in 1.3.x; lowers initial latency but can get very bad
* This PR is the same settings as column 3, but with code to improve latency; it's extremely competitive with column 1 at the cost of increased concurrent connection usage to reset the connections in the background

  | CR = false | CR=true;DCR=true | CR=true;DCR=false | This PR
--- | --- | --- | --- | ---
Max Connections | 100 | 100 | 200 | 182
P50 | 19.07 | 27.60 | 18.82 | 18.77
P75 | 19.63 | 28.06 | 19.25 | 19.17
P90 | 20.43 | 28.59 | 19.74 | 19.58
P99 | 23.57 | 31.63 | 39.26 | 23.92
P99.9 | 35.04 | 38.14 | 164.99 | 36.64
P99.99 | 166.14 | 220.16 | 865.28 | 76.22
P99.999 | 230.01 | 247.55 | 1,150.00 | 235.65
P100 | 230.14 | 256.51 | 1,180.00 | 235.65

bgrainger:
Fixed in [1.3.8](https://github.com/mysql-net/MySqlConnector/releases/tag/1.3.8).

PR #983, opened 2021-05-01
Use `StringBuilder.GetChunks` in `MySqlParameter.AppendBinary`

### LINKED ISSUE #977: Use StringBuilder.GetChunks in MySqlParameter.AppendBinary
If `MySqlParameter.Value` is set to a `StringBuilder` instance (#975) and `MySqlCommand.Prepare()` is called, the code currently calls `StringBuilder.ToString()` and writes it as a length-encoded UTF-8 string.

In .NET 5.0, it should be more performant to use `GetChunks`:

1. ~~loop over the chunks and sum the results of `GetByteCount` (for a `UTF8Encoder`)~~ calculate the total length of the UTF-8 string
2. write the length prefix
3. loop over the chunks and write each out (encoding each one into UTF-8)

novak-as:
Fixes #977

### DIFF
--- a/src/MySqlConnector/Protocol/Serialization/ByteBufferWriter.cs
+++ b/src/MySqlConnector/Protocol/Serialization/ByteBufferWriter.cs
@@ -217,6 +217,7 @@ namespace MySqlConnector.Protocol.Serialization
 
 	internal static class ByteBufferWriterExtensions
 	{
+
 		public static void WriteLengthEncodedInteger(this ByteBufferWriter writer, ulong value)
 		{
 			switch (value)
@@ -258,6 +259,37 @@ namespace MySqlConnector.Protocol.Serialization
 			writer.Write(value);
 		}
 #endif
+
+		public static void WriteLengthEncodedString(this ByteBufferWriter writer, StringBuilder builder)
+		{
+#if !NET45 && !NET461 && !NET471 && !NETSTANDARD1_3 && !NETSTANDARD2_0 && !NETSTANDARD2_1 && !NETCOREAPP2_1
+
+			char? byteCountSurrogatePartBuffer = null;
+			var totalLength = 0;
+
+			foreach (var chunk in builder.GetChunks())
+			{
+				totalLength += SlicedSurrogatesUtility.GetUtf8ByteCount(chunk.Span, ref byteCountSurrogatePartBuffer);
+			}
+
+			writer.WriteLengthEncodedInteger((ulong) totalLength);
+
+			char? preparationSurrogatePartBuffer = null;
+			foreach (var chunk in builder.GetChunks())
+			{
+				SlicedSurrogatesUtility.PrepareSurrogateFreeSpan(chunk,ref preparationSurrogatePartBuffer, out var previousSurrogate, out var nextSlice);
+
+				if (previousSurrogate != null)
+				{
+					writer.Write(previousSurrogate);
+				}
+
+				writer.Write(nextSlice);
+			}
+#else
+			writer.WriteLengthEncodedString(builder.ToString());
+#endif
+		}
 
 		public static void WriteNullTerminatedString(this ByteBufferWriter writer, string value)
 		{
--- a/src/MySqlConnector/Protocol/Serialization/SlicedSurrogatesUtility.cs
+++ b/src/MySqlConnector/Protocol/Serialization/SlicedSurrogatesUtility.cs
@@ -1,0 +1,81 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace MySqlConnector.Protocol.Serialization
+{
+#if !NET45 && !NET461 && !NET471 && !NETSTANDARD1_3 && !NETSTANDARD2_0 && !NETSTANDARD2_1 && !NETCOREAPP2_1
+
+	internal static class SlicedSurrogatesUtility
+	{
+		public static int GetUtf8ByteCount(ReadOnlySpan<char> span, ref char? surrogatePartBuffer)
+		{
+			var containsSurrogateAtBeginning = BitConverter.IsLittleEndian
+				? char.IsLowSurrogate(span[0])
+				: char.IsHighSurrogate(span[0]);
+			var containsSurrogateAtEnding = BitConverter.IsLittleEndian
+				? char.IsHighSurrogate(span[^1])
+				: char.IsLowSurrogate(span[^1]);
+
+			if (containsSurrogateAtBeginning || containsSurrogateAtEnding)
+			{
+				var startPosition = containsSurrogateAtBeginning ? 1 : 0;
+				var length = containsSurrogateAtEnding ? span.Length - 1 : span.Length;
+
+				var operationalSpan = span[startPosition..length];
+
+				var totalCount = Encoding.UTF8.GetByteCount(operationalSpan);
+
+				if (containsSurrogateAtBeginning)
+				{
+					totalCount += Encoding.UTF8.GetByteCount(new[] { surrogatePartBuffer!.Value, span[0] });
+				}
+
+				surrogatePartBuffer = containsSurrogateAtEnding ? span[^1] : null;
+
+				return totalCount;
+			}
+			else
+			{
+				return Encoding.UTF8.GetByteCount(span);
+			}
+		}
+
+		public static void PrepareSurrogateFreeSpan(ReadOnlyMemory<char> mem, ref char? surrogatePartBuffer, out string? previousSurrogatePair, out ReadOnlySpan<char> nextSlice)
+		{
+			var firstChar = mem.Slice(0, 1).Span[0];
+			var lastChar = mem.Slice(mem.Length - 1, 1).Span[0];
+
+			var containsSurrogateAtBeginning = BitConverter.IsLittleEndian
+				? char.IsLowSurrogate(firstChar)
+				: char.IsHighSurrogate(firstChar);
+			var containsSurrogateAtEnding = BitConverter.IsLittleEndian
+				? char.IsHighSurrogate(lastChar)
+				: char.IsLowSurrogate(lastChar);
+
+			previousSurrogatePair = null;
+
+			if (containsSurrogateAtBeginning || containsSurrogateAtEnding)
+			{
+				var startPosition = containsSurrogateAtBeginning ? 1 : 0;
+				var length = containsSurrogateAtEnding ? mem.Length - 1 : mem.Length;
+
+				nextSlice = mem[startPosition..length].Span;
+
+				if (containsSurrogateAtBeginning)
+				{
+					previousSurrogatePair = new string(new[] { surrogatePartBuffer!.Value, firstChar });
+				}
+
+				surrogatePartBuffer = containsSurrogateAtEnding ? lastChar : null;
+			}
+			else
+			{
+				nextSlice = mem.Span;
+			}
+		}
+	}
+#endif
+}
--- a/src/MySqlConnector/MySqlParameter.cs
+++ b/src/MySqlConnector/MySqlParameter.cs
@@ -671,7 +671,7 @@ namespace MySqlConnector
 #endif
 			else if (Value is StringBuilder stringBuilder)
 			{
-				writer.WriteLengthEncodedString(stringBuilder.ToString());
+				writer.WriteLengthEncodedString(stringBuilder);
 			}
 			else if (MySqlDbType == MySqlDbType.Int16)
 			{


src/MySqlConnector/Protocol/Serialization/ByteBufferWriter.cs
```diff
@@ -259,6 +259,27 @@ public static void WriteLengthEncodedString(this ByteBufferWriter writer, ReadOn
...
 #endif
 
+		public static void WriteLengthEncodedString(this ByteBufferWriter writer, StringBuilder builder)
+		{
+#if !NET45 && !NET461 && !NET471 && !NETSTANDARD1_3 && !NETSTANDARD2_0 && !NETSTANDARD2_1 && !NETCOREAPP2_1
+			var totalLength = 0;
+
+			foreach (var chunk in builder.GetChunks())
+			{
+				totalLength += Encoding.UTF8.GetByteCount(chunk.Span);
```
bgrainger, line 269: I suspect this will fail if the just happens to end with a leading surrogate code unit, and the next chunk begins with the matching trailing surrogate.

This would probably need to use an `Encoder` that tracks state, and can account for a the two code units being split across two chunks.

src/MySqlConnector/Protocol/Serialization/ByteBufferWriter.cs
```diff
@@ -259,6 +259,27 @@ public static void WriteLengthEncodedString(this ByteBufferWriter writer, ReadOn
...
+			}
+
+			writer.WriteLengthEncodedInteger((ulong) totalLength);
+
+			foreach (var chunk in builder.GetChunks())
+			{
+				writer.Write(chunk.Span);
```
bgrainger, line 276: There's kind of a latent bug here, in that this method assumes a complete UTF-16 `Span<char>`. Again, this will need to use an `Encoder` that maintains its own internal state across chunk boundaries.

bgrainger (changes requested):
Needs to use an `Encoder`. (It can be initialised with `m_encoder ??= Encoding.UTF8.GetEncoder();`.)

bgrainger:
Here's a test to reproduce the problem: it will split the surrogate code units of a non-BMP character across chunks:

```csharp
var sb = new StringBuilder();
sb.Append("a");
for (int i = 0; i < 50; i++)
	sb.Append("😀");

var totalLength = 0;
foreach (var chunk in sb.GetChunks())
	totalLength += Encoding.UTF8.GetByteCount(chunk.Span);

var bytes = new byte[totalLength];
var output = bytes.AsSpan();

foreach (var chunk in sb.GetChunks())
{
	var bytesWritten = Encoding.UTF8.GetBytes(chunk.Span, output);
	output = output.Slice(bytesWritten);
}

Console.WriteLine(Encoding.UTF8.GetString(bytes));
```

Prints:

```
a😀😀😀😀😀😀😀��😀😀😀😀😀😀😀��😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀��😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀😀
````

The Unicode replacement characters show the bug occurring.

**Comment by bgrainger** [comment]
Unfortunately, some experimentation shows that `GetByteCount` can only return an accurate count for the second chunk if `GetBytes` has already been called to process the first chunk, not `GetByteCount`. That is, only `GetBytes` updates the encoder's internal state. Thus, you can't call `GetByteCount` in a loop to determine the total length, and the suggested approach in https://github.com/mysql-net/MySqlConnector/issues/977 isn't viable.

**Comment by novak-as** [comment]
Thanks for pointing it out, to be honest - I totally missed the case with surrogate characters being split (and thus, corrupted).
I was able to pass a test with utf16 smiles (by handling surrogates explicitly), but it will require a few more iterations to polish/profile everything up. As for now - the current solution is messy, so I'm converting this PR back to draft.

**Comment by bgrainger** [comment]
Thanks for the work on this; given how long it took for someone to ask for `StringBuilder` support, and given how uncommon calling `MySqlCommand.Prepare` is, I do wonder if it's worth it?

I did think of a possibly easier approach: reserve _n_ bytes for the length-encoded integer up front. It could always reserve nine bytes, for `0xFE` then 8 integer bytes, or it could calculate the worst-case scenario (3 UTF-8 bytes per UTF-16 code unit to encode `U+FEFF` for unpaired surrogates) and write `0xFD` + 4 bytes or `0xFC` + 2 bytes based on `stringBuilder.Length * 3`. (Saving 4-6 bytes may not be worth this extra complexity.)

Then write all all the UTF-8 encoded bytes (one chunk at a time), calculate the actual length (in bytes), and go back and write it in.

**Comment by novak-as** [comment]
Actually, I believe it's up to you to decide if it worth introducing additional complexity here, but here some measurable results I was able to get:

```
// * Summary *

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19041.928 (2004/?/20H1)
Intel Core i7-8850H CPU 2.60GHz (Coffee Lake), 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=5.0.202
  [Host]     : .NET Core 5.0.5 (CoreCLR 5.0.521.16609, CoreFX 5.0.521.16609), X64 RyuJIT
  DefaultJob : .NET Core 5.0.5 (CoreCLR 5.0.521.16609, CoreFX 5.0.521.16609), X64 RyuJIT


|                      Method |        Mean |     Error |      StdDev |   Gen 0 |  Gen 1 | Gen 2 | Allocated |
|---------------------------- |------------:|----------:|------------:|--------:|-------:|------:|----------:|
|            OldBehaviorShort |    159.3 ns |   0.98 ns |     0.77 ns |  0.1273 |      - |     - |     600 B |
|            NewBehaviorShort |    167.8 ns |   1.30 ns |     1.16 ns |  0.0799 |      - |     - |     376 B |
|  OldBehaviorShortSurrogates |    878.7 ns |   4.08 ns |     3.61 ns |  0.3395 | 0.0010 |     - |    1600 B |
|  NewBehaviorShortSurrogates |  1,275.1 ns |   8.34 ns |     7.81 ns |  0.3071 |      - |     - |    1448 B |
|           OldBehaviorMedium |    548.7 ns |   4.38 ns |     3.88 ns |  0.4873 | 0.0019 |     - |    2296 B |
|           NewBehaviorMedium |    509.9 ns |   3.03 ns |     2.53 ns |  0.2699 |      - |     - |    1272 B |
| OldBehaviorMediumSurrogates |  2,918.1 ns |  18.04 ns |    16.00 ns |  1.3428 | 0.0153 |     - |    6336 B |
| NewBehaviorMediumSurrogates |  3,339.6 ns |  18.91 ns |    16.76 ns |  1.0109 | 0.0038 |     - |    4776 B |
|             OldBehaviorLong |    843.2 ns |   4.67 ns |     4.14 ns |  0.9155 | 0.0086 |     - |    4312 B |
|             NewBehaviorLong |    741.6 ns |   6.04 ns |     5.05 ns |  0.4854 | 0.0019 |     - |    2288 B |
|   OldBehaviorLongSurrogates |  5,280.4 ns |  45.15 ns |    40.02 ns |  2.6245 | 0.0381 |     - |   12384 B |
|   NewBehaviorLongSurrogates |  5,640.4 ns |  41.88 ns |    34.97 ns |  1.8921 | 0.0229 |     - |    8920 B |
|           OldBehaviorExLong |  4,725.1 ns |  94.10 ns |   108.36 ns |  9.8953 | 1.0986 |     - |   46672 B |
|           NewBehaviorExLong |  3,534.0 ns |  70.18 ns |   122.92 ns |  5.6496 | 0.3319 |     - |   26648 B |
| OldBehaviorExLongSurrogates | 47,340.7 ns | 356.22 ns |   297.46 ns | 30.7617 | 7.6294 |     - |  145824 B |
| NewBehaviorExLongSurrogates | 45,830.9 ns | 897.19 ns | 1,134.66 ns | 22.5220 | 3.7231 |     - |  107160 B |
```
`*Short*` = 100 symbols
`*Medium*` = 500 symbols
`*Long*` = 1000 symbols
`*ExLong*` = 10000 symbols

`*Surrogates` = worst-case scenario, symbol = surrogate pair (i.e having surrogate pair split at the end of every chunk).

I don't think that it's common to have that's many surrogate characters in the SQL statement, so it's probably worth focusing on these results:
```
|                      Method |        Mean |     Error |      StdDev |   Gen 0 |  Gen 1 | Gen 2 | Allocated |
|---------------------------- |------------:|----------:|------------:|--------:|-------:|------:|----------:|
|            OldBehaviorShort |    159.3 ns |   0.98 ns |     0.77 ns |  0.1273 |      - |     - |     600 B |
|            NewBehaviorShort |    167.8 ns |   1.30 ns |     1.16 ns |  0.0799 |      - |     - |     376 B |
|           OldBehaviorMedium |    548.7 ns |   4.38 ns |     3.88 ns |  0.4873 | 0.0019 |     - |    2296 B |
|           NewBehaviorMedium |    509.9 ns |   3.03 ns |     2.53 ns |  0.2699 |      - |     - |    1272 B |
|             OldBehaviorLong |    843.2 ns |   4.67 ns |     4.14 ns |  0.9155 | 0.0086 |     - |    4312 B |
|             NewBehaviorLong |    741.6 ns |   6.04 ns |     5.05 ns |  0.4854 | 0.0019 |     - |    2288 B |
|           OldBehaviorExLong |  4,725.1 ns |  94.10 ns |   108.36 ns |  9.8953 | 1.0986 |     - |   46672 B |
|           NewBehaviorExLong |  3,534.0 ns |  70.18 ns |   122.92 ns |  5.6496 | 0.3319 |     - |   26648 B |
```

From this point of view - `.Chunks` usage looks promising even with a straightforward implementation.

**Comment by bgrainger** [comment]
Are those benchmarks with the `SlicedEncoder` in this PR?

**Comment by novak-as** [comment]
Well, to be 100% precise - I used a bit modified version of `SlicedEncoder` from the PR. [Here](https://github.com/novak-as/mysql-chunks-playground) is a standalone version with everything required to reproduce these results. You just need to run benchmark against `perf.csproj`

**Comment by novak-as** [comment]
Updated benchmark (most performance improvements came from fixing the benchmark itself). Thresholds are the same: 100, 500, 1000, and 10000 symbols (just to check the trend)

```
// * Summary *

BenchmarkDotNet=v0.12.1.1538-nightly, OS=Windows 10.0.19041.928 (2004/May2020Update/20H1)
Intel Core i7-8850H CPU 2.60GHz (Coffee Lake), 1 CPU, 12 logical and 6 physical cores
.NET SDK=5.0.203
  [Host]     : .NET 5.0.6 (5.0.621.22011), X64 RyuJIT
  DefaultJob : .NET 5.0.6 (5.0.621.22011), X64 RyuJIT


|                      Method |        Mean |     Error |    StdDev |   Gen 0 |  Gen 1 | Gen 2 | Allocated |
|---------------------------- |------------:|----------:|----------:|--------:|-------:|------:|----------:|
|            OldBehaviorShort |    136.9 ns |   1.07 ns |   0.89 ns |  0.1001 |      - |     - |     472 B |
|            NewBehaviorShort |    145.2 ns |   2.78 ns |   2.73 ns |  0.0527 |      - |     - |     248 B |
|           OldBehaviorMedium |    500.4 ns |   9.24 ns |   8.19 ns |  0.3757 | 0.0010 |     - |   1,768 B |
|           NewBehaviorMedium |    477.4 ns |   9.14 ns |   8.98 ns |  0.1578 |      - |     - |     744 B |
|             OldBehaviorLong |    770.6 ns |   7.44 ns |   6.60 ns |  0.6962 | 0.0038 |     - |   3,280 B |
|             NewBehaviorLong |    674.7 ns |  10.06 ns |   9.41 ns |  0.2661 | 0.0010 |     - |   1,256 B |
|           OldBehaviorExLong |  3,882.0 ns |  76.25 ns |  63.67 ns |  7.7515 | 0.8545 |     - |  36,640 B |
|           NewBehaviorExLong |  2,588.2 ns |  10.47 ns |   8.75 ns |  3.5210 |      - |     - |  16,616 B |
```

File: src/MySqlConnector/Protocol/Serialization/ByteBufferWriter.cs (hunk 2 above)
Comment (novak-as) [reply src/MySqlConnector/Protocol/Serialization/ByteBufferWriter.cs]: this was fixed in the further commits, thus I'm marking this as a resolved

File: src/MySqlConnector/Protocol/Serialization/ByteBufferWriter.cs (hunk 1 above)
Comment (novak-as) [reply src/MySqlConnector/Protocol/Serialization/ByteBufferWriter.cs]: this was fixed in the further commits, thus I'm marking this as a resolved

File: src/MySqlConnector/Protocol/Serialization/SlicedSurrogatesUtility.cs [hunk 3]
```diff
@@ -0,0 +1,81 @@
...
+	internal static class SlicedSurrogatesUtility
+	{
+		public static int GetUtf8ByteCount(ReadOnlySpan<char> span, ref char? surrogatePartBuffer)
+		{
+			var containsSurrogateAtBeginning = BitConverter.IsLittleEndian
+				? char.IsLowSurrogate(span[0])
+				: char.IsHighSurrogate(span[0]);
+			var containsSurrogateAtEnding = BitConverter.IsLittleEndian
+				? char.IsHighSurrogate(span[^1])
+				: char.IsLowSurrogate(span[^1]);
```
bgrainger, line 20:
```suggestion
			var containsSurrogateAtBeginning = char.IsLowSurrogate(span[0]);
			var containsSurrogateAtEnding = char.IsHighSurrogate(span[^1]);
```

`IsLittleEndian` shouldn't make any difference here: the high-surrogate code unit is always stored first, followed by the low-surrogate code unit.

(If these were serialised to bytes, the byte order of the UTF-16 code units themselves could be in little-endian or big-endian byte order, but that's irrelevant when dealing with `Span<char>`.)

This should just be collapsed down to the case where `IsLittleEndian == true`.

src/MySqlConnector/Protocol/Serialization/SlicedSurrogatesUtility.cs
```diff
@@ -0,0 +1,81 @@
...
+
+			if (containsSurrogateAtBeginning || containsSurrogateAtEnding)
+			{
+				var startPosition = containsSurrogateAtBeginning ? 1 : 0;
+				var length = containsSurrogateAtEnding ? span.Length - 1 : span.Length;
+
+				var operationalSpan = span[startPosition..length];
```
bgrainger, line 27: `length` is a misnomer; it's actually (one past) the end index.

src/MySqlConnector/Protocol/Serialization/SlicedSurrogatesUtility.cs
```diff
@@ -0,0 +1,81 @@
...
+
+				var totalCount = Encoding.UTF8.GetByteCount(operationalSpan);
+
+				if (containsSurrogateAtBeginning)
+				{
+					totalCount += Encoding.UTF8.GetByteCount(new[] { surrogatePartBuffer!.Value, span[0] });
```
bgrainger, line 33: Seems like a potential bug if the user passes in malformed data starting with an unpaired low-surrogate code unit? (It should write out `U+FFFD`, not crash.)

src/MySqlConnector/Protocol/Serialization/SlicedSurrogatesUtility.cs
```diff
@@ -0,0 +1,81 @@
...
+		{
+			var firstChar = mem.Slice(0, 1).Span[0];
+			var lastChar = mem.Slice(mem.Length - 1, 1).Span[0];
+
+			var containsSurrogateAtBeginning = BitConverter.IsLittleEndian
+				? char.IsLowSurrogate(firstChar)
+				: char.IsHighSurrogate(firstChar);
+			var containsSurrogateAtEnding = BitConverter.IsLittleEndian
+				? char.IsHighSurrogate(lastChar)
+				: char.IsLowSurrogate(lastChar);
```
bgrainger, line 56: See previous comments.

bgrainger:
> I did think of a possibly easier approach: reserve n bytes for the length-encoded integer up front.

Not only is the code simpler, it appears to be faster (as it only needs to make one pass over the buffer). (I haven't actually tested it for correctness, though 😬)

Code here: https://github.com/bgrainger/mysql-chunks-playground/commit/0095d518ba02b19f07b3275850a4f82efcff7266

Benchmark results (as "Simple") below:

``` ini

BenchmarkDotNet=v0.13.0, OS=Windows 10.0.19043.964 (21H1/May2021Update)
Intel Core i7-10875H CPU 2.30GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK=6.0.100-preview.3.21202.5
  [Host]     : .NET 5.0.6 (5.0.621.22011), X64 RyuJIT
  DefaultJob : .NET 5.0.6 (5.0.621.22011), X64 RyuJIT


```
|                         Method |        Mean |     Error |    StdDev |   Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------------------------- |------------:|----------:|----------:|--------:|-------:|------:|----------:|
|               OldBehaviorShort |    145.2 ns |   2.10 ns |   1.86 ns |  0.0715 |      - |     - |     600 B |
|               NewBehaviorShort |    156.6 ns |   2.16 ns |   1.91 ns |  0.0448 |      - |     - |     376 B |
|            SimpleBehaviorShort |    115.2 ns |   1.52 ns |   1.42 ns |  0.0459 |      - |     - |     384 B |
|      OldBehaviorShortSurrogate |    814.9 ns |  15.93 ns |  18.96 ns |  0.1907 |      - |     - |   1,600 B |
|     NewBehaviorShortSurrogates |  1,188.8 ns |  20.61 ns |  19.28 ns |  0.1717 |      - |     - |   1,448 B |
|  SimpleBehaviorShortSurrogates |  1,282.9 ns |  25.14 ns |  25.82 ns |  0.1411 |      - |     - |   1,192 B |
|              OldBehaviorMedium |    542.6 ns |  10.39 ns |  11.97 ns |  0.2737 | 0.0010 |     - |   2,296 B |
|              NewBehaviorMedium |    521.4 ns |  10.30 ns |  14.44 ns |  0.1516 |      - |     - |   1,272 B |
|           SimpleBehaviorMedium |    416.9 ns |   7.48 ns |   6.63 ns |  0.1526 |      - |     - |   1,280 B |
|    OldBehaviorMediumSurrogates |  2,527.8 ns |  47.18 ns |  44.13 ns |  0.7553 | 0.0038 |     - |   6,336 B |
|    NewBehaviorMediumSurrogates |  2,883.2 ns |  51.49 ns |  48.16 ns |  0.5684 | 0.0038 |     - |   4,776 B |
| SimpleBehaviorMediumSurrogates |  2,936.0 ns |  58.35 ns |  54.58 ns |  0.5150 |      - |     - |   4,328 B |
|                OldBehaviorLong |    805.1 ns |  13.46 ns |  12.59 ns |  0.5150 | 0.0057 |     - |   4,312 B |
|                NewBehaviorLong |    709.6 ns |  11.08 ns |  10.36 ns |  0.2728 | 0.0010 |     - |   2,288 B |
|             SimpleBehaviorLong |    631.1 ns |  12.64 ns |  11.82 ns |  0.2737 | 0.0010 |     - |   2,296 B |
|      OldBehaviorLongSurrogates |  4,526.1 ns |  81.15 ns |  71.94 ns |  1.4801 | 0.0305 |     - |  12,384 B |
|      NewBehaviorLongSurrogates |  4,840.3 ns |  49.08 ns |  45.91 ns |  1.0605 |      - |     - |   8,920 B |
|   SimpleBehaviorLongSurrogates |  4,800.6 ns |  81.70 ns |  76.42 ns |  0.9995 | 0.0153 |     - |   8,376 B |
|              OldBehaviorExLong |  4,295.1 ns |  33.91 ns |  31.72 ns |  5.5695 | 0.3433 |     - |  46,672 B |
|              NewBehaviorExLong |  3,349.1 ns |  65.14 ns |  84.69 ns |  3.1776 | 0.0992 |     - |  26,648 B |
|           SimpleBehaviorExLong |  2,785.6 ns |  26.71 ns |  24.98 ns |  3.1776 | 0.0992 |     - |  26,656 B |
|    OldBehaviorExLongSurrogates | 39,012.3 ns | 600.64 ns | 561.84 ns | 17.3340 | 2.8687 |     - | 145,824 B |
|    NewBehaviorExLongSurrogates | 39,902.5 ns | 496.72 ns | 440.33 ns | 12.6953 | 1.5869 |     - | 107,160 B |
| SimpleBehaviorExLongSurrogates | 36,473.7 ns | 531.51 ns | 471.17 ns | 12.5732 | 1.5259 |     - | 105,976 B |

bgrainger:
I added a version of my fix here: https://github.com/mysql-net/MySqlConnector/commit/971aa75018f0bb0ac073df4091419b9668c689f5

It was more complicated than the prototype, as I had make sure the last chunk was flushed properly. I also found some latent bugs in the text protocol implementation.

I added some new tests (of emojis and unpaired surrogates), and now the `netcoreapp2.1` and `netcoreapp3.1` tests are failing (but only in CI integration tests, not locally). This is surprising as that implementation didn't change. It's possible that the current shipping code in 1.3.8 would also fail with the new tests. This will require further investigation.

bgrainger:
Fixed version here: https://github.com/mysql-net/MySqlConnector/commit/ddc733bebcc44cd84fb8cfe8f3d79eb22d5bae02. I'll ship that and close this.

PR #998, opened 2021-06-07
Fix link to MySQL docs for BOOL.

ddunkin:


bgrainger:
It's not even October! 😂

Thanks.

PR #1037, opened 2021-09-18
Report telemetry with ActivitySource 

### LINKED ISSUE #1036: Add ActivitySource
## Proposal

ADO.NET libraries should be instrumented with the [`ActivitySource`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.activitysource) API. 

This document is based on the [Instrumenting a library/application with .NET Activity API](https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Api/README.md#instrumenting-a-libraryapplication-with-net-activity-api) recommendations from OpenTelemetry.

## Basics

Install v5.0.1 or later of the [System.Diagnostics.DiagnosticSource](https://www.nuget.org/packages/System.Diagnostics.DiagnosticSource) NuGet package.

Create a single `ActivitySource` named after the assembly name of the library being instrumented:

```csharp
ActivitySource activitySource = new(typeof(ExampleDbConnection).Assembly.GetName();
```

## Activities

The `ActivityKind` MUST be set to `ActivityKind.Client`.

The standard activity names are:

* `Open` - establishing a connection to a database server
  * The actual implementation may simply be retrieving a connection from a pool; this is still recorded as an `Activity`.
* `Execute` - executing a command
  * OpenTelemetry recommends that this be named `<db.operation> <db.name>.<db.sql.table>`, `<db.operation> <db.name>`, or `<db.name>`. Since those values will already be set as custom tags on the `Activity`, an `ActivityListener` that is reporting `Activity` objects to an OpenTelemetry endpoint should override the span name (with the values of those tags) if desired.
  * Is it not necessary to record `ExecuteScalar`, `ExecuteNonQuery`, and `ExecuteReader` as different activities.
  * The `Execute` activity ends when the `DbDataReader` returned by `ExecuteReader` is closed/disposed. `AddEvent` will be used to record when the first response is returned by the server. For MySqlConnector, the event name is `read-result-set-header`, and can occur multiple times during an activity if multiple result sets are read. For Npgsql, the event name is `received-first-response`.
* `Commit` - committing a transaction
* `Rollback` - rolling back a transaction
  * _NOTE:_ These are currently represented as their own activities, but could alternatively be events added to a "Transaction" Activity that encompasses the entire transaction. This may change in the future based on developments in the OpenTelemetry database conventions.

Future activity names may include the following. These should not be used until formalized by this specification.

* `Close` - closing a database connection
* `Prepare` - preparing a statement

## Tags

These tags are taken from the [OpenTelemetry Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#connection-level-attributes); that document is the authoritative source and should take precedence in case of a conflict with the recommendations here.

Tag | Type | Description | Examples | Required
--- | --- | --- | --- | ---
`db.system` | string | The DBMS. See OTel spec for allowable values. | `mysql`, `postgresql`, `other_sql` | yes
`db.connection_string` | string | The connection string. Sensitive information should be removed. (In many ADO.NET providers, this is the default, but users can override it with `Persist Security Info = true`.) | `Server=dbserver; Database=mydb; User ID=myuser` |  no
`db.connection_id` | string | The "connection ID" or "server thread" etc. that uniquely identifies this connection on the DB server. (This is an extension to the standard OpenTelemetry attributes.) | `173` | no
`db.user` | string | User name for accessing the database. | `myuser` | no
`net.peer.name` | string | Remote hostname or similar. Usually the same as `DbConnection.DataSource`. | `dbserver` | [1]
`net.peer.ip` | string | Remote IP address of the peer. | `127.0.0.1` | [1]
`net.peer.port` | int | Remote port number. | `1433`, `3306` | [2]
`net.transport` | string | Transport protocol used. | `ip_tcp`, `inproc` | yes
`db.name` | string | The name of the database being accessed. For commands that switch the database, this should be set to the target database. Usually the same as `DbConnection.Database`. | `customers`; `main` | yes
`db.statement` | string | The database statement being executed, i.e., `DbCommand.CommandText`. [3] | `SELECT * FROM example_table;`  | yes [4]
`db.operation` | string | The name of the operation being executed, e.g. the SQL keyword. [5] | `SELECT`, `INSERT` | no
`db.sql.table` | string | The name of the primary table that the operation is acting upon, including the schema name (if applicable). [6] | `public.users`; `customers` | recommended
`thread.id` | int | Current managed thread ID, i.e., `Thread.CurrentThread.ManagedThreadId`. | `42` | no
`thread.name` | string | Current thread name, i.e., `Thread.CurrentThread.Name`. | `Worker Thread` | no
`exception.type` | string | The type of the exception (its fully-qualified class name, if applicable). | `ExampleLib.ExampleDbException` | [7]
`exception.message` | string | `The exception message.` | `Could not connect to remote server` | [7]
`exception.stacktrace` | string | A managed stack trace |  | no
`otel.status_code` | string | Represents pass/fail result of the work. | `UNSET`, `OK`, `ERROR` | yes
`otel.status_description` | string | Optional Description that provides a descriptive message of the Status. | `Could not connect to remote server` | [8]

1. One of `net.peer.name` or `net.peer.ip` should be specified.
2. Required if using a non-default port.
3. The value may be sanitized to exclude sensitive information. In particular, parameter placeholders SHOULD NOT be replaced with parameter values.
4. Users may want to opt in to sending this data (due to confidentiality and size).
5. When setting this to an SQL keyword, it is not recommended to attempt any client-side parsing of `db.statement` just to get this property, but it should be set if the operation name is provided by the library being instrumented. If the SQL statement has an ambiguous operation, or performs more than one operation, this value may be omitted.
6. It is not recommended to attempt any client-side parsing of `db.statement` just to get this property, but it should be set if it is provided by the library being instrumented. If the operation is acting upon an anonymous table, or more than one table, this value MUST NOT be set.
7. If an error occurs, at least one of `exception.type` and `exception.message` is required.
8. MUST only be set if `otel.status_code` is `ERROR`.

## Exceptions

If an exception occurs, the status (`otel.status_code` tag) should be set to `ERROR` and the exception details should be added as an event named `exception`, as per https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#record-exception.

## Example

```csharp
public class ExampleDbConnection : DbConnection
{
	static ActivitySource _activitySource = new(typeof(ExampleDbConnection).Assembly.GetName().Name,
		typeof(ExampleDbConnection).Assembly.GetName().Version.ToString());

	public override void Open()
	{
		using var activity = _activitySource.StartActivity("Open", ActivityKind.Client);
		try
		{
			if (activity is { IsAllDataRequested: true })
			{
				activity.SetTag("db.system", "other_sql");
				activity.SetTag("db.connection_string", this.ConnectionString);
				activity.SetTag("db.user", userId);
				activity.SetTag("db.name", this.Database);
				activity.SetTag("net.peer.name", this.DataSource);
				activity.SetTag("net.transport", "ip_tcp");
				activity.SetTag("thread.id", Thread.CurrentThread.ManagedThreadId);
			}
			
			// ... actual Open implementation elided ...
			
			activity?.SetTag("otel.status_code", "OK");
		}
		catch (Exception ex) when (activity is { IsAllDataRequested: true })
		{
			activity.SetTag("otel.status_code", "ERROR");
			activity.SetTag("otel.status_description", ex.Message);
			activity.AddEvent(new ActivityEvent("exception", tags: new ActivityTagsCollection
			{
				{ "exception.type", exception.GetType().FullName },
				{ "exception.message", exception.Message },
				{ "exception.stacktrace", exception.ToString() },
			}));
			throw;
		}
	}
}
```

bgrainger:
Fixes #1036.

Sample telemetry produced by connecting this PR to OpenTelemetry.Exporter.Jaeger.

image


src/MySqlConnector/MySqlDataReader.cs
```diff
@@ -604,6 +606,9 @@ internal async Task DisposeAsync(IOBehavior ioBehavior, CancellationToken cancel
 			Command.CancellableCommand.SetTimeout(Constants.InfiniteTimeout);
 			connection.FinishQuerying(m_hasWarnings);
 
+			m_activity.SetSuccess();
```
vonzshik, line 609: Here's another interesting thing: [specification for database providers](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md) does not mention whether statement is completed whenever the first row is received, or every single row is read. SqlClient does the former.

vonzshik (commented):
@roji for some thoughts.

src/MySqlConnector/Utilities/ActivitySourceHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+	public static readonly IEnumerable<KeyValuePair<string, object?>> DefaultTags = new KeyValuePair<string, object?>[]
+	{
+			new(DatabaseSystemTagName, DatabaseSystemValue),
+	};
+
+	public static void SetSuccess(this Activity? activity) => activity?.SetTag(StatusCodeTagName, "OK");
+
+	public static void SetException(this Activity activity, Exception exception)
+	{
+		activity.SetTag(StatusCodeTagName, "ERROR");
```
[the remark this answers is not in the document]
bgrainger: Those aren't contradictory: the status should be set to `ERROR` and then the exception (with the `exception.type`, `exception.message` etc. tags) should be recorded as an event (not set as extra tags on the Activity, as this implementation does): https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#record-exception

src/MySqlConnector/MySqlDataReader.cs
bgrainger: Yes, I noted that here: https://github.com/mysql-net/MySqlConnector/issues/1036:

> TODO: Does the `Execute` activity end when `ExecuteReader` returns a `DbDataReader`, or when that `DbDataReader` has finished consuming the result set (and is closed/disposed)?

My preference is the latter; as an application developer, I usually care about when the entire `using (var reader = cmd.ExecuteReader()) { ... }` block has finished, not about when it "begins".

I suggested adding an event for when the `DbDataReader` is created and returned, for those who care about breaking the whole span into two parts.

It would be ideal to standardise this, have a good name for the event, etc.

src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -521,6 +522,16 @@ public async Task DisposeAsync(IOBehavior ioBehavior, CancellationToken cancella
...
 			m_payloadHandler.ByteHandler.RemainingTimeout = Constants.InfiniteTimeout;
+
+			ActivityTags = new KeyValuePair<string, object?>[]
+			{
+				new(ActivitySourceHelper.DatabaseSystemTagName, ActivitySourceHelper.DatabaseSystemValue),
+				new(ActivitySourceHelper.DatabaseConnectionStringTagName, cs.ConnectionStringBuilder.GetConnectionString(cs.ConnectionStringBuilder.PersistSecurityInfo)),
+				new(ActivitySourceHelper.DatabaseNameTagName, cs.Database),
+				new(ActivitySourceHelper.DatabaseUserTagName, cs.UserID),
+				new(ActivitySourceHelper.NetPeerNameTagName, cs.HostNames?.Count > 0 ? cs.HostNames[0] : ""), // TODO
+				new(ActivitySourceHelper.NetTransportTagName, "ip_tcp"), // TODO
```
bgrainger, line 533: Needs a proper implementation.

bgrainger, line 533: Need to add port (if it's non-standard).

src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -65,6 +65,7 @@ public ServerSession(ConnectionPool? pool, int poolGeneration, int id)
 	public bool SupportsDeprecateEof => m_supportsDeprecateEof;
 	public bool SupportsSessionTrack => m_supportsSessionTrack;
 	public bool ProcAccessDenied { get; set; }
+	public IEnumerable<KeyValuePair<string, object?>>? ActivityTags { get; private set; }
```
bgrainger, line 68: Use `ActivityTagsCollection`.

src/MySqlConnector/Utilities/ActivitySourceHelper.cs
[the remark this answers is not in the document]
bgrainger: That is, this implementation should be switched to how you've done it for npgsql (but also keep the status): https://github.com/npgsql/npgsql/pull/3967/files#diff-472f4042d307b9d23ac1aab6332ef284d7bc1605c2031c8015b87a657c6d69f7R48

src/MySqlConnector/Utilities/ActivitySourceHelper.cs
```diff
@@ -0,0 +1,47 @@
...
+		activity.SetTag("otel.status_description", exception is MySqlException mySqlException ? mySqlException.ErrorCode.ToString() : exception.Message);
+		activity.SetTag("exception.type", exception.GetType().FullName);
+		activity.SetTag("exception.message", exception.Message);
+		activity.SetTag("exception.stacktrace", exception.StackTrace);
```
bgrainger, line 36:
```suggestion
		activity.SetTag("exception.stacktrace", exception.ToString());
```

src/MySqlConnector/Core/ServerSession.cs
bgrainger: Actually; don't use that as the public type, because it's mutable.

src/MySqlConnector/MySqlDataReader.cs
roji: See ongoing conversation in https://github.com/npgsql/npgsql/pull/3967#discussion_r705255298

PR #1046, opened 2021-10-01
Remove `ConnectionIdlePingTime` behaviour

### LINKED ISSUE #1042: Don't ping when Connection Reset = false
When `Connection Reset = true`, the connection is reset (with a `COM_RESET` packet) when it's retrieved from the pool. This has the side-effect of checking connection liveness.

When `Connection Reset = false`, no reset occurs, but a `COM_PING` packet is still sent to check liveness (thus incurring a server roundtrip). Clients who wish to opt out of this can set `Connection Idle Ping Time = <large value>` but this is documented as an experimental setting.

There's no ADO.NET requirement that the server be pinged; it was just done to improve the odds that `MySqlConnection.Open` always returns a valid connection (instead of one that is in the pool but has been closed by the server). We could allow `Connection Reset = false` to be a "power user" / performance setting that also disables any server roundtrip (both reset and ping) and simply returns a connection from the pool if one is present. (This supersedes the more complicated options proposed in https://github.com/mysql-net/MySqlConnector/issues/967.)

The documentation would be updated to call out the effects of this change, and to note that `MySqlConnection.Open` might actually return an invalid connection that throws as soon as it's used. (Technically, this is always possible, because a network partition can happen at any time. Thus, code that's already written to retry wouldn't be affected.) Users could always call `MySqlConnection.Ping(Async)` if they still want the old behaviour; there's no reason it needs to be built into the library.

`MySqlConnectionStringBuilder.ConnectionIdlePingTime` would be marked `[Obsolete]`, stop having any effect, and eventually be removed.

bgrainger:
Fixes #1042.

* `Connection Reset = True` will perform a server roundtrip to reset the connection.
* `Connection Reset = False` will return a connection from the pool (if one exists); the client should use appropriate retry logic to reopen the connection if the first communication with the server fails.


src/MySqlConnector/Core/ConnectionPool.cs
```diff
@@ -67,17 +67,9 @@ public async ValueTask<ServerSession> GetSessionAsync(MySqlConnection connection
 				else
 				{
 					if (ConnectionSettings.ConnectionReset || session.DatabaseOverride is not null)
-					{
```
danielgindi, line 70: The release notes are confusing. It's not clear that an unstable pooled connection may only be returned if `ConnectionReset = false`, can worry anyone who did not dive into the code to look at this set of IFs. Please update the release notes :-)

bgrainger: > If Connection Reset = False, MySqlConnection.Open will return a pooled connection (if one is available) without testing it for liveness (by pinging the server). 

Do you want this to say "If and only if"? Do you have a suggested edit?

danielgindi: Just make it clearer for which "if" this paragraph belongs too :)

When ConnectionReset=false, if the server has closed the pooled connection, MySqlConnection.Open will succeed, but the first operation on that connection will fail. Command execution should be wrapped in a try/catch/retry loop to handle this scenario.

bgrainger: Ah, gotcha. Yes, I see now that that is confusing. (Should have kept reading 😀)

bgrainger: Consolidated the bullet points here: https://github.com/mysql-net/MySqlConnector/commit/5151ca4df84103fc954c25a699bb8881d5a5ab99

Issue #1070, opened 2021-10-26
Implement MySqlDecimal


sumitdvlp:
Hi,

The GetDecimal() or GetValue() functions of MySqlDataReader [class](https://github.com/mysql-net/MySqlConnector/blob/master/src/MySqlConnector/MySqlDataReader.cs) are unable to handle very large value like
'1234567890123456789012345678901234567890123456789012345678901234.5'
They throw exception like value is too large because it return C# decimal datatype.

We saw MySQL connecter net has this functionality [here](https://github.com/mysql/mysql-connector-net/blob/8.0/MySQL.Data/src/datareader.cs). which return MySqlDecimal data type.

Can you please add method like GetMySqlDecimal which can support very large Decimal values.

sumitdvlp:
Created a PR https://github.com/mysql-net/MySqlConnector/pull/1075

bgrainger:
Added in [2.1.0](https://github.com/mysql-net/MySqlConnector/releases/tag/2.1.0); thanks for the PR!

PR #1075, opened 2021-11-08
added MySqlDecimal support

sumitdvlp:
Implementing solution for this [issue](https://github.com/mysql-net/MySqlConnector/issues/1070) raised earlier.
Signed-off-by: Sumit Kumar sumitdvlp@gmail.com

### DIFF
--- a/src/MySqlConnector/Core/Row.cs
+++ b/src/MySqlConnector/Core/Row.cs
@@ -441,6 +441,14 @@ internal abstract class Row
 		throw new InvalidCastException("Can't convert {0} to MySqlGeometry.".FormatInvariant(ResultSet.ColumnDefinitions![ordinal].ColumnType));
 	}
 
+	public MySqlDecimal GetMySqlDecimal(int ordinal)
+	{
+		var value = GetValue(ordinal);
+		if (value is string decimalString && ResultSet.ColumnDefinitions![ordinal].ColumnType == ColumnType.Decimal)
+			return new MySqlDecimal(decimalString);
+		throw new InvalidCastException("Can't convert {0} to MySqlDecimal.".FormatInvariant(ResultSet.ColumnDefinitions![ordinal].ColumnType));
+	}
+
 	public int GetValues(object[] values)
 	{
 		int count = Math.Min((values ?? throw new ArgumentNullException(nameof(values))).Length, ResultSet.ColumnDefinitions!.Length);
--- a/src/MySqlConnector/Core/Row.cs
+++ b/src/MySqlConnector/Core/Row.cs
@@ -241,8 +241,6 @@ internal abstract class Row
 
 	protected abstract int GetInt32Core(ReadOnlySpan<byte> data, ColumnDefinitionPayload columnDefinition);
 
-	protected abstract MySqlDecimal GetMySqlDecimalAsStringCore(ReadOnlySpan<byte> data, ColumnDefinitionPayload columnDefinition);
-
 	public long GetInt64(int ordinal)
 	{
 		var value = GetValue(ordinal);
--- a/src/MySqlConnector/Core/Row.cs
+++ b/src/MySqlConnector/Core/Row.cs
@@ -241,6 +241,8 @@ internal abstract class Row
 
 	protected abstract int GetInt32Core(ReadOnlySpan<byte> data, ColumnDefinitionPayload columnDefinition);
 
+	protected abstract MySqlDecimal GetMySqlDecimalAsStringCore(ReadOnlySpan<byte> data, ColumnDefinitionPayload columnDefinition);
+
 	public long GetInt64(int ordinal)
 	{
 		var value = GetValue(ordinal);
@@ -443,9 +445,14 @@ internal abstract class Row
 
 	public MySqlDecimal GetMySqlDecimal(int ordinal)
 	{
-		var value = GetValue(ordinal);
-		if (value is string decimalString && ResultSet.ColumnDefinitions![ordinal].ColumnType == ColumnType.Decimal)
-			return new MySqlDecimal(decimalString);
+		var data = m_data.Slice(m_dataOffsets[ordinal], m_dataLengths[ordinal]).Span;
+		var columnType = ResultSet.ColumnDefinitions![ordinal].ColumnType;
+		if (columnType == ColumnType.NewDecimal || columnType == ColumnType.Decimal)
+		{
+			var value = Encoding.UTF8.GetString(data);
+			return new MySqlDecimal(value);
+		}
+
 		throw new InvalidCastException("Can't convert {0} to MySqlDecimal.".FormatInvariant(ResultSet.ColumnDefinitions![ordinal].ColumnType));
 	}
 
--- a/src/MySqlConnector/Core/Row.cs
+++ b/src/MySqlConnector/Core/Row.cs
@@ -445,12 +445,8 @@ internal abstract class Row
 	{
 		var data = m_data.Slice(m_dataOffsets[ordinal], m_dataLengths[ordinal]).Span;
 		var columnType = ResultSet.ColumnDefinitions![ordinal].ColumnType;
-		if (columnType == ColumnType.NewDecimal || columnType == ColumnType.Decimal)
-		{
-			var value = Encoding.UTF8.GetString(data);
-			return new MySqlDecimal(value);
-		}
-
+		if (columnType is ColumnType.NewDecimal or ColumnType.Decimal)
+			return new MySqlDecimal(Encoding.UTF8.GetString(data));
 		throw new InvalidCastException("Can't convert {0} to MySqlDecimal.".FormatInvariant(ResultSet.ColumnDefinitions![ordinal].ColumnType));
 	}
 
--- a/src/MySqlConnector/Core/TextRow.cs
+++ b/src/MySqlConnector/Core/TextRow.cs
@@ -103,7 +103,7 @@ internal sealed class TextRow : Row
 
 		case ColumnType.Decimal:
 		case ColumnType.NewDecimal:
-			return Utf8Parser.TryParse(data, out decimal decimalValue, out int bytesConsumed) && bytesConsumed == data.Length ? decimalValue : throw new FormatException();
+			return Encoding.UTF8.GetString(data);
 
 		case ColumnType.Geometry:
 			return data.ToArray();
--- a/src/MySqlConnector/Core/TextRow.cs
+++ b/src/MySqlConnector/Core/TextRow.cs
@@ -127,8 +127,4 @@ internal sealed class TextRow : Row
 
 	private static long ParseInt64(ReadOnlySpan<byte> data) =>
 		!Utf8Parser.TryParse(data, out long value, out var bytesConsumed) || bytesConsumed != data.Length ? throw new FormatException() : value;
-	protected override MySqlDecimal GetMySqlDecimalAsStringCore(ReadOnlySpan<byte> data, ColumnDefinitionPayload columnDefinition)
-	{
-		return new MySqlDecimal(Encoding.UTF8.GetString(data));
-	}
 }
--- a/src/MySqlConnector/Core/TextRow.cs
+++ b/src/MySqlConnector/Core/TextRow.cs
@@ -103,7 +103,7 @@ internal sealed class TextRow : Row
 
 		case ColumnType.Decimal:
 		case ColumnType.NewDecimal:
-			return Encoding.UTF8.GetString(data);
+			return Utf8Parser.TryParse(data, out decimal decimalValue, out int bytesConsumed) && bytesConsumed == data.Length ? decimalValue : throw new FormatException();
 
 		case ColumnType.Geometry:
 			return data.ToArray();
@@ -127,4 +127,8 @@ internal sealed class TextRow : Row
 
 	private static long ParseInt64(ReadOnlySpan<byte> data) =>
 		!Utf8Parser.TryParse(data, out long value, out var bytesConsumed) || bytesConsumed != data.Length ? throw new FormatException() : value;
+	protected override MySqlDecimal GetMySqlDecimalAsStringCore(ReadOnlySpan<byte> data, ColumnDefinitionPayload columnDefinition)
+	{
+		return new MySqlDecimal(Encoding.UTF8.GetString(data));
+	}
 }
--- a/src/MySqlConnector/MySqlDecimal.cs
+++ b/src/MySqlConnector/MySqlDecimal.cs
@@ -1,104 +1,33 @@
 using System;
 using System.Globalization;
+using System.Text.RegularExpressions;
 
 namespace MySqlConnector;
 
-	public struct MySqlDecimal : IComparable, IComparable<MySqlDecimal>, IConvertible, IEquatable<MySqlDecimal>
+public readonly struct MySqlDecimal
+{
+private readonly string value;
+
+internal MySqlDecimal(string val)
+{
+	string regexWithDecimal = @"^-?([1-9][0-9]*|0)(\.[0-9]+)?$";
+	string regexWithOutDecimal = @"^-?([1-9][0-9]*|0)$";
+	var matchWithDecimal = Regex.Match(val, regexWithDecimal, RegexOptions.IgnoreCase);
+	var matchWithOutDecimal = Regex.Match(val, regexWithOutDecimal, RegexOptions.IgnoreCase);
+
+	if (!(matchWithDecimal.Success && val.Length <= 66) && !(matchWithOutDecimal.Success && val.Length <= 67))
 	{
-		private readonly string value;
-
-		internal MySqlDecimal(string val)
-		{
-			if (val.Length > 65)
-			{
-				throw new MySqlConversionException("Value is too large for Conversion");
-			}
-			value = val;
-		}
-
-		public decimal Value => Convert.ToDecimal(value, CultureInfo.InvariantCulture);
-
-		public int CompareTo(object? obj) => obj is MySqlDecimal other ?
-			((IComparable<MySqlDecimal>) this).CompareTo(other) :
-			throw new ArgumentException("CompareTo can only be called with another MySqlDateTime", nameof(obj));
-
-		public int CompareTo(MySqlDecimal other)
-		{
-			if(other.value == value)
-			{
-				return 1;
-			}
-			else
-			{
-				return -1;
-			}
-		}
-
-		public bool Equals(MySqlDecimal other)
-		{
-			return value == other.value;
-		}
-		public TypeCode GetTypeCode()
-		{
-			return TypeCode.Decimal;
-		}
-		public bool ToBoolean(IFormatProvider? provider) => throw new InvalidCastException();
-		public byte ToByte(IFormatProvider? provider)
-		{
-			return Byte.Parse(value, provider);
-		}
-		public char ToChar(IFormatProvider? provider) => throw new InvalidCastException();
-		public DateTime ToDateTime(IFormatProvider? provider) => throw new InvalidCastException();
-		public decimal ToDecimal(IFormatProvider? provider)
-		{
-			return Decimal.Parse(value, provider);
-		}
-
-		public double ToDouble()
-		{
-			return Double.Parse(value, CultureInfo.InvariantCulture);
-		}
-
-		public double ToDouble(IFormatProvider? provider)
-		{
-			return Double.Parse(value, provider);
-		}
-		public short ToInt16(IFormatProvider? provider)
-		{
-			return Int16.Parse(value, provider);
-		}
-		public int ToInt32(IFormatProvider? provider)
-		{
-			return Int32.Parse(value, provider);
-		}
-		public long ToInt64(IFormatProvider? provider)
-		{
-			return Int64.Parse(value, provider);
-		}
-		public sbyte ToSByte(IFormatProvider? provider)
-		{
-			return SByte.Parse(value, provider);
-		}
-		public float ToSingle(IFormatProvider? provider)
-		{
-			return Single.Parse(value, provider);
-		}
-		public override string ToString() => value;
-		public string ToString(IFormatProvider? provider)
-		{
-			return value;
-		}
-		public object ToType(Type conversionType, IFormatProvider? provider) => throw new InvalidCastException();
-		public ushort ToUInt16(IFormatProvider? provider)
-		{
-			return UInt16.Parse(value, provider);
-		}
-		public uint ToUInt32(IFormatProvider? provider)
-		{
-			return UInt32.Parse(value, provider);
-		}
-		public ulong ToUInt64(IFormatProvider? provider)
-		{
-			return UInt64.Parse(value, provider);
-		}
+		throw new MySqlConversionException("Value is too large for Conversion or Format is wrong.");
 	}
+	value = val;
+}
+
+public decimal Value => Convert.ToDecimal(value, CultureInfo.InvariantCulture);
+
+public double ToDouble()
+{
+	return Double.Parse(value, CultureInfo.InvariantCulture);
+}
+
+public override string ToString() => value;
+}
--- a/src/MySqlConnector/MySqlDecimal.cs
+++ b/src/MySqlConnector/MySqlDecimal.cs
@@ -1,48 +1,37 @@
 using System.Globalization;
 using System.Text.RegularExpressions;
+using MySqlConnector.Utilities;
 
 namespace MySqlConnector;
 
 public readonly struct MySqlDecimal
 {
-	private readonly string value;
-	static readonly string regexWithDecimal = @"^-?([1-9][0-9]*|0)(\.[0-9]+)$";
-	static readonly string regexWithOutDecimal = @"^-?([1-9][0-9]*)$";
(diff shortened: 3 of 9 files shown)


src/MySqlConnector/MySqlDecimal.cs
```diff
@@ -0,0 +1,104 @@
+using System;
+using System.Globalization;
+
+namespace MySqlConnector;
+
+	public struct MySqlDecimal : IComparable, IComparable<MySqlDecimal>, IConvertible, IEquatable<MySqlDecimal>
```
bgrainger, line 6: This should be a `readonly struct`.

Unnecessary indentation here (and through rest of file).

src/MySqlConnector/MySqlDecimal.cs
```diff
@@ -0,0 +1,104 @@
...
+			if (val.Length > 65)
+			{
+				throw new MySqlConversionException("Value is too large for Conversion");
+			}
+			value = val;
+		}
+
+		public decimal Value => Convert.ToDecimal(value, CultureInfo.InvariantCulture);
+
+		public int CompareTo(object? obj) => obj is MySqlDecimal other ?
```
bgrainger, line 21: Should be an explicit interface implementation.

bgrainger:
```suggestion
		int IComparable.CompareTo(object? obj) => obj is MySqlDecimal other ?
```

src/MySqlConnector/MySqlDecimal.cs
```diff
@@ -0,0 +1,104 @@
...
+			((IComparable<MySqlDecimal>) this).CompareTo(other) :
```
bgrainger, line 22: Cast shouldn't be necessary if `CompareTo` is public (which it is).

src/MySqlConnector/MySqlDecimal.cs
```diff
@@ -0,0 +1,104 @@
...
+			throw new ArgumentException("CompareTo can only be called with another MySqlDateTime", nameof(obj));
```
bgrainger, line 23:
```suggestion
			throw new ArgumentException("CompareTo can only be called with another MySqlDecimal", nameof(obj));
```

src/MySqlConnector/MySqlDecimal.cs
```diff
@@ -0,0 +1,104 @@
...
+		{
+			if(other.value == value)
+			{
+				return 1;
+			}
+			else
+			{
+				return -1;
+			}
+		}
```
bgrainger, line 35: `CompareTo` needs to return `0` if both values are equal; otherwise, it needs to return a complete ordering over all values. 

I don't think it's necessary to implement `IComparable<T>`; it wasn't in the API I suggested here: https://github.com/mysql-net/MySqlConnector/issues/1070#issuecomment-952193539

src/MySqlConnector/MySqlDecimal.cs
bgrainger: I don't think `IConvertible`, `IComparable`, `IComparable<T>` are necessary to implement; my API sketch (in https://github.com/mysql-net/MySqlConnector/issues/1070#issuecomment-952193539) didn't call for them.

Even `IEquatable<T>` could be dropped; are `2.5` and `2.500` "equal" or not? Implementing it would just add unnecessary complexity.

src/MySqlConnector/MySqlDecimal.cs
```diff
@@ -0,0 +1,104 @@
...
+
+namespace MySqlConnector;
+
+	public struct MySqlDecimal : IComparable, IComparable<MySqlDecimal>, IConvertible, IEquatable<MySqlDecimal>
+	{
+		private readonly string value;
+
+		internal MySqlDecimal(string val)
+		{
+			if (val.Length > 65)
```
bgrainger, line 12: I _think_ this is incorrect, because 65 is the number of digits, not including the `.` and optional `-` character. (Would need to double-check though.)

bgrainger: I'd like to see additional validation here, such as `/^-?([1-9][0-9]*|0)(\.[0-9]+)?$/`.

bgrainger (changes requested):
`MySqlDecimal` doesn't need to implement all those interfaces.

Needs tests. (I know you noted those were missing.)

> All PRs that add new functionality must also include tests that verify the new functionality.

Need to add support to `MySqlParameter` to write values of this type.

All commits need a DCO signoff (in the commit message, not a PR comment).

src/MySqlConnector/MySqlDecimal.cs
sumitdvlp: I am removing implementation of IComparable<T> interface all together now

sumitdvlp: I am removing implementation of IComparable interface all together now,
Thanks for the review.

src/MySqlConnector/MySqlDecimal.cs
sumitdvlp: Not needed as I am removing implementation of IComparable interface all together now,
Thanks for the review.

src/MySqlConnector/MySqlDecimal.cs
sumitdvlp: Not needed as I am removing implementation of IComparable interface all together now,
Thanks for the review.

src/MySqlConnector/MySqlDecimal.cs
sumitdvlp: Not needed as I am removing implementation of IComparable interface all together now,
Thanks for the review.

src/MySqlConnector/MySqlDecimal.cs
sumitdvlp: Got it.
Thanks for the review

src/MySqlConnector/MySqlDecimal.cs
sumitdvlp: Reading from  [here](https://dev.mysql.com/doc/refman/8.0/en/fixed-point-types.html) and [here](https://dev.mysql.com/doc/refman/8.0/en/precision-math-decimal-characteristics.html)

The declaration syntax for a DECIMAL column is DECIMAL(M,D). The ranges of values for the arguments are as follows:

M is the maximum number of digits (the precision). It has a range of 1 to 65.
D is the number of digits to the right of the decimal point (the scale). It has a range of 0 to 30 and must be no larger than M.
DECIMAL columns do not store a leading + character or - character or leading 0 digits. If you insert +0003.1 into a DECIMAL(5,1) column, it is stored as 3.1. For negative numbers, a literal - character is not stored.

Not sure how _-999.99_ which can be seen in mysql table on mysql workbench is stored and how it will be converted to a string and then converted back.

bgrainger:
```sql
create table test(dec1 decimal(65,0), dec2 decimal(65,30));
insert into test(dec1, dec2) values('-12345678901234567890123456789012345678901234567890123456789012345.0', '-12345678901234567890123456789012345.012345678901234567890123456789');
select * from test;
```

dec1 | dec2
--- | ---
-12345678901234567890123456789012345678901234567890123456789012345 | -12345678901234567890123456789012345.012345678901234567890123456789

String lengths are 66 and 67 respectively.

src/MySqlConnector/MySqlDecimal.cs
sumitdvlp: I am removing the length check for now just checking format using regex.

sumitdvlp:
> ```sql
> create table test(dec1 decimal(65,0), dec2 decimal(65,30));
> insert into test(dec1, dec2) values('-12345678901234567890123456789012345678901234567890123456789012345.0', '-12345678901234567890123456789012345.012345678901234567890123456789');
> select * from test;
> ```
> 
> dec1	dec2
> -12345678901234567890123456789012345678901234567890123456789012345	-12345678901234567890123456789012345.012345678901234567890123456789
> String lengths are 66 and 67 respectively.

I am removing the length check for now just checking the format using regex.

bgrainger:
I think any tests could live in `InsertTests`, similar to the ones added for `BigInteger`: https://github.com/mysql-net/MySqlConnector/commit/39a3e8f031b5149fb40ecd5dcff799016171b904

There should also be new methods on `MySqlDataReader`, similar to `MySqlGeometry` https://github.com/mysql-net/MySqlConnector/commit/3764da49257a089b55e6e6f33fdefb7470bf168a and `DateOnly`/`TimeOnly` https://github.com/mysql-net/MySqlConnector/commit/39a3e8f031b5149fb40ecd5dcff799016171b904. (You don't need to add tests to `DataTypes.cs`, as was done in those commits.)

Finally, create a `MySqlDecimalTests` class (in MySqlConnector.Tests) that performs unit tests of the `MySqlDecimal` type; this should be shared between the `BASELINE` and non-`BASELINE` tests to verify that the API is as similar as possible between MySql.Data and MySqlConnector.

bgrainger:
Hopefully you can run and debug the tests locally (with a Docker container), but you should also be able to see the results here: https://dev.azure.com/mysqlnet/MySqlConnector/_build/results?buildId=1959&view=ms.vss-test-web.build-test-results-tab

tests/SideBySide/InsertTests.cs
```diff
@@ -411,6 +411,52 @@ public void InsertBigInteger(bool prepare)
...
+
+	[Theory]
+	[InlineData(false)]
+	[InlineData(true)]
+	public void InsertMySqlDecimalAsDecimal(bool prepare)
+	{
+		using var connection = new MySqlConnection(AppConfig.ConnectionString);
+		connection.Open();
+		connection.Execute(@"drop table if exists insert_mysql_decimal;
+create table insert_mysql_decimal(rowid integer not null primary key auto_increment, value decimal(40, 2));");
```
bgrainger, line 445: `decimal(40, 2)` seems like a mistake, as `-12345678901234567890123456789012345.012345678901234567890123456789` will not fit in that column size. Did you mean `decimal(65, 30)`?

tests/SideBySide/InsertTests.cs
```diff
@@ -411,6 +411,52 @@ public void InsertBigInteger(bool prepare)
...
 
+	[Theory]
+	[InlineData(false)]
+	[InlineData(true)]
+	public void InsertMySqlDecimal(bool prepare)
+	{
+		using var connection = new MySqlConnection(AppConfig.ConnectionString);
+		connection.Open();
+		connection.Execute(@"drop table if exists insert_mysql_decimal;
+create table insert_mysql_decimal(rowid integer not null primary key auto_increment, value bigint);");
```
bgrainger, line 422: `value bigint` seems like the wrong data type.

sumitdvlp: sorry about this stupid mistake

tests/SideBySide/InsertTests.cs
sumitdvlp: Yes, you are right sorry about this stupid mistake.

sumitdvlp:
@bgrainger Can you help me figure out what changes are needed [here](https://github.com/mysql-net/MySqlConnector/blob/master/src/MySqlConnector/Core/TextRow.cs#:~:text=case%20ColumnType.NewDecimal,throw%20new%20FormatException()%3B) to make InsertTests.cs to work.
We need to convert current state or decimal won't work. 
Should I consider converting it to string like [here](https://github.com/mysql-net/MySqlConnector/blob/master/src/MySqlConnector/Core/TextRow.cs#:~:text=case%20ColumnType.String,case%20ColumnType.VarString%3B) 
Like this - return Encoding.UTF8.GetString(data);

bgrainger:
`Decimal` values should not be returned as a `string` by default. That will break a lot of code.

Needing to use `MySqlDecimal` is a very infrequent use case, and I'm happy for it to be 100% opt-in (by calling `MySqlDataReader.GetMySqlDecimal` in the cases where it's needed). The implementation will likely need to retrieve the string directly instead of delegating to `GetValue`, which would still try to read the column as a `decimal` and fail for large values.

src/MySqlConnector/Core/Row.cs
```diff
@@ -441,6 +441,14 @@ public MySqlGeometry GetMySqlGeometry(int ordinal)
 		throw new InvalidCastException("Can't convert {0} to MySqlGeometry.".FormatInvariant(ResultSet.ColumnDefinitions![ordinal].ColumnType));
 	}
 
+	public MySqlDecimal GetMySqlDecimal(int ordinal)
+	{
+		var value = GetValue(ordinal);
```
bgrainger, line 446: Will likely need a `GetMySqlDecimalCore` method that TextRow and BinaryRow can override to read the raw bytes as UTF-8 and return a string-based MySqlDecimal. (Or `GetMySqlDecimalAsStringCore` if both row kinds can return UTF8 bytes and the common logic can be centralised here.)

src/MySqlConnector/Core/TextRow.cs
```diff
@@ -103,7 +103,7 @@ protected override object GetValueCore(ReadOnlySpan<byte> data, ColumnDefinition
 
 		case ColumnType.Decimal:
 		case ColumnType.NewDecimal:
-			return Utf8Parser.TryParse(data, out decimal decimalValue, out int bytesConsumed) && bytesConsumed == data.Length ? decimalValue : throw new FormatException();
+			return Encoding.UTF8.GetString(data);
```
bgrainger, line 106: We can't change the default type of Decimal columns from `decimal` to `string`.

src/MySqlConnector/MySqlDecimal.cs
```diff
@@ -0,0 +1,48 @@
...
+public readonly struct MySqlDecimal
+{
+	private readonly string value;
+
+	internal MySqlDecimal(string val)
+	{
+		string regexWithDecimal = @"^-?([1-9][0-9]*|0)(\.[0-9]+)?$";
+		string regexWithOutDecimal = @"^-?([1-9][0-9]*|0)$";
+		var matchWithDecimal = Regex.Match(val, regexWithDecimal, RegexOptions.IgnoreCase);
+		var matchWithOutDecimal = Regex.Match(val, regexWithOutDecimal, RegexOptions.IgnoreCase);
```
bgrainger, line 16: These could be `static readonly Regex` fields to avoid depending on the global regex cache here. I also think they could be combined into one regex.

These regexes allow `-0` which arguably shouldn't be allowed (as MySQL doesn't support the concept of negative zero AFAIK).

src/MySqlConnector/MySqlDecimal.cs
```diff
@@ -0,0 +1,48 @@
...
+
+		bool negative = val[0] == '-';
+		// If its valid negative value with . then length should be less than 68
+		bool withDecimalNegative = (matchWithDecimal.Success && (val.Length >= 68 && negative));
+		// If its valid positive value with . then length should be less than 67
+		bool withDecimalPositive = (matchWithDecimal.Success && (val.Length >= 67 && !negative));
+		// If its valid negative value without . then length should be less than 67
+		bool withOutDecimalNegative = (matchWithOutDecimal.Success && (val.Length >= 67 && negative));
+		// If its valid positive value without . then length should be less than 66
+		bool withOutDecimalPositive = (matchWithOutDecimal.Success && (val.Length >= 66 && !negative));
```
bgrainger, line 31: Length constraints could probably be built into the regex instead of tested imperatively.

src/MySqlConnector/MySqlDecimal.cs
```diff
@@ -0,0 +1,48 @@
...
+	internal MySqlDecimal(string val)
+	{
+		string regexWithDecimal = @"^-?([1-9][0-9]*|0)(\.[0-9]+)?$";
+		string regexWithOutDecimal = @"^-?([1-9][0-9]*|0)$";
+		var matchWithDecimal = Regex.Match(val, regexWithDecimal, RegexOptions.IgnoreCase);
+		var matchWithOutDecimal = Regex.Match(val, regexWithOutDecimal, RegexOptions.IgnoreCase);
+
+		if(!(matchWithDecimal.Success || matchWithOutDecimal.Success))
+		{
+			throw new MySqlConversionException("Format is wrong.");
```
bgrainger, line 20: Use `FormatException`. (`MySqlConversionException` was for backwards compatibility for MySql.Data and is documented as "thrown when a MySQL value can't be converted to another type", which isn't true here: we're trying to construct a MySQL value.)

bgrainger (changes requested):

src/MySqlConnector/MySqlDecimal.cs
sumitdvlp: its a good idea but I am afraid it might induce bugs, length check is simple and understandable with comments.

src/MySqlConnector/MySqlDecimal.cs
sumitdvlp: right , regexWithOutDecimal is updated to take care of this bug here

src/MySqlConnector/Core/TextRow.cs
sumitdvlp: Okay I have added it back

sumitdvlp: @bgrainger Can you review the new iteration?

src/MySqlConnector/Core/Row.cs
```diff
@@ -241,6 +241,8 @@ public int GetInt32(int ordinal)
 
 	protected abstract int GetInt32Core(ReadOnlySpan<byte> data, ColumnDefinitionPayload columnDefinition);
 
+	protected abstract MySqlDecimal GetMySqlDecimalAsStringCore(ReadOnlySpan<byte> data, ColumnDefinitionPayload columnDefinition);
```
bgrainger, line 244: When is this method used?

tests/MySqlConnector.Tests/MySqlDecimalTests.cs
```diff
@@ -0,0 +1,99 @@
...
+		var invalidValue = "123456789012345678901234567890123456012345678901234567890123456789";
+		Assert.Throws<FormatException>(() => new MySqlDecimal(invalidValue));
+	}
+
+	[Fact]
+	public void TestValidFormatWithOutDecimalNegativeZero()
+	{
+		// If its valid positive value without . then length should be less than 66
+		var invalidValue = "-0";
+		Assert.Throws<FormatException>(() => new MySqlDecimal(invalidValue));
```
bgrainger, line 88: This test appears to be failing (because an exception is not thrown).

src/MySqlConnector/Core/Row.cs
sumitdvlp: whenever the user have data type.
I have only added this because you asked me to in earlier comments, btw its same as method above GetInt32core maybe we want to keep the consistancy
i can remove it if you want me to

tests/MySqlConnector.Tests/MySqlDecimalTests.cs
sumitdvlp: oh yes , not sure how i missed it , fixed the regex for it

src/MySqlConnector/Core/Row.cs
bgrainger: I didn't ask for it; I merely suggested that "[you] will likely need" such a method based on how your code looked at the time. Since then, you've changed the code.

But as far as I can tell, this method is never invoked, so it's dead code. If this method is never called, please delete it. (And if it is called, I must have missed it.)

bgrainger:
Please ensure that the "baseline" tests (that run this code against MySql.Data) work. You can run them locally (no MySQL instance needed) with:

```
cd tests\MySqlConnector.Tests
dotnet restore /p:Configuration=Baseline; dotnet test -c Baseline
```

Add `#if BASELINE` or `#if !BASELINE` around just the parts that are incompatible with MySql.Data (I would expect most of the unit tests to still pass, which verifies that MySqlConnector has a similar API).

src/MySqlConnector/Core/Row.cs
sumitdvlp: oh okay, sorry for the confusion, I will delete it, and thank you for your suggestions

bgrainger:
Just to be clear: all checks need to pass for this to be able to be merged.

And I generally don't start reviewing code until the tests are passing, because I prefer not to review code that isn't known to be working.

sumitdvlp:
> Just to be clear: all checks need to pass for this to be able to be merged.
> 
> And I generally don't start reviewing code until the tests are passing, because I prefer not to review code that isn't known to be working.

@bgrainger all test cases are running now. Please let me know if I can handle any other issues.

src/MySqlConnector/MySqlDecimal.cs
```diff
@@ -0,0 +1,48 @@
...
+
+		bool negative = val[0] == '-';
+		// If its valid negative value with . then length should be less than 68
+		bool withDecimalNegative = (matchWithDecimal.Success && (val.Length >= 68 && negative));
+		// If its valid positive value with . then length should be less than 67
+		bool withDecimalPositive = (matchWithDecimal.Success && (val.Length >= 67 && !negative));
+		// If its valid negative value without . then length should be less than 67
+		bool withOutDecimalNegative = (matchWithOutDecimal.Success && (val.Length >= 67 && negative));
+		// If its valid positive value without . then length should be less than 66
+		bool withOutDecimalPositive = (matchWithOutDecimal.Success && (val.Length >= 66 && !negative));
```
bgrainger, line 31: Could be simplified by testing the length of each digit group, rather than implicitly including the `-` and `.`.

src/MySqlConnector/MySqlDecimal.cs
```diff
@@ -0,0 +1,48 @@
+using System.Globalization;
+using System.Text.RegularExpressions;
+
+namespace MySqlConnector;
+
+public readonly struct MySqlDecimal
+{
+	private readonly string value;
+	static readonly string regexWithDecimal = @"^-?([1-9][0-9]*|0)(\.[0-9]+)$";
+	static readonly string regexWithOutDecimal = @"^-?([1-9][0-9]*)$";
```
bgrainger, line 10: Doesn't match naming conventions for private fields.

tests/MySqlConnector.Tests/MySqlDecimalTests.cs
```diff
@@ -0,0 +1,106 @@
+using System;
+using Xunit;
+
+namespace MySqlConnector.Tests;
+public class MySqlDecimalTests
+{
+#if !BASELINE
```
bgrainger, line 7: I missed earlier that MySql.Data has no public constructor for this type, so that there's no way to test it. In that case, the entire file can be excluded in the csproj instead of `#if`'ed out here.

tests/MySqlConnector.Tests/MySqlDecimalTests.cs
```diff
@@ -0,0 +1,106 @@
...
+	[Fact]
+	public void TestValidFormatWithDecimalNegative67Length()
+	{
+		// valid value with negative and decimal
+		var invalidValue = "-12345678901234567890123456789012345.012345678901234567890123456789";
+		MySqlDecimal decimalVal = new MySqlDecimal(invalidValue);
+		Assert.Equal(invalidValue, decimalVal.ToString());
+	}
+#endif
+}
```
bgrainger, line 106: Good breadth of tests 👍. Some overhead could be avoided by using `[Theory]` to add test cases without duplicating test code.

Missing test for `new MySqlDecimal("0")`, which (incorrectly) throws.

tests/SideBySide/InsertTests.cs
```diff
@@ -411,6 +411,70 @@ public void InsertBigInteger(bool prepare)
...
+		mySqlCommand.CommandText = @"insert into insert_mysql_decimal(value) values(@value);";
+		mySqlCommand.Parameters.AddWithValue("@value", new MySqlDecimal(value));
+		mySqlCommand.ExecuteNonQuery();
+		connection.Close();
+		connection.Open();
+
+		mySqlCommand.CommandText = @"select value from insert_mysql_decimal order by rowid;";
+		MySqlDataReader mySqlDataReader = mySqlCommand.ExecuteReader();
+		Assert.True(mySqlDataReader.Read());
+		var val = mySqlDataReader.GetMySqlDecimal(0);
```
bgrainger, line 473: Could test `GetMySqlDecimal("value")` overload; could test `GetFieldValue<MySqlDecimal>`.

bgrainger (changes requested):
There are bugs in the implementation that a full complement of tests would have caught.

Need to add further tests and fix the bugs they reveal; I will push a commit to do this.

bgrainger (approved):

src/MySqlConnector/MySqlDecimal.cs
alexr17: @bgrainger 

Turns out `/^-?([1-9][0-9]*|0)(\.[0-9]+)?$/` should have been `/^-?([0-9][0-9]*|0)(\.[0-9]+)?$/` to support zero fill decimals.

Will submit a PR for this later on at some point, just wanted to comment on the reference now that I've found it.

bgrainger: FWIW `[0-9][0-9]*` is equivalent to `[0-9]+` and then the trailing `|0` is superfluous.

bgrainger: https://github.com/mysql-net/MySqlConnector/issues/1354

Issue #1119, opened 2022-01-18
MySqlParameterCollection : Add Constructor/Method to MySqlParameterCollection which allow memory allocation in advance for m_parameters if the size is known 


sumitdvlp:
**Is your feature request related to a problem? Please describe.**

Everytime I add new Parameter in MySqlParameterCollection, it might have to resize which might cause perf issues if I am adding many Parameters, If i know number parameters I planning to add , it does make sense for me allocate that much memory upfront to avoid overfill or allocating again and again.

**Describe the solution you'd like**

Current constructor is internal not sure why but maybe we can add new method which can allocate the element in the list (m_parameters) or assign a list to m_parameters.

**Describe alternatives you've considered**
I looked at the code I haven't found anything which can be used to pre allocate number of items to m_parameters.

**Additional context**

badbod99:
m_parameters within MySqlParameterCollection is a List.  Sure, it might have to internally resize, but this is optimised in dotnet.  The internal array resized to double it's current size.  It starts at a size of 4.

This really isn't going to cause you any performance issues.

sumitdvlp:
We actually in some cases (where perf matters) add thousands of parameters to MySqlParameterCollection.
We are hoping that because its ArrayList( worse if its a LinkedList) it will give us slight improvement in perf when we pre allocate the size in memory.

Where we use MySqlParameterCollection? To make bulk insert statement.
Why not use MySqlBulkCopy? MySqlBulkCopy use file stream which is where we cannot implement our retry strategy, 
we are afraid that in case of failure it might leave some data behind (maybe we are wrong) and we want write parallelly to a single table.

sumitdvlp:
> m_parameters within MySqlParameterCollection is a List. Sure, it might have to internally resize, but this is optimised in dotnet. The internal array resized to double it's current size. It starts at a size of 4.
> 
> This really isn't going to cause you any performance issues.

I think it will when you have thousands of parameters and they are re-allocated everytime when it resizes.

bgrainger:
> MySqlBulkCopy use file stream

Are you thinking of `MySqlBulkLoader`? `MySqlBulkCopy` can load from a `DataTable` or even `IEnumerable<DataRow>`.

>> This really isn't going to cause you any performance issues.
> 
> I think it will when you have thousands of parameters and they are re-allocated everytime when it resizes.

[...]

PR #1127, opened 2022-02-03
Add Method to MySqlParameterCollection datastructure which allow memory allocation in advance for m_parameters if the size is known

sumitdvlp:
Add Constructor/Method to MySqlParameterCollection which allow memory allocation in advance for m_parameters if the size is known.

Related to issue : https://github.com/mysql-net/MySqlConnector/issues/1119

### DIFF
--- a/src/MySqlConnector/MySqlParameterCollection.cs
+++ b/src/MySqlConnector/MySqlParameterCollection.cs
@@ -11,6 +11,13 @@ public sealed class MySqlParameterCollection : DbParameterCollection, IEnumerabl
 		m_nameToIndex = new(StringComparer.OrdinalIgnoreCase);
 	}
 
+	public void SetParameterCapacity(int capacity)
+	{
+		if(capacity < 0)
+			throw new ArgumentOutOfRangeException(nameof(capacity));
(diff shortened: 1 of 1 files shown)


src/MySqlConnector/MySqlParameterCollection.cs
```diff
@@ -11,6 +11,13 @@ internal MySqlParameterCollection()
 		m_nameToIndex = new(StringComparer.OrdinalIgnoreCase);
 	}
 
+	public void SetParameterCapacity(int capacity)
+	{
+		if(capacity < 0)
+			throw new ArgumentOutOfRangeException(nameof(capacity));
+		m_parameters.Capacity = capacity;
```
bgrainger, line 18:
```suggestion
		m_parameters.Capacity = capacity;
#if NETSTANDARD2_1_OR_GREATER || NETCOREAPP2_1_OR_GREATER
		m_nameToIndex.EnsureCapacity(size);
#endif
```

We should resize the other collection, too.

bgrainger:
> If you have alternative benchmarks that show that this change is an improvement, please provide them.

No benchmarks justifying this change have been provided; can reopen if they are.

PR #1131, opened 2022-02-05
Fix deadlock when command is being cancelled from two threads

ahydrax:
Hi there, 
A couple of days ago we encountered a deadlock inside mysql connector which lead our system to downtime.

**What happened?**
Our monitoring detected that there are no new writes are happened to database in about 10 minutes. We dumped process memory and started looking for a problem. It turned out that deadlock happened in TimerQueue and connection Session.

**The problem**
The problem is that timer callback is being called under TimerQueue lock. Since callback can be almost any user code no one restricts it from acquiring other locks inside which exactly happened in our case. 
`ServerSession.TryStartCancel`  and `ServerSession.DoCancel` already uses a lock inside to protect internal session state and when it's called from TimerQueue it also implicitly acquires the lock on TimerQueue.
I attached screenshots which clearly illustrates the problem.

**Reproduction**
I was able to create a pretty stable repro. All you need is to make cancellation happenned simultaneously both from external cancellation source (CancellationTokenSource with timeout/HttpContext.RequestAborted etc.) and command timeout watchdog.
You can examine the code here https://github.com/ahydrax/StableMysqlConnectorDeadlockRepro

**The solution**
From my perspective solution should be pretty simple: timer queue calls pending callbacks outside of TimerQueue internal lock.

**Closing notes**
After fixing deadlock I've encountered a small bug inside `Session.TryStartCancel` which didn't respect all command states inside during state verifying.

### DIFF
--- a/src/MySqlConnector/Core/ServerSession.cs
+++ b/src/MySqlConnector/Core/ServerSession.cs
@@ -1750,8 +1750,9 @@ internal sealed class ServerSession
 	{
 		if (m_state != state1 && m_state != state2 && m_state != state3 && m_state != state4 && m_state != state5 && m_state != state6)
 		{
-			Log.Error("Session{0} should have SessionStateExpected {1} or SessionStateExpected2 {2} or SessionStateExpected3 {3} but was SessionState {4}", m_logArguments[0], state1, state2, state3, m_state);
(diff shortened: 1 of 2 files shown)


bgrainger:
Thanks for the diagnosis and fix!

bgrainger:
Fixed in [2.1.6](https://github.com/mysql-net/MySqlConnector/releases/tag/2.1.6).

Issue #1132, opened 2022-02-09
yaSSL/Schannel TLS 1.2 workaround causing unintended TLS downgrade


lostatredrock:
**Software versions**
MySqlConnector version: 1.3.14 (issue also exists in latest version 2.1.5)
Server type (MySQL, MariaDB, Aurora, etc.) and version:  MySQL 8.0.27
.NET version: 5.0.14
(Optional) ORM NuGet packages and versions:  Pomelo.EntityFrameworkCore.MySql 5.0.3

**Describe the bug**
We have been experiencing connectivity issues in which our application server looses the ability to establish new connections to the database server.  Reviewing the logs we see repeats of the message: "The server doesn't support the client's specified TLS versions."

Stepping back to the start of this connectivity issue we are seeing the message "Session1.944 failed negotiating TLS; falling back to TLS 1.1" logged as a warning.

Looking into the connector code I see that this is part of a work around to an issue establishing TLS 1.2 connections between a windows server a yaSSL based MySQL instance ([comment](https://github.com/mysql-net/MySqlConnector/blob/e82f18f754323177cc27af3bc2157608b34be035/src/MySqlConnector/Core/ServerSession.cs#L391)).  In this case our MySQL instance is OpenSSL based and we are experiencing an IOException due to an unknown cause (we are still investigating the cause of the IO Exception).

Stepping into the connect code the crtiteria for retry is [here](https://github.com/mysql-net/MySqlConnector/blob/e82f18f754323177cc27af3bc2157608b34be035/src/MySqlConnector/Core/ServerSession.cs#L403) and the downgrade code is [here ](https://github.com/mysql-net/MySqlConnector/blob/e82f18f754323177cc27af3bc2157608b34be035/src/MySqlConnector/Core/ServerSession.cs#L478).  The effect of this is that if you are running a windows server and experience an IO Exception when establishing a new database connection your pool will permenatly fall back to TLS 1.1.  As far as I can see there is no way to override this fallback.  Specifying TLS 1.2 in the connection string will get overidden and specifying nothign (and leaving it to the OS) will also make the application subject to this downgrade.  

In our case the downgrade crashed the application until restart because we disallow TLS 1.0 and 1.1 connections at the OS level on the app and db servers.

**Exception**

**_The fallback is initiated when this exception occurs:_**

```
`System.IO.IOException: Unable to write data to the transport connection: An established connection was aborted by the software in your host machine..
[...]

PR #1134, opened 2022-02-10
 Fix Bug from Issue #1132

lostatredrock:
Change the logic used to determine when a connection attempt is failing
due to a lack of support for TLS 1.2+ in yaSSL-based MySQL Server so
that a downgrade no longer occurs if the user has explicitly configured
supported TLS Version(s) in the connection string.

Signed-off-by: Andrew Nagel <andrew.nagel@gmail.com>

### DIFF
--- a/src/MySqlConnector/Core/ServerSession.cs
+++ b/src/MySqlConnector/Core/ServerSession.cs
@@ -416,7 +416,7 @@ internal sealed class ServerSession
 			InitialHandshakePayload initialHandshake;
 			do
 			{
-				shouldRetrySsl = (sslProtocols == SslProtocols.None || (sslProtocols & SslProtocols.Tls12) == SslProtocols.Tls12) && Utility.IsWindows();
+				shouldRetrySsl = sslProtocols == SslProtocols.None && Utility.IsWindows();
 
 				var connected = false;
 				if (cs.ConnectionProtocol == MySqlConnectionProtocol.Sockets)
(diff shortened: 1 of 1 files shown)


src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -416,7 +416,7 @@ public async Task DisposeAsync(IOBehavior ioBehavior, CancellationToken cancella
 			InitialHandshakePayload initialHandshake;
 			do
 			{
-				shouldRetrySsl = (sslProtocols == SslProtocols.None || (sslProtocols & SslProtocols.Tls12) == SslProtocols.Tls12) && Utility.IsWindows();
+				shouldRetrySsl = sslProtocols == SslProtocols.None && Utility.IsWindows();
```
bgrainger, line 419: This logic is incorrect; we should still retry if the user put `TlsVersion=Tls11,Tls12` in their connection string, for example.

```suggestion
				shouldRetrySsl = (sslProtocols == SslProtocols.None || ((sslProtocols & SslProtocols.Tls12) == SslProtocols.Tls12) && (sslProtocols & (SslProtocols.Tls | SslProtocols.Tls11) != SslProtocols.None)) && Utility.IsWindows();
```

(That's getting a little hard to read, so could be refactored into some temporary variables.)

We'd also want to change line 528 to fall back to the non-1.2 protocols the user had specified (unless they used `None`, in which case 1.0/1.1).

bgrainger:
Thanks for the quick fix!

lostatredrock:
Updated the fix to address comments and allow the retry if the user included 1.0 or 1.1 in addition to 1.2 and also to restict the retry to be a subset of the original values.

PR #1171, opened 2022-05-13
Update activity status on failure MySqlDataReader.

qq362220083:
When MySqlDataReader has exception, the activity status needs to be updated.

### DIFF
--- a/src/MySqlConnector/MySqlDataReader.cs
+++ b/src/MySqlConnector/MySqlDataReader.cs
@@ -474,8 +474,13 @@ public sealed class MySqlDataReader : DbDataReader, IDbColumnSchemaGenerator
 				await dataReader.NextResultAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
 			}
 		}
-		catch (Exception)
+		catch (Exception ex)
 		{
+			if (activity is { IsAllDataRequested: true })
+			{
+				activity.SetException(ex);
+				activity.Stop();
+			}
 			dataReader.Dispose();
 			throw;
 		}
--- a/src/MySqlConnector/Utilities/ActivitySourceHelper.cs
+++ b/src/MySqlConnector/Utilities/ActivitySourceHelper.cs
@@ -17,7 +17,9 @@ internal static class ActivitySourceHelper
 	public const string NetPeerNameTagName = "net.peer.name";
 	public const string NetPeerPortTagName = "net.peer.port";
 	public const string NetTransportTagName = "net.transport";
+#if !NET6_0_OR_GREATER
 	public const string StatusCodeTagName = "otel.status_code";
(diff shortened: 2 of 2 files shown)


src/MySqlConnector/Utilities/ActivitySourceHelper.cs
```diff
@@ -36,7 +36,13 @@ internal static class ActivitySourceHelper
 		return activity;
 	}
 
-	public static void SetSuccess(this Activity activity) => activity.SetTag(StatusCodeTagName, "OK");
+	public static void SetSuccess(this Activity activity)
+	{
+		if (activity.Duration == TimeSpan.Zero)
```
bgrainger, line 41: What's the purpose of this change?

src/MySqlConnector/MySqlDataReader.cs
```diff
@@ -474,8 +474,13 @@ internal static async Task<MySqlDataReader> CreateAsync(CommandListPosition comm
...
 		}
-		catch (Exception)
+		catch (Exception ex)
 		{
+			if (activity is { IsAllDataRequested: true })
+			{
+				activity.SetException(ex);
+				activity.Stop();
+			}
 			dataReader.Dispose();
```
bgrainger, line 484: `MySqlDataReader.DisposeAsync` will still Stop the activity and set it to successful. Seems like that logic ought to change to if `if (Activity.Status == ActivityStatusCode.Unset)` to avoid overwriting the status if it's already an error.

bgrainger:
Thanks! This is definitely an omission in MySqlConnector's activity-handling code; thanks for fixing it. Left just a couple of comments about state management in the face of tracking this exception.

src/MySqlConnector/Utilities/ActivitySourceHelper.cs
bgrainger: Oh, is this a workaround for a lack of an `IsStopped` property (https://github.com/dotnet/runtime/issues/63353) to check if the Activity is still running? A comment would be helpful here, but I think checking `Status` would be clearer.

qq362220083: uh, activity only supports `Status` property in .net 6 and above.

Here is the resolve to set the Exception first, and then set the Success to cause the state to be overwritten.

set the Exception should have the highest priority.

bgrainger: Thanks; I hadn't checked the docs thoroughly to see where the API was available.

bgrainger:
Thanks for the contribution! This is included in [2.1.9](https://github.com/mysql-net/MySqlConnector/releases/tag/2.1.9).

Issue #1177, opened 2022-05-31
NullReferenceException in MySqlConnector.Utilities.TimerQueue.Callback


csandr:
**Software versions**
MySqlConnector version: 2.1.9
CoreCLR Version: 6.0.522.21309
.NET Version: 6.0.5

```
Description: The process was terminated due to an unhandled exception.
Exception Info: System.NullReferenceException: Object reference not set to an instance of an object.
   at MySqlConnector.MySqlConnection.Cancel(ICancellableCommand command, Int32 commandId, Boolean isCancel) in /_/src/MySqlConnector/MySqlConnection.cs:line 780
   at MySqlConnector.MySqlCommand.CancelCommandForTimeout() in /_/src/MySqlConnector/MySqlCommand.cs:line 424
   at MySqlConnector.Utilities.TimerQueue.Callback(Object obj) in /_/src/MySqlConnector/Utilities/TimerQueue.cs:line 92
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.TimerQueueTimer.Fire(Boolean isThreadPool)
   at System.Threading.TimerQueue.FireNextTimers()
```

[Link to the exception line](https://github.com/mysql-net/MySqlConnector/blob/62ad22da8564345f8b342ec76a53023f10821c00/src/MySqlConnector/MySqlConnection.cs#L780)

Last second logs

MySqlConnector.CommandExecutor:

```
04:06:00.2176 Trace Session1.225 ExecuteReader Synchronous CommandCount: 1 
04:06:01.3484 Trace Session1.225 ExecuteReader Synchronous CommandCount: 1 
04:06:02.5396 Trace Session1.225 ExecuteReader Synchronous CommandCount: 1 
04:06:02.6449 Trace Session1.226 ExecuteReader Synchronous CommandCount: 1 
04:06:02.9968 Trace Session1.225 ExecuteReader Synchronous CommandCount: 1 
04:06:03.1121 Trace Session1.226 ExecuteReader Synchronous CommandCount: 1 
```

MySqlConnector.ConnectionPool

```
04:06:00.1477 Trace Pool1 receiving Session1.226 back 
04:06:00.2176 Trace Pool1 returning pooled Session1.225 to caller; LeasedSessionsCount=2 
04:06:00.4365 Trace Pool1 receiving Session1.225 back 
[...]

PR #1183, opened 2022-06-14
Fix issue #1177

maicodio:
Fix for: 

- #1177 

The issue can happen when MySQL server is experiencing degraded performance.

Steps:

1. When a command starts being executed, the Cancel method is scheduled to run at the timeout. (MySqlCommand.cs#L406)
2. After the timeout period, the callback runs the Cancel method in an arbitrary thread.
3. The Cancel method hangs trying to open another connection for the kill query command. 
4. A socket or stream timeout exception happens, triggering a Dispose
5. The dispose method sets null on m_session.
6. The kill query connection fails to open (or fails to run) and there is a null reference exception trying to log the exception with the `m_session.Id`

Solution:

- The `m_session` attribute is backed up into a new variable on method startup.
- For settings and logs purposes, the new variable is used.
- For the `ServiceSession`'s method calls is used the original `m_session` with the null propagator to avoid calling theses methods when `m_session` is null;

### DIFF
--- a/src/MySqlConnector/MySqlConnection.cs
+++ b/src/MySqlConnector/MySqlConnection.cs
@@ -746,13 +746,14 @@ public sealed class MySqlConnection : DbConnection, ICloneable
 
 	internal void Cancel(ICancellableCommand command, int commandId, bool isCancel)
 	{
-		if (m_session is null || State != ConnectionState.Open || !m_session.TryStartCancel(command))
+		var session = m_session;
+		if (session is null || State != ConnectionState.Open || !session.TryStartCancel(command))
 		{
 			Log.Trace("Ignoring cancellation for closed connection or invalid CommandId {0}", commandId);
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


bgrainger:
I think I preferred the original version in 2f0e81ca0018aa231e59c3c360bdcb95199a7f79 that just had a local for `sessionId` and avoided potential confusion about when the code should use `session` vs when it should use `m_session`. I see a way to improve that code even further with C# pattern matching.

maicodio:
> I think I preferred the original version in [2f0e81c](https://github.com/mysql-net/MySqlConnector/commit/2f0e81ca0018aa231e59c3c360bdcb95199a7f79) that just had a local for `sessionId` and avoided potential confusion about when the code should use `session` vs when it should use `m_session`. I see a way to improve that code even further with C# pattern matching.

I saw your refactoring, your code looks great. Thanks.

PR #1188, opened 2022-07-05
#375 - fixing GetSchema + Add Foreign Keys-, IndexColumns-, Indexes- …

phillipc:
#375 - fixing GetSchema + Add Foreign Keys-, IndexColumns-, Indexes- and Foreign Key Columns-Support
#375 - fixing GetSchema + Add restrictionValues support

### DIFF
--- a/src/MySqlConnector/Core/SchemaProvider.cs
+++ b/src/MySqlConnector/Core/SchemaProvider.cs
@@ -1,3 +1,4 @@
+using System.Text;
 using MySqlConnector.Protocol.Serialization;
 using MySqlConnector.Utilities;
 
@@ -38,24 +39,62 @@ internal sealed class SchemaProvider
 			{ "Triggers", FillTriggers },
 			{ "UserPrivileges", FillUserPrivileges },
 			{ "Views", FillViews },
+			{ "ForeignKeys", FillForeignKeys },
+			{ "Indexes", FillIndexes },
+			{ "IndexColumns", FillIndexes },
 		};
 	}
 
-	public ValueTask<DataTable> GetSchemaAsync(IOBehavior ioBehavior, CancellationToken cancellationToken) => GetSchemaAsync(ioBehavior, "MetaDataCollections", cancellationToken);
+	private async Task FillForeignKeys(IOBehavior ioBehavior, DataTable dataTable, string?[] restrictionValues, CancellationToken cancellationToken)
+	{
+		dataTable.Columns.AddRange(new DataColumn[]
+		{
+			new("CONSTRAINT_CATALOG", typeof(string)),
+			new("CONSTRAINT_SCHEMA", typeof(string)),
+			new("CONSTRAINT_NAME", typeof(string)),
+			new("TABLE_CATALOG", typeof(string)),
+			new("TABLE_SCHEMA", typeof(string)),
+			new("TABLE_NAME", typeof(string)),
+			new("COLUMN_NAME", typeof(string)),
+		});
 
-	public async ValueTask<DataTable> GetSchemaAsync(IOBehavior ioBehavior, string collectionName, CancellationToken cancellationToken)
+		await FillDataTableAsync(ioBehavior, dataTable, "KEY_COLUMN_USAGE", restrictionValues, cancellationToken).ConfigureAwait(false);
+	}
+
+	private async Task FillIndexes(IOBehavior ioBehavior, DataTable dataTable, string?[] restrictionValues, CancellationToken cancellationToken)
+	{
+		dataTable.Columns.AddRange(new DataColumn[]
+		{
+			new("TABLE_CATALOG", typeof(string)),
+			new("TABLE_SCHEMA", typeof(string)),
+			new("TABLE_NAME", typeof(string)),
+			new("NON_UNIQUE", typeof(string)),
+			new("INDEX_SCHEMA", typeof(string)),
+			new("INDEX_NAME", typeof(string)),
+			new("COLUMN_NAME", typeof(string)),
+			new("COLLATION", typeof(string)),
+			new("CARDINALITY", typeof(string)),
+		});
+
+		await FillDataTableAsync(ioBehavior, dataTable, "STATISTICS", restrictionValues, cancellationToken).ConfigureAwait(false);
+	}
+
+	public ValueTask<DataTable> GetSchemaAsync(IOBehavior ioBehavior, string?[] restrictionValues, CancellationToken cancellationToken) => GetSchemaAsync(ioBehavior, "MetaDataCollections", restrictionValues, cancellationToken);
+
+	public async ValueTask<DataTable> GetSchemaAsync(IOBehavior ioBehavior, string collectionName, string?[] restrictionValues, CancellationToken cancellationToken)
 	{
 		if (collectionName is null)
 			throw new ArgumentNullException(nameof(collectionName));
+		collectionName = collectionName.Replace(" ", "");
 		if (!m_schemaCollections.TryGetValue(collectionName, out var fillAction))
 			throw new ArgumentException("Invalid collection name.", nameof(collectionName));
 
 		var dataTable = new DataTable(collectionName);
-		await fillAction(ioBehavior, dataTable, cancellationToken).ConfigureAwait(false);
+		await fillAction(ioBehavior, dataTable, restrictionValues, cancellationToken).ConfigureAwait(false);
 		return dataTable;
 	}
 
-	private Task FillDataSourceInformation(IOBehavior ioBehavior, DataTable dataTable, CancellationToken cancellationToken)
+	private Task FillDataSourceInformation(IOBehavior ioBehavior, DataTable dataTable, string?[] restrictionValues, CancellationToken cancellationToken)
 	{
 		dataTable.Columns.AddRange(new DataColumn[] {
 			new("CompositeIdentifierSeparatorPattern", typeof(string)),
@@ -106,7 +145,7 @@ internal sealed class SchemaProvider
 		static string GetVersion(Version v) => "{0:00}.{1:00}.{2:0000}".FormatInvariant(v.Major, v.Minor, v.Build);
 	}
 
-	private Task FillMetadataCollections(IOBehavior ioBehavior, DataTable dataTable, CancellationToken cancellationToken)
+	private Task FillMetadataCollections(IOBehavior ioBehavior, DataTable dataTable, string?[] restrictionValues, CancellationToken cancellationToken)
 	{
 		dataTable.Columns.AddRange(new DataColumn[] {
 			new("CollectionName", typeof(string)),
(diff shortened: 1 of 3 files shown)


src/MySqlConnector/Core/SchemaProvider.cs
```diff
@@ -987,7 +1032,34 @@ private async Task FillDataTableAsync(IOBehavior ioBehavior, DataTable dataTable
...
+				+ string.Join(", ", dataTable.Columns.Cast<DataColumn>().Select(static x => x!.ColumnName))
+				+ " FROM INFORMATION_SCHEMA." + tableName;
+
+			if (restrictionValues.Any())
+			{
+				StringBuilder sbWhere = new StringBuilder();
+				List<string> whereParts = new List<string>(4);
+
+				if (restrictionValues.Length >= 1 && !string.IsNullOrEmpty(restrictionValues[0]))
+					whereParts.Add($"UPPER(TABLE_CATALOG) = '{restrictionValues[0]?.ToUpperInvariant()}'");
```
bgrainger, line 1045: * This doesn't handle a literal `'` in the restriction value (and will generate a SQL parsing error or, worse, a SQL injection bug).
  * It's possible this is bug-compatible with MySql.Data (I haven't checked), but we shouldn't allow security vulnerabilities in order to do so
* Why `UPPER`? This will fail for case-sensitive identifiers, i.e., `lower_case_table_names = 0`.
* The first restriction is not always `TABLE_CATALOG`; the set of restrictions varies according to the schema being retrieved. For example: for `"Users"`, `restrictions[0]` is a pattern to match against the user name; for `"Procedures"`, `restrictions[1]` is `ROUTINE_SCHEMA`.
  * This appears to be very poorly documented for MySql.Data, but I'm wary of being gratuitously different here; at a minimum, we should provide a porting guide mapping MySql.Data restriction indexes onto MySqlConnector restriction indexes (if there's a good reason to be different).
* `restrictionValues[0]?` is unnecessary if it's already been tested with `string.IsNullOrEmpty`.

src/MySqlConnector/Core/SchemaProvider.cs
```diff
@@ -38,24 +39,62 @@ public SchemaProvider(MySqlConnection connection)
...
+		await FillDataTableAsync(ioBehavior, dataTable, "STATISTICS", restrictionValues, cancellationToken).ConfigureAwait(false);
+	}
+
+	public ValueTask<DataTable> GetSchemaAsync(IOBehavior ioBehavior, string?[] restrictionValues, CancellationToken cancellationToken) => GetSchemaAsync(ioBehavior, "MetaDataCollections", restrictionValues, cancellationToken);
+
+	public async ValueTask<DataTable> GetSchemaAsync(IOBehavior ioBehavior, string collectionName, string?[] restrictionValues, CancellationToken cancellationToken)
 	{
 		if (collectionName is null)
 			throw new ArgumentNullException(nameof(collectionName));
+		collectionName = collectionName.Replace(" ", "");
```
bgrainger, line 88: What motivates this?

src/MySqlConnector/Core/SchemaProvider.cs
```diff
@@ -106,7 +145,7 @@ private Task FillDataSourceInformation(IOBehavior ioBehavior, DataTable dataTabl
 		static string GetVersion(Version v) => "{0:00}.{1:00}.{2:0000}".FormatInvariant(v.Major, v.Minor, v.Build);
 	}
 
-	private Task FillMetadataCollections(IOBehavior ioBehavior, DataTable dataTable, CancellationToken cancellationToken)
+	private Task FillMetadataCollections(IOBehavior ioBehavior, DataTable dataTable, string?[] restrictionValues, CancellationToken cancellationToken)
```
bgrainger, line 148: This method need to be updated to return the `NumberOfRestrictions` for each schema collection; it's currently hard-coded to 0.

src/MySqlConnector/Core/SchemaProvider.cs
phillipc: - Literal `'` : correct, but i think this isn't the correct layer for injection-prevention
- Upper: case-sensitive database configuration and duplicated-name-using are insane. If we remove upper then we are always case-sensitive.
- The restrictions-mapping is a kind of magic. I implemented it only for some known "`GetSchema()`"-calls. However, these are already sufficient for using NHibernate's schema generator. I will comment this in the code.
- restrictionValues[0]? -> "`string?`" can be `STRING.EMPTY` or not?

src/MySqlConnector/Core/SchemaProvider.cs
phillipc: `Foreign Keys or ForeignKeys`: NHibernate call this Methode with "Foreign Keys". I'm not sure which is the correct magic string variant.

src/MySqlConnector/Core/SchemaProvider.cs
phillipc: here is a docu https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/schema-restrictions of the restrictions-array

phillipc: - lower_case_table_names: In the future, we should read the server configurations such as lower_case_table_names or max_allowed_packet in order to dynamically adjust the connector. See https://github.com/mysql-net/MySqlConnector/blob/683ccf2c932255cab15e98b0522f9818e55a435c/src/MySqlConnector/Core/CommandExecutor.cs#L65

src/MySqlConnector/Core/SchemaProvider.cs
phillipc: i changed it

bgrainger:
Thanks for opening this PR and getting the ball rolling.

I fixed the spurious failures in "Linux Tests MySQL 5.7" that were (incorrectly) flagging this PR as broken: https://github.com/mysql-net/MySqlConnector/commit/bf06ea96c1bbaf7b17baf15960321619607bc334.

However, I'd prefer to move in a more data-driven direction for defining the schema collection, which should ultimately allow documentation and tests to be generated, as well as the implementation code. https://github.com/mysql-net/MySqlConnector/pull/1190

src/MySqlConnector/Core/SchemaProvider.cs
```diff
@@ -38,24 +39,99 @@ public SchemaProvider(MySqlConnection connection)
...
+		{
+			new("TABLE_CATALOG", typeof(string)),
+			new("TABLE_SCHEMA", typeof(string)),
+			new("TABLE_NAME", typeof(string)),
+			new("NON_UNIQUE", typeof(string)),
+			new("INDEX_SCHEMA", typeof(string)),
+			new("INDEX_NAME", typeof(string)),
+			new("COLUMN_NAME", typeof(string)),
+			new("COLLATION", typeof(string)),
+			new("CARDINALITY", typeof(string)),
```
bgrainger, line 111: MySql.Data returns the following columns from its `IndexColumns` schema:

```
INDEX_CATALOG
INDEX_SCHEMA
INDEX_NAME
TABLE_NAME
COLUMN_NAME
ORDINAL_POSITION
SORT_ORDER
```

Is there a reason to be different here?

src/MySqlConnector/Core/SchemaProvider.cs
```diff
@@ -38,24 +39,99 @@ public SchemaProvider(MySqlConnection connection)
...
+	{
+		dataTable.Columns.AddRange(new DataColumn[]
+		{
+			new("CONSTRAINT_CATALOG", typeof(string)),
+			new("CONSTRAINT_SCHEMA", typeof(string)),
+			new("CONSTRAINT_NAME", typeof(string)),
+			new("TABLE_CATALOG", typeof(string)),
+			new("TABLE_SCHEMA", typeof(string)),
+			new("TABLE_NAME", typeof(string)),
+			new("COLUMN_NAME", typeof(string)),
```
bgrainger, line 71: MySql.Data returns the following columns:

```
CONSTRAINT_CATALOG
CONSTRAINT_SCHEMA
CONSTRAINT_NAME
TABLE_CATALOG
TABLE_SCHEMA
TABLE_NAME
MATCH_OPTION
UPDATE_RULE
DELETE_RULE
referenced_table_catalog
REFERENCED_TABLE_SCHEMA
REFERENCED_TABLE_NAME
```

Should this be expanded to include those columns?

mgexo:
Hi @bgrainger whats the status of this PR? 
We recently migrated to NHibernate mysqlconnector due to .NET6 migration, only to find out that it is the only NHibernate connector that does not support "Foreign Keys" and "IndexColumns" in auto-update procedure. This is kind of a showstopper for updating our software. 

What  would be your suggested workaround until this PR is merged/finished?

bgrainger:
> whats the status of this PR?

It appears that the author closed it.

> This is kind of a showstopper for updating our software. What would be your suggested workaround until this PR is merged/finished?

If this is a showstopper for updating your software, and assuming that you have software development skills, the solution seems obvious: create a PR that meets your needs and submit it as a contribution to this freely-available open-source library.

mgexo:
@phillipc any reason why it was closed? Did you find a solution?

mgexo:
> > whats the status of this PR?
> 
> It appears that the author closed it.
> 
> > This is kind of a showstopper for updating our software. What would be your suggested workaround until this PR is merged/finished?
> 
> If this is a showstopper for updating your software, and assuming that you have software development skills, the solution seems obvious: create a PR that meets your needs and submit it as a contribution to this freely-available open-source library.

@bgrainger we actually considered that but then we saw already two people that tried and failed or were not able to get their PRs through. I am afraid that we will end up with yet another open source library fork ...
Since you reviewed the PR here already, isnt there a chance to finish that?

bgrainger:
This project has a high bar for contributions (see the [Code of Conduct](https://github.com/mysql-net/MySqlConnector?tab=coc-ov-file#readme)), but it regularly merges contributions (see the [contributors list](https://github.com/mysql-net/MySqlConnector/graphs/contributors)); there's no a priori reason to assume a PR you opened wouldn't be merged.

I haven't looked into these particular schemas in detail recently.

If they're just reporting data from an `INFORMATION_SCHEMA.XYZ` table, then you can add a block to `SchemaCollections.yaml` that specifies the table name and columns, then run `SchemaCollectionGenerator` to generate the code.

However, it may not be that simple or I might have just added it in #1190. If it's more complex, then it might be best to follow the example of `ReservedWords`: define the schema in `SchemaCollections.yaml` but set a `custom:` method name, then implement that method in `SchemaProvider.cs`.

In either case, don't modified the generated code in `SchemaProvider.g.cs` _except_ by modifying and re-running `SchemaCollectionGenerator`.

eaexo:
Hi @bgrainger 

I guess I have a using SchemaCollections.yaml, I will try to create PR later, however I agree with mgexo that it could end up as 2 previous PRs

PR #1193, opened 2022-07-18
Remove DATETIME_PRECISION as it's not present in MySQL 5.5

### LINKED ISSUE #1192: MySQL 5.5 does not have DATETIME_PRECISION
**Software versions**
MySqlConnector version:
Server type (MySQL, MariaDB, Aurora, etc.) and version: MySQL 5.5
.NET version: 6.0

**Describe the bug**
When using the SchemaProvider to get the columns within the FillColumns the requested data for the DATETIME_PRECISION is simply not present for MySQL 5.5 servers.

**Exception**
image

malessainray:
Fixes #1192


bgrainger:
Please wait until https://github.com/mysql-net/MySqlConnector/pull/1190 is merged in, as it's going to replace all this code.

I also disagree with removing the column from later MySQL versions, just because an unsupported version of MySQL (5.5) doesn't have it. We need to find a better way to test for it at runtime.

tests/SideBySide/SchemaProviderTests.cs
```diff
@@ -49,7 +49,6 @@ public void ColumnsSchema()
 		AssertHasColumn("CHARACTER_MAXIMUM_LENGTH", typeof(long));
 		AssertHasColumn("NUMERIC_PRECISION", typeof(ulong));
 		AssertHasColumn("NUMERIC_SCALE", typeof(ulong));
-		AssertHasColumn("DATETIME_PRECISION", typeof(uint));
```
bgrainger, line 52: This code exists to verify that changes haven't introduced a regression; you can't just delete it to make the tests pass.

bgrainger (changes requested):
Don't delete existing tests.

PR #1197, opened 2022-07-31
Add GetBytes method with name

CorentinPtrl:
Just add a GetBytes method without ordinal


bgrainger:
(I've fixed the spurious build error.)

CorentinPtrl:
Yeah i agree with the DCO, but i just thought it could simplify the use of GetBytes since every other getter has his overload

bgrainger:
Added in [2.1.13](https://github.com/mysql-net/MySqlConnector/releases/tag/2.1.13).

PR #1206, opened 2022-08-18
Avoid culture aware comparisons and a string allocation

iamcarbon:
The standard string == operator is a culture aware comparison. Updating these to "is" makes these culture invariant.

This PR removes the few instances of these.

It also removes a single string allocation when checking for HasDelimiter. There are also a few other opportunities to avoid string allocations if we raise our minimum target to NET461 (ideally NET462).

### DIFF
--- a/src/MySqlConnector/Core/ConnectionSettings.cs
+++ b/src/MySqlConnector/Core/ConnectionSettings.cs
@@ -27,7 +27,7 @@ internal sealed class ConnectionSettings
 			if (csb.LoadBalance != MySqlLoadBalance.RoundRobin)
 				throw new NotSupportedException("LoadBalance not supported when ConnectionProtocol=NamedPipe");
 			ConnectionProtocol = MySqlConnectionProtocol.NamedPipe;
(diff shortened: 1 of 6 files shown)


bgrainger:
> The standard string == operator is a culture aware comparison.

I'm extremely skeptical of this claim; do you have proof?

bgrainger:
[`==` and `is`](https://sharplab.io/#v2:EYLgtghglgdgNAFxFANgHwAICYCMBYAKAwGYACbUgYVIG9DSHyzgB7FlUgWRwAoMcADKQDOASlr1GUjAHYRpALwLSAIhYwApioDckhgF89TUq3ZcsfQSPF0CU6XOGkoTlQgDuLHUcMF9QA==) look identical to me.

bgrainger:
> It also removes a single string allocation when checking for HasDelimiter. 

Thanks. Can you reduce the PR down to just this one change?

iamcarbon:
@bgrainger Totally right, == and is operates are both do ordinal (case-sensitive and culture-insensitive) comparisons! Thanks for double checking the source generation too -- I've updated my knowledge!

I've posted a single clean PR for the single allocation reduction.

PR #1252, opened 2022-12-10
Avoid normalize parameter every operation in prepared query

neuecc:
If using prepared query,
`SingleCommandPayloadCreator.WritePreparedStatement` calls `parameterCollection?.NormalizedIndexOf` and it allocates string(Trim `@`).
However, since the Parameter of a Prepared Statement is always the same, if it is Normalized in advance, it will not be Allocated each time.

This PR adds `List<string?> NormalizedParameterNames` to `ParsedStatement` (`ParsedStatement` is only used prepared query).
Also added MySqlParameterCollection.UnsafeIndexOf, which is an IndexOf that expects the argument to be already normalized.

This improves the performance of Prepared Queries that are executed multiple times.

### DIFF
--- a/src/MySqlConnector/Core/ParsedStatement.cs
+++ b/src/MySqlConnector/Core/ParsedStatement.cs
@@ -18,6 +18,12 @@ internal sealed class ParsedStatement
 	public List<string?> ParameterNames { get; } = new();
 
 	/// <summary>
+	/// The normalized names of the parameters (if known) of the parameters in the prepared statement. There
+	/// is one entry in this list for each parameter, which will be <c>null</c> if the name is unknown.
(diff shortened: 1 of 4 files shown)


bgrainger (approved):
Do you affirm the Developer Certificate of Origin for this contribution? https://developercertificate.org/

neuecc:
> Do you affirm the Developer Certificate of Origin for this contribution? https://developercertificate.org/

yes, and sorry for missed DCO.

bgrainger:
Shipped in [2.2.3](https://github.com/mysql-net/MySqlConnector/releases/tag/2.2.3); thanks for the contribution!

PR #1253, opened 2022-12-10
Remove `FormatInvariant` helper method

bgrainger:
Many of these usages are in creating exception messages, so this won't improve runtime performance much. However, there are a few usages when serialising parameter values.


bgrainger:
Removed in [2.2.3](https://github.com/mysql-net/MySqlConnector/releases/tag/2.2.3).

PR #1269, opened 2022-12-28
Make MySqlDataSource available in all builds of the library

bgrainger:
This will be the recommended way to configure logging and other settings.


src/MySqlConnector/Shims/DbDataSource.cs
```diff
@@ -0,0 +1,93 @@
+#if !NET7_0_OR_GREATER
+namespace System.Data.Common;
```
ejball, line 2: This wouldn't play nice with another package with the same shim, but maybe that's not much a risk?

bgrainger: The obvious example is Npgsql: https://github.com/npgsql/npgsql/blob/main/src/Npgsql/Shims/DbDataSource.cs

MySqlConnector has had a `DbColumn` shim for a long time, too, and no one has reported any problems: https://github.com/mysql-net/MySqlConnector/blob/master/src/MySqlConnector/Shims/DbColumn.cs

It would only be a problem if someone were trying to reference this type, right? Which seems very unlikely.

ejball: > It would only be a problem if someone were trying to reference this type, right?

Yeah, I think that's right.

Issue #446, opened 2018-03-06
Support CLIENT_OPTIONAL_RESULTSET_METADATA


bgrainger:
A new client flag in MySQL 8.0:

* https://dev.mysql.com/doc/refman/8.0/en/c-api-optional-metadata.html
* https://dev.mysql.com/worklog/task/?id=8134

Would need to be controlled by a new connection string option (`OmitResultSetMetadata=true`?).

Required for aspnet/DataAccessPerformance#34.

bgrainger:
The client can control this on a per-result-set basis by executing `SET @@resultset_metadata=NONE|FULL;`; a connection string option may be unnecessary?

PR #1301, opened 2023-03-31
Permit skipping metadata for MariaDB

rusher:
Since MariaDB 10.6 (with https://jira.mariadb.org/browse/MDEV-19237) binary result-set can skip sending metadata when they haven't changed.
This avoid network data and parsing client side.

Benchmarks results using DDL:
```
CREATE TABLE test100 (i1 int,i2 int,...,i100 int);
INSERT INTO test100 value (1,2,...,100);
```

test methods:
```
    [Benchmark]
    public async Task<int> readOneRow100FieldsAsync()
    {
        var total = 0;
        {
            cmd.CommandText = "select * FROM test100";
            {
                    for (var i = 0; i < 100; i++)
                        total += reader.GetInt32(i);
            }
        }
        return total;
    }

    [Benchmark]
    public async Task<int> readOneRow100FieldsPrepareAsync()
    {
        var total = 0;
        {
            cmd.CommandText = "select * FROM test100";
            cmd.Prepare();
            {
                    for (var i = 0; i < 100; i++)
                        total += reader.GetInt32(i);
            }
        }
        return total;
    }
```

before PR:
```
|                          Method |        Library |      Mean |    Error |
|-------------------------------- |--------------- |----------:|---------:|
|        readOneRow100FieldsAsync | MySqlConnector | 104.47 us | 1.468 us |
| readOneRow100FieldsPrepareAsync | MySqlConnector |  92.49 us | 1.154 us |
```
After PR:
```
|                          Method |        Library |      Mean |    Error |
|-------------------------------- |--------------- |----------:|---------:|
|        readOneRow100FieldsAsync | MySqlConnector | 103.99 us | 0.914 us |
| readOneRow100FieldsPrepareAsync | MySqlConnector |  71.50 us | 1.402 us |
```

### DIFF
--- a/src/MySqlConnector/Core/CommandListPosition.cs
+++ b/src/MySqlConnector/Core/CommandListPosition.cs
@@ -8,6 +8,7 @@ internal struct CommandListPosition
 	public CommandListPosition(IReadOnlyList<IMySqlCommand> commands)
 	{
 		Commands = commands;
+		PreparedStatements = null;
 		CommandIndex = 0;
 		PreparedStatementIndex = 0;
 	}
@@ -18,6 +19,11 @@ internal struct CommandListPosition
 	public IReadOnlyList<IMySqlCommand> Commands { get; }
 
 	/// <summary>
+	/// Associated prepared statements of commands
+	/// </summary>
+	public PreparedStatements? PreparedStatements;
+
+	/// <summary>
 	/// The index of the current command.
 	/// </summary>
 	public int CommandIndex;
@@ -26,4 +32,9 @@ internal struct CommandListPosition
 	/// If the current command is a prepared statement, the index of the current prepared statement for that command.
 	/// </summary>
 	public int PreparedStatementIndex;
+
+	/// <summary>
+	/// Retrieve the last used prepared statement
+	/// </summary>
+	public PreparedStatement? LastUsedPreparedStatement;
 }
--- a/src/MySqlConnector/Core/ResultSet.cs
+++ b/src/MySqlConnector/Core/ResultSet.cs
@@ -116,42 +116,55 @@ internal sealed class ResultSet
 				}
 				else
 				{
-					static int ReadColumnCount(ReadOnlySpan<byte> span)
+					var columnCountPacket = ColumnCountPayload.Create(payload.Span, Session.SupportsMetaSkip);
+					if (!columnCountPacket.MetadataFollows)
 					{
-						var reader = new ByteArrayReader(span);
-						var columnCount_ = (int) reader.ReadLengthEncodedInteger();
-						if (reader.BytesRemaining != 0)
-							throw new MySqlException("Unexpected data at end of column_count packet; see https://github.com/mysql-net/MySqlConnector/issues/324");
-						return columnCount_;
+						// reuse previous metadata
+						var preparedStatement = DataReader.CurrentPrepareStmt()!;
+						ColumnDefinitions = preparedStatement.Columns!;
+						ColumnTypes = new MySqlDbType[columnCountPacket.ColumnCount];
+						for (var column = 0; column < columnCountPacket.ColumnCount; column++)
+						{
+							ColumnTypes[column] = TypeMapper.ConvertToMySqlDbType(ColumnDefinitions[column],
+								Connection.TreatTinyAsBoolean, Connection.GuidFormat);
+						}
 					}
-					var columnCount = ReadColumnCount(payload.Span);
-
-					// reserve adequate space to hold a copy of all column definitions (but note that this can be resized below if we guess too small)
-					Utility.Resize(ref m_columnDefinitionPayloads, columnCount * 96);
-
-					ColumnDefinitions = new ColumnDefinitionPayload[columnCount];
-					ColumnTypes = new MySqlDbType[columnCount];
-
-					for (var column = 0; column < ColumnDefinitions.Length; column++)
+					else
 					{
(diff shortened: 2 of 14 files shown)


bgrainger:
Thanks! I will take a look when I'm back from vacation.

I assume this may be related to https://github.com/mysql-net/MySqlConnector/issues/446 although MariaDB uses a different flag value to enable the capability? (Or maybe it's a completely different on-the-wire way to avoid sending metadata?)

rusher:
In term of protocol, that correspond to the mysql change you indicate. 
Still this is not the exact same thing : 
* for MySQL there a capability CLIENT_OPTIONAL_RESULTSET_METADATA that permit to enable this functionality, then when @@session.resultset_metadata is "NONE", all result-set metadata won't then be sent.
* for MariaDB, when client set capability [MARIADB_CLIENT_CACHE_METADATA](https://mariadb.com/kb/en/connection/#capabilities), metadatas *MIGHT* not be send. We don't want the client to remember all metadata for each type of query. When MARIADB_CLIENT_CACHE_METADATA is set, server will not resent metadata to COM_STMT_EXECUTE result-set if metadata has not changed. So to resume, first execute result will have metadata, second will not.

src/MySqlConnector/Core/CommandListPosition.cs
```diff
@@ -17,6 +18,11 @@ public CommandListPosition(IReadOnlyList<IMySqlCommand> commands)
 	/// </summary>
 	public IReadOnlyList<IMySqlCommand> Commands { get; }
 
+	/// <summary>
+	/// Associated prepared statements of commands
+	/// </summary>
+	public PreparedStatements? PreparedStatements;
```
bgrainger, line 24: Note to self: this struct is starting to accumulate a lot of mutable public fields; redesign?

src/MySqlConnector/Core/ResultSet.cs
```diff
@@ -116,42 +116,55 @@ public async Task ReadResultSetHeaderAsync(IOBehavior ioBehavior)
...
-						return columnCount_;
+						// reuse previous metadata
+						var preparedStatement = DataReader.CurrentPrepareStmt()!;
+						ColumnDefinitions = preparedStatement.Columns!;
+						ColumnTypes = new MySqlDbType[columnCountPacket.ColumnCount];
+						for (var column = 0; column < columnCountPacket.ColumnCount; column++)
+						{
+							ColumnTypes[column] = TypeMapper.ConvertToMySqlDbType(ColumnDefinitions[column],
+								Connection.TreatTinyAsBoolean, Connection.GuidFormat);
+						}
```
bgrainger, line 130: Note to self: Can we move this outside the `if` statement (and remove the corresponding code from the `else` branch)?

src/MySqlConnector/Protocol/Payloads/ColumnCountPayload.cs
```diff
@@ -0,0 +1,28 @@
+using MySqlConnector.Protocol.Serialization;
+
+namespace MySqlConnector.Protocol.Payloads;
+
+/// <summary>
+///     Helper class to parse Column count packet.
+///     https://mariadb.com/kb/en/result-set-packets/#column-count-packet
+///     Packet contains a columnCount, and - if capability MARIADB_CLIENT_CACHE_METADATA is set - a flag to indicate if metadata follows
+/// </summary>
+internal sealed class ColumnCountPayload
```
bgrainger, line 10: Note to self: Most other `Payload` objects are `class`es, but would it be worthwhile making this (and others?) a `readonly struct`?

src/MySqlConnector/Protocol/ProtocolCapabilities.cs
```diff
@@ -147,4 +147,14 @@ internal enum ProtocolCapabilities : ulong
...
+
+	/// <summary>
+	/// metadata extended information
+	/// </summary>
+	MariaDbExtendedTypeInfo = 0x8_0000_0000,
+
+	/// <summary>
+	/// permit metadata caching
+	/// </summary>
+	MariaDbCacheMetadata = 0x16_0000_0000,
```
bgrainger, line 159:
```suggestion
	MariaDbCacheMetadata = 0x10_0000_0000,
```

I assume this should just be the single bit?

bgrainger (changes requested):
I believe the `MariaDbCacheMetadata` value is defined incorrectly and needs to be fixed.

src/MySqlConnector/Core/ResultSet.cs
rusher: I think best would be to get rid of the ColumnTypes variable, saving some calculation that are useful for metadata only, so that may not be used at all. => save ConnectionSettings and just rely on ColumnDefinitions

src/MySqlConnector/Protocol/ProtocolCapabilities.cs
rusher: corrected with latest push

bgrainger (approved):

src/MySqlConnector/Core/ResultSet.cs
bgrainger: Agreed: https://github.com/mysql-net/MySqlConnector/commit/5f72a2f741629fc5c2fa7781621315a329eb941d.

src/MySqlConnector/Protocol/Payloads/ColumnCountPayload.cs
bgrainger: https://github.com/mysql-net/MySqlConnector/commit/3e746d8af2fc7e17bee26dbbc7503f7b3570eb1f

PR #1303, opened 2023-04-06
Permit multi-authentication

rusher:
Authentication switch request can result in 3 type of responses:
* OK_Packet
* ERR_Packet
* another authentication switch request.

Actual implementation does not permit another authentication switch request.

This can be the case when using multiple authentication, like :
```
create user mysqltest1 identified via mysql_native_password as password("good") OR unix_socket;
``` 
Existing in:
* MariaDB 10.4.3+ (https://mariadb.com/kb/en/create-user/#identified-viawith-authentication_plugin) 
* MySQL 8.0.27+ multifactor authentication (MFA) (https://dev.mysql.com/doc/refman/8.0/en/create-user.html)


bgrainger (approved):
Thanks for the improvement!

Issue #1290, opened 2023-03-09
Excessive lock contention and CPU usage in TimerQueue


saintarian:
**Software versions**
MySqlConnector version: 2.2.5
Server type (MySQL, MariaDB, Aurora, etc.) and version:  Mysql 8
.NET version: 6
(Optional) ORM NuGet packages and versions: 

**Describe the bug**
With frequent reads of many rows from many concurrent async "threads", profiling shows excessive lock contention and CPU usage in TimerQueue.Add/Remove. This has led to thread-blocking issues for us, causing delays in async scheduling. 

**Exception**
Full exception message and call stack (if applicable)
See repro steps

**Code sample**
* Clone https://github.com/saintarian/mysqlconnector-timerqueue-repro
* Run `docker compose build`
* Run `docker compose up`
* Let it run for ~15m
* Navigate to [http://localhost:4040](http://localhost:4040) to see profiling results. Choose `db-load-generator.mutex_duration` and `db-load-generator.cpu` in the dropdown at the top. 

```csharp
/* A concise code sample to reproduce the bug */
```

**Expected behavior**
Less lock contention and CPU usage. 

**Additional context**
TimerQueue.Remove+Add happens in 3 places as far as I can tell:
* MySqlCommand.cs: SetTImeout seems to be called each time a row is read, causing lots of TimerQueue churn. 
* SocketByteHandler.cs: Add+Remove is done each time IO is done
* StreamByteHandler.cs: Add+Remove is done each time IO is done

saintarian:
A few observations (I'm new to the codebase so please correct me if I'm wrong):
* TimerQueue uses an ordered list to store fireable-actions in fire-time order. Insertion is O(n) (but likely less because most additions go to the end) and removal is O(n). 
* When reading n rows, I believe it results in n Remove+Adds in MySqlCommand.cs. 
* When using DbDataReader.ReadAsync() to iterate through n rows, I believe it results in n Remove+Adds in Socket/StreamByteHandler.cs. 

Some thoughts on possible fixes (again from a total newbie):
* The vast majority of TimerQueue additions won't fire before they are removed. 
* The smaller majority of TimerQueue additions won't get to their initial fire time (before the timeout is reset). This is an opportunity to reduce TimerQueue churn by letting it fire at the initial fire time, and at that point re-enqueueing for the updated fire-time. 
[...]

PR #1304, opened 2023-04-07
Query timeout using per query variables

rusher:
MariaDB and percona server permits per query variables
(see https://mariadb.com/kb/en/set-statement/ and https://www.percona.com/blog/using-per-query-variable-statements-in-percona-server/)

This corresponds to commands like
```
SET STATEMENT <variable=value> FOR <statement>
```
These 2 servers have a `max_statement_time` variable that allows the abort of requests taking longer than this time.

This PR set session `max_statement_time` to avoid using a TimerQueue for each execution, and if reaching timeout, using `KILL QUERY` commands on another connection. This just leave the server handling the timeout.

The server will then return an error "The execution of the request was interrupted (max_statement_time exceeded)".

This functionality is available in MySQL since version 5.7.4, renamed max_execution_time but is limited to read-only SELECT queries and not in stored procedures, which makes it unusable for connectors.

this also allows CommandBehavior.SchemaOnly and CommandBehavior.SingleRow to not use multiple statements, which helps improve performance.

Benchmark results using default command timeout and a MariaDB server:

```

    [Benchmark]
    public async Task<bool> ReadSingleRow()
    {
	    {
		    cmd.CommandText = "SELECT * FROM seq_10_to_20";
		    using (var reader = await cmd.ExecuteReaderAsync(CommandBehavior.SingleRow))
		    {
			    await reader.ReadAsync();
			    reader.GetInt32(0);
			    return await reader.ReadAsync();
		    }
	    }
    }

    [Benchmark]
    public async Task<System.Data.DataColumnCollection> ReadSchemaOnly()
    {
	    {
		    cmd.CommandText = "SELECT * FROM seq_10_to_20";
		    using (var reader = await cmd.ExecuteReaderAsync(CommandBehavior.SchemaOnly))
		    {
			    await reader.ReadAsync();
			    return reader.GetSchemaTable().Columns;
		    }
	    }
    }

    [Benchmark]
    public async Task CommandTimeout()
    {
	    {
		    cmd.CommandText = "SELECT 1";
		    // cmd.CommandTimeout = 1; //using default timeout
		    {
				reader.Read();
		    }
	    }
    }
```

results :
```
|         Method |        Library |     Mean |    Error |
|--------------- |--------------- |---------:|---------:|
|  ReadSingleRow | MySqlConnector | 60.69 us | 0.367 us |
| ReadSchemaOnly | MySqlConnector | 64.02 us | 1.238 us |
| CommandTimeout | MySqlConnector | 34.53 us | 0.606 us |
```
after PR:
```
|         Method |        Library |     Mean |    Error |
|--------------- |--------------- |---------:|---------:|
|  ReadSingleRow | MySqlConnector | 49.73 us | 0.222 us |
| ReadSchemaOnly | MySqlConnector | 57.26 us | 1.032 us |
| CommandTimeout | MySqlConnector | 34.25 us | 0.279 us |
```

Command using default timeout result are similar, the difference being minimal, but running the benchmark 15x, performance is always around 1% better.

In order to run test, i've slightly change Cancel and Timeout test, because of MySQL and MariaDB difference, killing a `SLEEP(x)` command resulting in a OK_Packet in MySQL, throwing an "Query execution was interrupted" exception for MariaDB

### DIFF
--- a/src/MySqlConnector/Core/ServerSession.cs
+++ b/src/MySqlConnector/Core/ServerSession.cs
@@ -48,6 +48,7 @@ internal sealed partial class ServerSession
 
 	public string Id { get; }
 	public ServerVersion ServerVersion { get; set; }
+	public bool SupportPerQueryVariables => ServerVersion.MariaDbVersion is not null && ServerVersion.MariaDbVersion >= ServerVersions.MariaDbSupportPerQueryVariables;
 	public int ActiveCommandId { get; private set; }
 	public int CancellationTimeout { get; private set; }
 	public int ConnectionId { get; set; }
@@ -195,7 +196,11 @@ internal sealed partial class ServerSession
 		}
 		else
 		{
-			commandToPrepare = commandText;
+			commandToPrepare = SupportPerQueryVariables
+			                   && command.CommandTimeout > 0
+			                   && command.CommandTimeout != command.Connection!.DefaultCommandTimeout
+				? "SET STATEMENT max_statement_time=" + command.CommandTimeout + " FOR " + commandText
+				: commandText;
 		}
 
 		var statementPreparer = new StatementPreparer(commandToPrepare, command.RawParameters, command.CreateStatementPreparerOptions());
@@ -572,13 +577,8 @@ internal sealed partial class ServerSession
 			if (m_useCompression)
 				m_payloadHandler = new CompressedPayloadHandler(m_payloadHandler.ByteHandler);
 
-			// set 'collation_connection' to the server default
-			await SendAsync(m_setNamesPayload, ioBehavior, cancellationToken).ConfigureAwait(false);
-			payload = await ReceiveReplyAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
-			OkPayload.Create(payload.Span, SupportsDeprecateEof, SupportsSessionTrack);
-
-			if (ShouldGetRealServerDetails(cs))
-				await GetRealServerDetailsAsync(ioBehavior, CancellationToken.None).ConfigureAwait(false);
+			// execute Post Connection commands
+			await PostConnectionsQueriesAsync(cs, ioBehavior, cancellationToken).ConfigureAwait(false);
 
 			m_payloadHandler.ByteHandler.RemainingTimeout = Constants.InfiniteTimeout;
 		}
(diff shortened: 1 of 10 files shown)


rusher:
Another solution would be to only set session max_statement_time. this would remove use of TimerQueue for all queries that have  the standard timeout.

src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -596,6 +596,63 @@ public async Task DisposeAsync(IOBehavior ioBehavior, CancellationToken cancella
 		return statusInfo;
 	}
 
+	public async Task PostConnectionsQueriesAsync(ConnectionSettings cs,
+		IOBehavior ioBehavior, CancellationToken cancellationToken)
+	{
+		var payloadDatas = new List<PayloadData>();
```
bgrainger, line 602: The extra allocations here (on a common path) initially concerned me, but I suppose the benefit of pipelining should outweigh the cost of any extra garbage.

src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -596,6 +596,63 @@ public async Task DisposeAsync(IOBehavior ioBehavior, CancellationToken cancella
...
+				await ReadRealServerDetailsAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
+			});
+		}
+
+		var querySendTasks = new List<Task>();
+		if (m_supportsPipelining)
+		{
+			foreach (var payloadData in payloadDatas)
+			{
+				querySendTasks.Add(SendAsync(payloadData, ioBehavior, cancellationToken).AsTask());
```
bgrainger, line 637: I don't know if my code makes the guarantee that `SendAsync` can be called in parallel (e.g., if a payload ends up getting broken up across multiple `Socket` calls). This may be unsafe.

tests/IntegrationTests/CommandTimeoutTests.cs
```diff
@@ -72,13 +77,16 @@ public void CommandTimeoutWithSleepSync()
...
+	[SkippableTheory(ServerFeatures.CancelSleepSuccessfully | ServerFeatures.Timeout)]
+	[InlineData(true)]
+	[InlineData(false)]
+	public async Task CommandTimeoutWithSleepAsync(bool prepare)
 	{
 		var connectionState = m_connection.State;
 		using (var cmd = new MySqlCommand("SELECT SLEEP(120);", m_connection))
 		{
 			cmd.CommandTimeout = 2;
+			if (prepare) cmd.Prepare();
```
bgrainger, line 89:
```suggestion
			if (prepare) await cmd.PrepareAsync();
```

Seems reasonable to test the async code path in this test method.

rusher:
Corrections after remarks: 
* use 'Supports' in place of 'Support', following other variables naming
* correcting post-queries pipelining, that clearly would cause race condition

bgrainger:
I only just realised that this will significantly change the meaning/behaviour of `MySqlCommand.CommandTimeout`. It's currently implemented to follow the documentation of [SqlCommand.CommandTimeout](https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlcommand.commandtimeout):

> This property is the cumulative time-out (for all network packets that are read during the invocation of a method) for all network reads during command execution or processing of the results. A time-out can still occur after the first row is returned, and does not include user processing time, only network read time.

> 
> For example, with a 30 second time out, if **Read** requires two network packets, then it has 30 seconds to read both network packets. If you call **Read** again, it will have another 30 seconds to read any data that it requires.

This would (if I understand correctly) change the CommandTimeout from per-call-to-Read to per-Execute(NonQuery/Scalar) or per result set returned by ExecuteReader.

If I were to accept this PR, I think the meaning of `CommandTimeout` for servers that don't support `max_statement_time` would need to be adjusted to match. (That might also resolve https://github.com/mysql-net/MySqlConnector/issues/1290, which is primarily caused by the need to reset the timeout for each network packet.) This could, however, be a significant breaking change. Queries that take longer than thirty seconds to return the entire result set would now start getting cancelled while data is coming in over the wire. (Previously, it would only be cancelled if no network packets were received for 30 seconds.)

That's not to say that this is a "bad" change, per se; your understanding of what `CommandTimeout` means may match most people's assumptions, and many current users might be surprised to learn that `CommandTimeout` _doesn't_ enforce an upper limit on query execution time. (But `ExecuteXAsync(tokenThatIsCancelledAfterADelay)` does.)

PR #1313, opened 2023-04-24
Mark some MySql.Data issues as resolved

MaceWindu:
Unexpected, but Oracle added async implementation. But as expected, not everything works properly - cancellation token is not respected.

Also following issues are marked as fixed, but I will leave it to you to decide as I didn't tested it:
- https://bugs.mysql.com/bug.php?id=92465
- https://bugs.mysql.com/bug.php?id=93370
- https://bugs.mysql.com/bug.php?id=97061
- https://bugs.mysql.com/bug.php?id=99091
- https://bugs.mysql.com/bug.php?id=106247
- https://bugs.mysql.com/bug.php?id=109683

And this one now is hidden:
- https://bugs.mysql.com/bug.php?id=97473


bgrainger:
Thanks for the list.

I couldn't run my integration tests to verify that they're fixed, because the tests immediately crash due to a new bug: https://bugs.mysql.com/bug.php?id=110717.

MaceWindu:
I think you can disable pooling in connection string as workaround

bgrainger:
Marked all those issues as resolved, and filed three more.

PR #1314, opened 2023-04-28
Permit real prepared stored procedure

rusher:
When using BINARY protocol, output parameters are sent using an additional result-set that is marked with intermediate EOF and ending EOF/OK Packet with flag [SERVER_PS_OUT_PARAMS](https://mariadb.com/kb/en/ok_packet/#server-status-flag) set (Capability PS_MULTI_RESULTS has been added to permit that).

This is an old functionality, working for any type of server > 5.0.

I think that using binary implementation must even be the basis : stored procedure can always be prepared, and only BINARY protocol permit to get output parameter.

This permits better performance than using multi-resultset with "SET @inParam1=xx;...; SELECT @outParam1,@outParam2".

Since implementation use streaming result-set, DEPRECATE_EOF capability is disabled, in order to identify "output parameter resultset" in intermediate EOF. It could be improved checking if next packet is a row packet and then fetching next packet to check if this is an EOF/OK_Packet with SERVER_PS_OUT_PARAMS flag set, but i didn't know how to make that properly.

 bench base on StoredProcedureCircle test, like
 ```
 		{
 			cmd.CommandText = "circle";
 			cmd.CommandType = CommandType.StoredProcedure;
 			cmd.Parameters.Add(new MySqlParameter()
 			{
 				ParameterName = "@radius",
 				DbType = DbType.Double,
 				Direction = ParameterDirection.Input,
 				Value = 1.0,
 			});
 			...
 			cmd.Parameters.Add(new MySqlParameter()
 			{
 				ParameterName = "@shape",
 				DbType = DbType.String,
 				Direction = ParameterDirection.Output,
 			});
 			{
 				if (await reader.ReadAsync())
 					reader.GetValue(0);
 			}

 		}
 ```
 show improved result:
 with 2.3.0-beta.2:
 ```
  |                 Method |        Library |     Mean |   Error |
 |----------------------- |--------------- |---------:|--------:|
 |     storeProcedureText | MySqlConnector | 120.9 us | 2.41 us |
 ```

 with PR:
 ```
 |                 Method |        Library |      Mean |    Error |
 |----------------------- |--------------- |----------:|---------:|
 |     storeProcedureText | MySqlConnector | 118.79 us | 2.311 us |
 | storeProcedurePrepared | MySqlConnector |  93.31 us | 1.021 us |
```

### DIFF
--- a/src/MySqlConnector/Core/CachedProcedure.cs
+++ b/src/MySqlConnector/Core/CachedProcedure.cs
@@ -20,9 +20,10 @@ internal sealed class CachedProcedure
 			{
 				using var cmd = connection.CreateCommand();
 				cmd.Transaction = connection.CurrentTransaction;
-				cmd.CommandText = @"SELECT param_list, returns FROM mysql.proc WHERE db = @schema AND name = @component";
-				cmd.Parameters.AddWithValue("@schema", schema);
-				cmd.Parameters.AddWithValue("@component", component);
+				cmd.CommandText = "SELECT param_list, returns FROM mysql.proc WHERE db = ? AND name = ?";
+				cmd.Prepare();
+				cmd.Parameters.Add(new() { Value = schema });
+				cmd.Parameters.Add(new() { Value = component });
 
 				using var reader = await cmd.ExecuteReaderNoResetTimeoutAsync(CommandBehavior.Default, ioBehavior, cancellationToken).ConfigureAwait(false);
 				var exists = await reader.ReadAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
@@ -66,13 +67,16 @@ internal sealed class CachedProcedure
 			cmd.Transaction = connection.CurrentTransaction;
 			cmd.CommandText = @"SELECT COUNT(*)
 				FROM information_schema.routines
-				WHERE ROUTINE_SCHEMA = @schema AND ROUTINE_NAME = @component;
+				WHERE ROUTINE_SCHEMA = ? AND ROUTINE_NAME = ?;
 				SELECT ORDINAL_POSITION, PARAMETER_MODE, PARAMETER_NAME, DTD_IDENTIFIER
 				FROM information_schema.parameters
-				WHERE SPECIFIC_SCHEMA = @schema AND SPECIFIC_NAME = @component
+				WHERE SPECIFIC_SCHEMA = ? AND SPECIFIC_NAME = ?
 				ORDER BY ORDINAL_POSITION";
-			cmd.Parameters.AddWithValue("@schema", schema);
-			cmd.Parameters.AddWithValue("@component", component);
+			cmd.Prepare();
+			cmd.Parameters.Add(new() { Value = schema });
+			cmd.Parameters.Add(new() { Value = component });
+			cmd.Parameters.Add(new() { Value = schema });
+			cmd.Parameters.Add(new() { Value = component });
 
 			using var reader = await cmd.ExecuteReaderNoResetTimeoutAsync(CommandBehavior.Default, ioBehavior, cancellationToken).ConfigureAwait(false);
 			await reader.ReadAsync(ioBehavior, cancellationToken).ConfigureAwait(false);
--- a/src/MySqlConnector/Core/ServerSession.cs
+++ b/src/MySqlConnector/Core/ServerSession.cs
@@ -152,48 +152,18 @@ internal sealed partial class ServerSession
 				var name = NormalizedSchema.MustNormalize(command.CommandText!, command.Connection.Database);
 				throw new MySqlException($"Procedure or function '{name.Component}' cannot be found in database '{name.Schema}'.");
 			}
-
+			command.CachedProc = cachedProcedure;
 			var parameterCount = cachedProcedure.Parameters.Count;
-#if NETCOREAPP2_1_OR_GREATER || NETSTANDARD2_1_OR_GREATER
-			commandToPrepare = string.Create(commandText.Length + 7 + parameterCount * 2 + (parameterCount == 0 ? 1 : 0), (commandText, parameterCount), static (buffer, state) =>
+			commandToPrepare = StoredProcedureUtils.Format(commandText, parameterCount);
+
+			// if already in cache, just returns immediately
+			var cached = TryGetPreparedStatement(commandToPrepare);
+
+			if (cached != null)
 			{
-				buffer[0] = 'C';
-				buffer[1] = 'A';
-				buffer[2] = 'L';
-				buffer[3] = 'L';
-				buffer[4] = ' ';
-				buffer = buffer[5..];
-				state.commandText.AsSpan().CopyTo(buffer);
-				buffer = buffer[state.commandText.Length..];
-				buffer[0] = '(';
-				buffer = buffer[1..];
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: src/MySqlConnector/Protocol/Payloads/HandshakeResponse41Payload.cs [hunk 1]
```diff
@@ -5,11 +5,8 @@ namespace MySqlConnector.Protocol.Payloads;
 
 internal static class HandshakeResponse41Payload
 {
-	private static ByteBufferWriter CreateCapabilitiesPayload(ProtocolCapabilities serverCapabilities, ConnectionSettings cs, bool useCompression, CharacterSet characterSet, ProtocolCapabilities additionalCapabilities = 0)
-	{
-		var writer = new ByteBufferWriter();
-
-		var clientCapabilities = (ProtocolCapabilities.Protocol41 |
+	public static ProtocolCapabilities CreateClientCapabilities(ProtocolCapabilities serverCapabilities, ConnectionSettings cs, bool useCompression, ProtocolCapabilities additionalCapabilities = 0) =>
+		(ProtocolCapabilities.Protocol41 |
```
bgrainger, line 9: I can't tell if this was a required change, or just a refactoring done along the way?

Either way, some change here seems to be causing a "Bad handshake" error.

src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -471,7 +445,11 @@ public async Task DisposeAsync(IOBehavior ioBehavior, CancellationToken cancella
 				ServerVersion = new(initialHandshake.ServerVersion);
 				ConnectionId = initialHandshake.ConnectionId;
 				AuthPluginData = initialHandshake.AuthPluginData;
-				m_useCompression = cs.UseCompression && (initialHandshake.ProtocolCapabilities & ProtocolCapabilities.Compress) != 0;
+				clientCapabilities =
+					HandshakeResponse41Payload.CreateClientCapabilities(initialHandshake.ProtocolCapabilities, cs, m_useCompression,
+						ProtocolCapabilities.Ssl);
```
bgrainger, line 450: Not sure that it's correct to advertise `ProtocolCapabilities.Ssl` here if a SSL connection may not be attempted.

src/MySqlConnector/Protocol/Payloads/HandshakeResponse41Payload.cs
```diff
@@ -26,12 +24,16 @@ private static ByteBufferWriter CreateCapabilitiesPayload(ProtocolCapabilities s
 		                          (useCompression ? ProtocolCapabilities.Compress : ProtocolCapabilities.None) |
 		                          ProtocolCapabilities.ConnectionAttributes |
 		                          ProtocolCapabilities.SessionTrack |
-		                          ProtocolCapabilities.DeprecateEof |
+
+		                          // ProtocolCapabilities.DeprecateEof |
```
bgrainger, line 28: Not sure that this tradeoff is worth the benefit; the EOF packet has been deprecated since MySQL 5.7.5, and this enables `CLIENT_SESSION_TRACK`, etc.

src/MySqlConnector/MySqlParameter.cs
```diff
@@ -12,6 +12,8 @@ namespace MySqlConnector;
 
 public sealed class MySqlParameter : DbParameter, IDbDataParameter, ICloneable
 {
+	public static readonly MySqlParameter NullParameter = new MySqlParameter("", null);
```
bgrainger, line 15: Avoid creating a `public static` property with a value that's a mutable object. User code could run `MySqlParameter.NullParameter.Value = 42`.

src/MySqlConnector/Core/CachedProcedure.cs
```diff
@@ -66,13 +67,16 @@ internal sealed class CachedProcedure
...
-				WHERE SPECIFIC_SCHEMA = @schema AND SPECIFIC_NAME = @component
+				WHERE SPECIFIC_SCHEMA = ? AND SPECIFIC_NAME = ?
 				ORDER BY ORDINAL_POSITION";
-			cmd.Parameters.AddWithValue("@schema", schema);
-			cmd.Parameters.AddWithValue("@component", component);
+			cmd.Prepare();
+			cmd.Parameters.Add(new() { Value = schema });
+			cmd.Parameters.Add(new() { Value = component });
+			cmd.Parameters.Add(new() { Value = schema });
+			cmd.Parameters.Add(new() { Value = component });
```
bgrainger, line 79: Are the changes in this file required for this PR, or just an optimisation (to reuse this command server-side) that seemed desirable?

bgrainger: Additionally, I think it should be unnecessary to switch from named to positional parameters (and double the parameter count), just because you're calling `Prepare`. The parameter binding code should figure that out IIRC.

bgrainger: I think this might actually be a "pessimisation"; since there are two statements, two `COM_EXECUTE_STMT` payloads will have to be sent (instead of one `COM_QUERY`), so there will be multiple DB roundtrips. A quick benchmark bears that out:

|   Method |     Mean |     Error |    StdDev | Allocated |
|--------- |---------:|----------:|----------:|----------:|
|   Normal | 3.190 ms | 0.0975 ms | 0.2766 ms |   3.43 KB |
| Prepared | 3.983 ms | 0.0783 ms | 0.1490 ms |   3.08 KB |

bgrainger (changes requested):
It seems like this PR mixes multiple refactorings (CachedProcedure, HandshakeResponse41Payload) with the actual functional change being made. Please limit the PR to the minimum code necessary to accomplish the change, and open additional PRs for the other changes. (Or structure it as a clear sequence of independent commits that can be reviewed independently.)

While you noted that this functionality is "working for any type of server > 5.0", I wonder if this code should be gated by a test for `ProtocolCapabilities.PreparedStatementMultiResults` to help compatibility with non-MySQL servers that speak the MySQL protocol?

src/MySqlConnector/Core/CachedProcedure.cs
rusher: this change can really be changed into just "cmd.Prepare();". 
About "pessimisation", this isn't exact: the first stored procedure will indeed have a PREPARE + EXECUTE command that will be slower that a COM_QUERY command, but each time another stored procedure command will be called, only an EXECUTE command will be issued, and this will be faster (especially when using mariadb now that metadata skip is implemented). I hesitate to improve PREPARE + EXECUTE in the same PR, but this will be in another PR : MariaDB permit pipelining PREPARE + EXECUTE command, EXECUTE using statement id 0xffffffff meaning using last prepared statement, this permits to have nearly the same execution time than a COM_QUERY, but next execution being just an EXECUTE command will then be way better.

src/MySqlConnector/Core/ServerSession.cs
rusher: yes, this is a bug

src/MySqlConnector/MySqlParameter.cs
rusher: correct, i'll correct that

PR #1328, opened 2023-06-10
Add ADO.NET metrics specification

bgrainger:
Opening a PR for review and feedback on a proposed specification creating a standard for the way ADO.NET providers report metrics. The goal is to use [System.Diagnostics.Metrics APIs](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics) to implement the [OpenTelemetry Semantic Conventions for Database Metrics](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/database/database-metrics.md).

The [`metrics` branch](https://github.com/mysql-net/MySqlConnector/compare/master...bgrainger:MySqlConnector:metrics) implements this specification.

### DIFF
--- a/docs/MetricsSpecification.md
+++ b/docs/MetricsSpecification.md
@@ -1,0 +1,164 @@
+# ADO.NET Metrics Specification
+
+## Introduction
+
+This specification describes how to add metrics (following the
+[OpenTelemetry Semantic Conventions for Database Metrics](https://github.com/open-telemetry/semantic-conventions/blob/main/specification/metrics/semantic_conventions/database-metrics.md))
+to ADO.NET providers using [System.Diagnostics.Metrics](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics).
+
+## Background
+
+[System.Diagnostics.Metrics APIs](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/compare-metric-apis#systemdiagnosticsmetrics) are the newest cross-platform APIs for collecting performance metrics,
+and were designed in close collaboration with the OpenTelemetry project.
+OpenTelemetry semantic conventions are standards for generating consistent,
+accessible telemetry across different databases and client libraries.
+
+## General
+
+This specification should be read in conjunction with the [OpenTelemetry Semantic Conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/specification/metrics/semantic_conventions/database-metrics.md).
+The OpenTelemetry Semantic Conventions MUST be followed except where this specification explicitly overrides them.
+
+ADO.NET Providers MUST use the [`System.Diagnostics.Metrics.Meter` class](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.meter) to expose metrics.
+It MUST be created with the name of the library and its version:
+
+```csharp
+public static Meter Meter { get; } = new("Xyzsql", "1.0.0");
+```
+
+## Pool Names
+
+OpenTelemetry Semantic Conventions assume that connection pools can be named.
+ADO.NET Providers SHOULD use the `DbDataSource` class to expose a way to create named connection pools.
+It is recommended that a `XyzsqlDataSourceBuilder` class be provided to construct DbDataSource instances,
+and that it have a `UseName(string name)` method to set the connection pool name.
+Client code could be written as follows:
+
+```csharp
+using var dataSource =
+  new XyzsqlDataSourceBuilder(connectionString)
+    .UseName(“My Pool”)
+    .Build();
+
+// Metrics for this connection have the name "My Pool" associated with them.
+using var connection = dataSource.OpenConnection();
+```
+
+Or, a dependency injection helper library might expose it as follows:
+
+```csharp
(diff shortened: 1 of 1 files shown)


bgrainger:
Need to make a decision on whether to use `IMeterFactory` on .NET 8 and/or support it in MySqlConnector.DependencyInjection.

docs/MetricsSpecification.md
```diff
@@ -0,0 +1,164 @@
...
+ADO.NET Providers MUST use the [`System.Diagnostics.Metrics.Meter` class](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.meter) to expose metrics.
+It MUST be created with the name of the library and its version:
+
+```csharp
+public static Meter Meter { get; } = new("Xyzsql", "1.0.0");
+```
+
+## Pool Names
+
+OpenTelemetry Semantic Conventions assume that connection pools can be named.
```
roji, line 30: Yeah - though note that no such concept currently exists in any ADO.NET provider as far as I'm aware. We did implement support for per-pool metrics in Npgsql via the older event counters API, using the connection string as the name - but that's not great.

I think introducing a name concept at the data source level is a great idea, opened https://github.com/npgsql/npgsql/issues/5108 to track on the Npgsql side.

API nit: I'd call it `HasName` rather than `UseName`.

docs/MetricsSpecification.md
```diff
@@ -0,0 +1,164 @@
...
+
+A highly simplified example of reporting metrics might look like this:
+
+```csharp
+public class XyzsqlConnection
+{
+  // Assume the existence of a connection pooling implementation.
+  private XyzsqlConnectionPool _pool;
+
+  public override async Task OpenAsync(CancellationToken cancellationToken)
```
roji, line 118: A similar CloseAsync example could be useful as well (e.g. to show decrementing the "usage" counter), though obviously this is all for illustration purposes anyway.

roji (approved):
This looks great @bgrainger, really looking forward to seeing it in action.

BTW I'm surprised the specs don't include anything around... command execution, which seems to be the single most important metric overall...

bgrainger:
> command execution

Commands being executed is tracked by the [semantic conventions for tracing](https://github.com/open-telemetry/semantic-conventions/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes); perhaps they didn't (yet?) want to duplicate count and timing of those operations in metrics compared to just aggregating information from the raw trace data?

roji:
Sure, but the tracing side is for non-aggregate command-by-command information; I'd expect a command execution histogram metric just like there's a histogram for how long connections were used. Not to mention metrics for command errors, timeouts...

docs/MetricsSpecification.md
```diff
@@ -0,0 +1,164 @@
...
+static readonly UpDownCounter<int> PendingRequestsCounter = Meter.CreateUpDownCounter<int>("db.client.connections.pending_requests",
+  unit: "{request}", description: "The number of pending requests for an open connection, cumulative for the entire pool.");
+static readonly Counter<int> TimeoutsCounter = Meter.CreateCounter<int>("db.client.connections.timeouts",
+  unit: "{timeout}", description: "The number of connection timeouts that have occurred trying to obtain a connection from the pool.");
+static readonly Histogram<float> CreateTimeHistogram = Meter.CreateHistogram<float>("db.client.connections.create_time",
+  unit: "ms", description: "The time it took to create a new connection.");
+static readonly Histogram<float> WaitTimeHistogram = Meter.CreateHistogram<float>("db.client.connections.wait_time",
+  unit: "ms", description: "The time it took to obtain an open connection from the pool.");
+static readonly Histogram<float> UseTimeHistogram = Meter.CreateHistogram<float>("db.client.connections.use_time",
+  unit: "ms", description: "The time it took to create a new connection.");
```
eranikid, line 79: This is copy-paste miss, reference uses `The time between borrowing a connection and returning it to the pool.` as a description for `db.client.connections.use_time`

bgrainger: Good catch; I found and fixed that copy/paste error in https://github.com/mysql-net/MySqlConnector/pull/1329/files#diff-22f5eec5ea345b8f0fd8c96f47bb2a7e59d35071c7ad2525d6f251a5ece19d5cR1963 but forgot to come back and update this example.

docs/MetricsSpecification.md
```diff
@@ -0,0 +1,164 @@
...
+OpenTelemetry Semantic Conventions assume that connection pools can be named.
+ADO.NET Providers SHOULD use the `DbDataSource` class to expose a way to create named connection pools.
+It is recommended that a `XyzsqlDataSourceBuilder` class be provided to construct DbDataSource instances,
+and that it have a `UseName(string name)` method to set the connection pool name.
+Client code could be written as follows:
+
+```csharp
+using var dataSource =
+  new XyzsqlDataSourceBuilder(connectionString)
+    .UseName(“My Pool”)
```
ejball, line 39:
```suggestion
    .UseName("My Pool")
```

bgrainger:
Archiving the proposed specification in this PR; the merged PR shall represent the definitive specification of the metrics behavior.

Issue #54, opened 2016-09-14
Support numeric conversions in MySqlDataReader


bgrainger:
PomeloFoundation/Pomelo.EntityFrameworkCore.MySql#52 suggests that other MySQL connectors allow `MySqlDataReader.GetInt32` to be called on a `decimal` column. Investigate this and add missing implicit numeric conversions.

bgrainger:
MySql.Data allows `GetInt32` to be called on a `DECIMAL` column, and will perform a lossy conversion. However, this implicit conversion isn't consistently available across all numeric data types; calling `GetByte` (for example) will throw an `InvalidCastException` for a `DECIMAL` or `INT` column, even if the value can be expressed as a `byte`.

`SqlDataReader.GetInt32` [documentation](https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqldatareader.getint32.aspx) states:

> No conversions are performed; therefore, the data retrieved must already be a 32-bit signed integer.

Calling `SqlDataReader.GetInt32` on a `NUMERIC` column will throw an `InvalidCastException`. This seems to better match the documented behaviour on the abstract base class `DbDataReader`.

I'm not in favour of cloning MySQL's lossy and inconsistent data conversion behaviour by default, but can see that it might be important for people porting from the MySql.Data connector.

The solution might be to add a new connection string parameter `Use Lossy Conversions` (or similar) that opts into behaviour that matches MySql.Data.

bgrainger:
Assuming we use a "strict mode" by default, someone who want's MySql.Data's behaviour can get it simply by calling `Convert.ToInt32(reader.GetValue(n))` instead of `reader.GetInt32(n)`. This makes it seem like it shouldn't be part of the default behaviour of the library, but opt-in by the caller if they want to pay the price of runtime conversions.

bgrainger:
One complication is that "The SUM() and AVG() functions return a DECIMAL value for exact-value arguments (integer or DECIMAL)" (http://dev.mysql.com/doc/refman/5.7/en/group-by-functions.html); that is, aggregate functions will promote an `int` column to `decimal`. 

Users of another DBMS may not be expecting this (for example, SQL Server returns an `int` when summing an `int` column), so we should allow an implicit conversion from MySQL's implicitly-promoted `decimal` back to `int`.

PR #1330, opened 2023-06-15
Faster parsing for multiple rows

rusher:
When result-set contains multiple rows, in order to get field value, the columns details are checked for each field parsing.

This PR permit to set parsing method once for the result-set. 
This doesn't change much in terms of performance, but still not negligible.

Performance results:

```
[Benchmark]
public async Task getInt64()
{
	using var cmd = Connection.CreateCommand();
	cmd.CommandText = "SELECT * FROM seq_1_to_100000";
	using var reader = await cmd.ExecuteReaderAsync();
	long total = 0;
	do
	{
		{
			total += reader.GetInt64(0);
		}
	} while (await reader.NextResultAsync());
}
```

Initial results:

```
|   Method |        Library |     Mean |    Error |
|--------- |--------------- |---------:|---------:|
| getInt64 | MySqlConnector | 18.01 ms | 0.171 ms |
```

PR results:

```
|   Method |        Library |     Mean |    Error |
|--------- |--------------- |---------:|---------:|
| getInt64 | MySqlConnector | 17.67 ms | 0.171 ms |
```


bgrainger (changes requested):
Thanks! I like the approach and it's not one I had considered. Looks like it replaces one virtual method call (to `Row.GetValueCore`) with one to `ColumnDefinitionPayload.GetValueCore`, so I don't think it will be adding any additional virtual dispatch overhead, although the former was probably much more likely to be optimised because there were only two possible derived types.

I don't like mixing the "business logic" of reading a column's value with the `ColumnDefinitionPayload` class; the `Payload` objects are simple "bundles of data" that represent exactly what gets received/sent on the wire. Instead, these should be a new hierarchy of `ColumnReader` classes (possibly in a new `MySqlConnector.ColumnReaders` namespace/folder) with `ReadValue`, `ReadInt32`, etc. methods. A factory method would create the correct derived type based on the data in the `ColumnDefinitionPayload`.

rusher:
right, i'll change PR accordingly

rusher:
force push correction. 
there is something that tickles me: I left it as it was before, but why the columns of type float and double are not allowed to retrieve a value for reader.GetInt32(<pos>) ?

rusher:
btw, good job compare to MySql.Data 8.0.33 : 
```
|   Method |        Library |      Mean |    Error |
|--------- |--------------- |----------:|---------:|
| getInt64 |     MySql.Data | 163.46 ms | 0.625 ms |
| getInt64 | MySqlConnector |  17.60 ms | 0.256 ms |

```

bgrainger:
> why the columns of type FLOAT and DOUBLE are not allowed to retrieve a value for reader.GetInt32() since it's allowed for DECIMAL

MySqlConnector doesn't implement lossy type conversions in `GetInt32`. (As a workaround, one can call `Convert.ToInt32(GetValue())` instead and potentially lose data accuracy.)

The exception is for `Decimal` because `SUM` and `AVG` will return a `DECIMAL` value even for integral data: https://github.com/mysql-net/MySqlConnector/issues/54#issuecomment-273275754. This is to avoid unexpected failure when use `GetInt32` to read the `SUM` of an `INT` column.

src/MySqlConnector/Core/ResultSet.cs
```diff
@@ -147,12 +147,12 @@ public async Task ReadResultSetHeaderAsync(IOBehavior ioBehavior)
...
 
-						if (!Session.SupportsDeprecateEof)
-						{
-							payload = await Session.ReceiveReplyAsync(ioBehavior, CancellationToken.None).ConfigureAwait(false);
-							EofPayload.Create(payload.Span);
-						}
+					if (!Session.SupportsDeprecateEof)
+					{
+						payload = await Session.ReceiveReplyAsync(ioBehavior, CancellationToken.None).ConfigureAwait(false);
+						EofPayload.Create(payload.Span);
```
bgrainger, line 155: Is this change correct/intentional?

It seems like it could require an additional EOF packet for the case when metadata from a prepared statement is reused. If so, it feels like integration tests shouldn't pass both before and after this change. Perhaps `MARIADB_CLIENT_CACHE_METADATA` always implies `CLIENT_DEPRECATE_EOF` so that this condition is always `false` and the `if` block is never entered?

bgrainger: Just reviewed https://mariadb.com/kb/en/result-set-packets/ and confirmed that this change is correct.

src/MySqlConnector/ColumnReaders/BinaryBooleanColumnReader.cs
```diff
@@ -0,0 +1,20 @@
+namespace MySqlConnector.ColumnReaders;
+using System.Buffers.Text;
+using MySqlConnector.Protocol.Payloads;
+using MySqlConnector.Protocol.Serialization;
+using MySqlConnector.Utilities;
+
+internal sealed class BinaryBooleanColumnReader : IColumnReader
+{
+	internal static BinaryBooleanColumnReader Instance { get; } = new BinaryBooleanColumnReader();
```
bgrainger, line 9: Singleton implementations of these classes is a good idea to reduce allocations 👍

src/MySqlConnector/ColumnReaders/BinaryDateTimeColumnReader.cs
```diff
@@ -0,0 +1,74 @@
...
+using MySqlConnector.Core;
+using MySqlConnector.Protocol.Payloads;
+using MySqlConnector.Protocol.Serialization;
+using MySqlConnector.Utilities;
+
+internal sealed class BinaryDateTimeColumnReader : IColumnReader
+{
+	private bool allowZeroDateTime;
+	private bool convertZeroDateTime;
+	private DateTimeKind dateTimeKind;
```
bgrainger, line 16: Please follow the style of existing classes:

* fields are placed at the end of the class definition
* private fields are prefixed with `m_`

Additionally, these fields should be `readonly`.

Will fix in https://github.com/mysql-net/MySqlConnector/pull/1331/commits/10a3c08a9d7cbea5a2de1a57c6cb048cb96390d7.

src/MySqlConnector/Core/Row.cs
```diff
@@ -567,4 +518,5 @@ private static void CheckBufferArguments<T>(long dataOffset, T[] buffer, int buf
 	private readonly int[] m_dataOffsets;
 	private readonly int[] m_dataLengths;
 	private ReadOnlyMemory<byte> m_data;
+	private IColumnReader[] columnReaders;
```
bgrainger, line 521: `private readonly IColumnReader[] m_columnReaders;`

Will fix in https://github.com/mysql-net/MySqlConnector/pull/1331/commits/1cc81efbb218269a98eaf23e4e53ae469a707e40.

src/MySqlConnector/ColumnReaders/BinaryBooleanColumnReader.cs
bgrainger: These `Instance` properties (or constructors for some types) are still logically part of the `public` API and should be `public` (even on an `internal` type).

bgrainger: We can also use target-typed `new` here.

Will fix in https://github.com/mysql-net/MySqlConnector/pull/1332/commits/4d0560412933a87d351478b08052510743d3ae46.

src/MySqlConnector/ColumnReaders/ColumnReaderFactory.cs
```diff
@@ -0,0 +1,204 @@
...
+							    or MySqlGuidFormat.LittleEndianBinary16) && columnDefinition.ColumnLength == 16)
+						{
+							switch (guidFormat)
+							{
+							case MySqlGuidFormat.Binary16:
+								return Guid16ColumnReader.Instance;
+							case MySqlGuidFormat.TimeSwapBinary16:
+								return TimeSwapBinary16ColumnReader.Instance;
+							default:
+								return GuidBytesColumnReader.Instance;
```
bgrainger, line 60: These classes have somewhat random and inconsistent names. It would make more sense for them to be named consistently with the `MySqlGuidFormat` enum values, e.g., `GuidBinary16ColumnReader`, `GuidTimeSwapBinary16ColumnReader`, `GuidLittleEndianBinary16ColumnReader`.

Will fix in https://github.com/mysql-net/MySqlConnector/pull/1332/commits/e400f960af65cc5e2d604321495cce75e64c22e6.

src/MySqlConnector/ColumnReaders/ColumnReaderFactory.cs
```diff
@@ -0,0 +1,204 @@
...
+				case ColumnType.Bit:
+					return BitColumnReader.Instance;
+
+				case ColumnType.String:
+					if (connection.GuidFormat == MySqlGuidFormat.Char36
+					    && columnDefinition.ColumnLength / ProtocolUtility.GetBytesPerCharacter(columnDefinition.CharacterSet) == 36)
+						return Guid36ColumnReader.Instance;
+					if (connection.GuidFormat == MySqlGuidFormat.Char32
+					    && columnDefinition.ColumnLength / ProtocolUtility.GetBytesPerCharacter(columnDefinition.CharacterSet) == 32)
+						return Guid32ColumnReader.Instance;
```
bgrainger, line 134: As above, should be `GuidChar36ColumnReader` and `GuidChar32ColumnReader`.

src/MySqlConnector/ColumnReaders/IColumnReader.cs
```diff
@@ -0,0 +1,9 @@
+namespace MySqlConnector.ColumnReaders;
+using MySqlConnector.Protocol.Payloads;
+
+internal interface IColumnReader
```
bgrainger, line 4: Seems like an `abstract` base class would be more useful here; it could contain the commonly repeated `throw new InvalidCastException` implementation for `GetInt32`.

Additionally, if a (say) `GetInt64` method were added in the future, a common implementation could be added once in the ABC, instead of having to be repeated in each concrete implementation.

Will fix in https://github.com/mysql-net/MySqlConnector/pull/1332/commits/e130a7dc8c28694f23d588c1be40ffca3313c22b.

src/MySqlConnector/ColumnReaders/BitColumnReader.cs
```diff
@@ -0,0 +1,61 @@
...
+		}
+		else
+		{
+			// when the Binary flag IS set, the BIT column is transmitted as text
+			if (!Utf8Parser.TryParse(data, out long value, out var bytesConsumed) || bytesConsumed != data.Length)
+			{
+				throw new FormatException();
+			}
+			return checked((int) value);
+		}
```
bgrainger, line 59: There's some significant code duplication happening here. This should be refactored into a helper method.

Will fix in https://github.com/mysql-net/MySqlConnector/pull/1332/commits/34bf3aa111a71dc57354989ec60a5f8987fd8ab8.

src/MySqlConnector/ColumnReaders/DecimalColumnReader.cs
```diff
@@ -0,0 +1,24 @@
...
+		return Utf8Parser.TryParse(data, out decimal decimalValue, out int bytesConsumed) && bytesConsumed == data.Length ? decimalValue : throw new FormatException();
+	}
+
+	public int ReadInt32(ReadOnlySpan<byte> data, ColumnDefinitionPayload columnDefinition)
+	{
+		if (!Utf8Parser.TryParse(data, out decimal decimalValue, out int bytesConsumed) || bytesConsumed != data.Length)
+		{
+			throw new FormatException();
+		}
+		return (int) decimalValue;
```
bgrainger, line 22: This code duplication could be refactored into a helper method, too.

src/MySqlConnector/ColumnReaders/BinaryDateTimeColumnReader.cs
```diff
@@ -0,0 +1,74 @@
...
+		}
+		catch (Exception ex)
+		{
+			throw new FormatException($"Couldn't interpret value as a valid DateTime: {Encoding.UTF8.GetString(data)}", ex);
+		}
+	}
+
+	public int ReadInt32(ReadOnlySpan<byte> data, ColumnDefinitionPayload columnDefinition)
+	{
+		throw new InvalidCastException($"Can't convert {columnDefinition.ColumnType} to Int32");
```
bgrainger, line 72: This uses the internal `ColumnType` enum. The [previous code](https://github.com/mysql-net/MySqlConnector/pull/1330/files#diff-dd36203ad08fa982528689516ef6d116f29ec3f8477e50d782cbc5e590526550L187) used a `MySqlDbType` value:

```csharp
throw new InvalidCastException($"Can't convert {ResultSet.GetColumnType(ordinal)} to Int32");
```

Will fix in https://github.com/mysql-net/MySqlConnector/pull/1332/commits/88a54a64bf13a872ebf930c1a694aea88e9bdf01.

src/MySqlConnector/ColumnReaders/ColumnReaderFactory.cs
```diff
@@ -0,0 +1,204 @@
...
+using MySqlConnector.Protocol.Serialization;
+
+internal class ColumnReaderFactory
+{
+	internal static IColumnReader GetReader(bool binary, ColumnDefinitionPayload columnDefinition, MySqlConnection connection)
+	{
+		var isUnsigned = (columnDefinition.ColumnFlags & ColumnFlags.Unsigned) != 0;
+		if (binary)
+		{
+			switch (columnDefinition.ColumnType)
```
bgrainger, line 13: There's quite a bit of duplicated logic between the `binary` and `!binary` `switch` statements that could be consolidated.

Will fix in https://github.com/mysql-net/MySqlConnector/pull/1332/commits/808f91fc42400b4bd88f75e3b7df6a9e590721d9.

src/MySqlConnector/Core/Row.cs
```diff
@@ -393,17 +370,17 @@ public int GetValues(object[] values)
...
 	public object this[string name] => GetValue(ResultSet.GetOrdinal(name));
 
-	protected Row(ResultSet resultSet)
+	protected Row(bool binary, ResultSet resultSet)
 	{
 		ResultSet = resultSet;
 		m_dataOffsets = new int[ResultSet.ColumnDefinitions!.Length];
 		m_dataLengths = new int[ResultSet.ColumnDefinitions.Length];
+		columnReaders = Array.ConvertAll(ResultSet.ColumnDefinitions,
+			new Converter<ColumnDefinitionPayload, IColumnReader>(column => ColumnReaderFactory.GetReader(binary, column, resultSet.Connection)));
```
bgrainger, line 379: Using `Array.ConvertAll` and a delegate allocation seems needlessly inefficient compared to a `foreach` loop.

Will fix in https://github.com/mysql-net/MySqlConnector/pull/1332/commits/3e48b93214d105b06e13b88f1531e471dc50ba10.

src/MySqlConnector/Core/BinaryRow.cs
```diff
@@ -11,7 +11,7 @@ namespace MySqlConnector.Core;
 internal sealed class BinaryRow : Row
```
bgrainger, line 11: There's so little code left in these classes they could probably be eliminated.

Will fix in https://github.com/mysql-net/MySqlConnector/pull/1332/commits/acd9b9475bf6841c637b1b7777ddfadd37df7ca5.

src/MySqlConnector/Core/ResultSet.cs
rusher: yes, this might have been done through another PR. This is a correction that might normally never occurs (there is no version that support skipping metadata without supporting EOF) but better to correct that just in case.

PR #1334, opened 2023-06-20
Activities - update handling status

Kielek:
This PR changes the way how to handle ActivityStatus

1. Manually setting `otel.status_code` and `otel.status_description` is no longer needed. System.Diagnostics.DiagnosticSource v6+ handles it natively. Ref: https://github.com/open-telemetry/opentelemetry-dotnet/issues/2569
2. Avoid setting spanstasus.ok. The unset value is preferred one https://opentelemetry.io/docs/concepts/signals/traces/#span-status
3. Removed unused fields/constants.
4. Conditional compilation simplified - it should work on any .NET version. As long as it is referencing System.Diagnostics.DiagnosticSource v6+


bgrainger:
Thanks for the improvement!

PR #1335, opened 2023-06-20
Small test improvement

rusher:
* MariaDB and MySQL server differ when executing KILL QUERY on SELECT SLEEP(x) command. MariaDB server send an ERR_Packet "Query execution was interrupted" while mysql send an OK_Packet. This PR permit to enable all tests using SLEEP(x).
* Correct ProvidePasswordCallback test when no password are used in testing environment

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -196,15 +196,15 @@ jobs:
       'MariaDB 10.6':
         image: 'mariadb:10.6'
         connectionStringExtra: ''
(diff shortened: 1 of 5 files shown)


bgrainger:
I've enabled the `CancelSleepSuccessfully` tests in https://github.com/mysql-net/MySqlConnector/commit/d9fffcb8be5e2e56073e85375f6ca291738eabfd with less code duplication.

The `ProvidePasswordCallback` test is a good catch, but should be skipped entirely with something like `ConfigSettings.PasswordlessUser` instead of bypassing the only `Assert` in the method.

bgrainger:
Tests for passwordless users fixed in https://github.com/mysql-net/MySqlConnector/commit/58af81126f451fbc823aef496aa7f6eba0cd78cf.

PR #1337, opened 2023-06-21
Various change about timeout

rusher:
(repushed of https://github.com/mysql-net/MySqlConnector/pull/1336 that was on wrong branch)

* removing ScanRowAsync RegisterCancel(cancellationToken) since it is always registered on public method before.
* removing SendReplyAsync and ReceiveReplyAsync CancellationToken parameter, since not used
* change reset command timeout, in order to have faster second execution.

Command ResetCommandTimeout implementation now in MySqlCommand/MySqlBatch have been completly redone. Some of the previous implementation was a little weird, so i've rewrite them. This part in particulary need review.

benchmarks show improvement for resultset with multiple rows and for single row resultset without timeout. No change for single row with timeout:

```
before
|                  Method |        Library |         Mean |
|------------------------ |--------------- |-------------:|
|            Get10000Rows | MySqlConnector | 17,596.28 us |
|                Get1Rows | MySqlConnector |     36.50 us |
| Get10000RowsWithTimeout | MySqlConnector | 21,416.11 us |
|     Get1RowsWithTimeout | MySqlConnector |     35.85 us |
after PR:
|------------------------ |--------------- |-------------:|
|            Get10000Rows | MySqlConnector | 16,293.13 us |
|                Get1Rows | MySqlConnector |     35.51 us |
| Get10000RowsWithTimeout | MySqlConnector | 20,952.15 us |
|     Get1RowsWithTimeout | MySqlConnector |     35.84 us |
```
with test:

```
	[Benchmark]
	public async Task<long> Get10000Rows()
	{
		using var cmd = Connection.CreateCommand();
		cmd.CommandText = "SELECT * FROM seq_1_to_100000";
		using var reader = await cmd.ExecuteReaderAsync();
		long total = 0;
		do
		{
			{
				total += reader.GetInt64(0);
			}
		} while (await reader.NextResultAsync());

		return total;
	}

	[Benchmark]
	public async Task<long> Get1Rows()
	{
		using var cmd = Connection.CreateCommand();
		cmd.CommandText = "SELECT 1";
		using var reader = await cmd.ExecuteReaderAsync();
		long total = 0;
		do
		{
			{
				total += reader.GetInt64(0);
			}
		} while (await reader.NextResultAsync());

		return total;
	}

	[Benchmark]
	public async Task<long> Get10000RowsWithTimeout()
	{
		using var cmd = Connection.CreateCommand();
		cmd.CommandText = "SELECT * FROM seq_1_to_100000";
		cmd.CommandTimeout = 1000;
		using var reader = await cmd.ExecuteReaderAsync();
		long total = 0;
		do
		{
			{
				total += reader.GetInt64(0);
			}
		} while (await reader.NextResultAsync());

		return total;
	}

	[Benchmark]
	public async Task<long> Get1RowsWithTimeout()
	{
		using var cmd = Connection.CreateCommand();
		cmd.CommandText = "SELECT 1";
		cmd.CommandTimeout = 1000;
		using var reader = await cmd.ExecuteReaderAsync();
		long total = 0;
		do
		{
			{
				total += reader.GetInt64(0);
			}
		} while (await reader.NextResultAsync());

		return total;
	}
```


src/MySqlConnector/MySqlCommand.cs
```diff
@@ -422,6 +427,58 @@ private void CancelCommandForTimeout()
 		Connection?.Cancel(this, m_commandId, false);
 	}
 
+	public void ResetCommandTimeout()
```
bgrainger, line 430: Why is this a new `public` method?

src/MySqlConnector/MySqlCommand.cs
```diff
@@ -390,8 +397,7 @@ public Task DisposeAsync()
 		if (!cancellationToken.CanBeCanceled)
 			return null;
 
-		m_cancelAction ??= Cancel;
-		return cancellationToken.Register(m_cancelAction);
+		return cancellationToken.Register(Cancel);
```
bgrainger, line 400: When this code was written, not caching the delegate would result in a lot more allocations; has that changed?

src/MySqlConnector/MySqlBatch.cs
```diff
@@ -254,8 +254,7 @@ public void Dispose()
 		if (!cancellationToken.CanBeCanceled)
 			return null;
 
-		m_cancelAction ??= Cancel;
-		return cancellationToken.Register(m_cancelAction);
+		return cancellationToken.Register(Cancel);
```
bgrainger, line 257: Same comment about allocations.

src/MySqlConnector/MySqlCommand.cs
```diff
@@ -422,6 +427,58 @@ private void CancelCommandForTimeout()
...
+				((ICancellableCommand) this).SetTimeout(CommandTimeout * 1000);
+				session.SetTimeout((session.CancellationTimeout <= 0 ?
+					                   CommandTimeout : CommandTimeout + session.CancellationTimeout)
+				                   * 1000);
+				m_reset = ResetCommandTimeoutSet;
+			}
+		}
+	}
+
+	private void ResetCommandTimeoutSet()
```
bgrainger, line 470: Method names should generally take the form `VerbNoun`. I'm not sure what `ResetCommandDefaultSet` means, exactly.

src/MySqlConnector/MySqlBatch.cs
```diff
@@ -314,6 +312,26 @@ private async Task<int> ExecuteNonQueryAsync(IOBehavior ioBehavior, Cancellation
 		return result;
 	}
 
+	public void ResetCommandTimeout()
```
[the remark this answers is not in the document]
bgrainger: But it could be `internal`, right? It doesn't need to be a new part of the `DbCommand` public API.

bgrainger:
Your benchmarks without a timeout still use the `DefaultCommandTimeout` from the connection string, unless you're setting that to `0` in the benchmark app, but not showing that here. https://mysqlconnector.net/connection-options/#DefaultCommandTimeout

src/MySqlConnector/MySqlCommand.cs
rusher: this may be not clear on first read, but this "ResetCommandDefaultSet" is the first "ResetCommandTimeout" that is called, setting m_reset for further  call of ResetCommandTimeout (if CommandTimeout is not changed)

src/MySqlConnector/Core/ICancellableCommand.cs
```diff
@@ -23,60 +23,5 @@ internal static class ICancellableCommandExtensions
...
-	/// to happen, then the "hard" timeout to occur.
-	/// </summary>
-	/// <remarks>As per the <a href="https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqlcommand.commandtimeout.aspx">MSDN documentation</a>,
-	/// "This property is the cumulative time-out (for all network packets that are read during the invocation of a method) for all network reads during command
-	/// execution or processing of the results. A time-out can still occur after the first row is returned, and does not include user processing time, only network
-	/// read time. For example, with a 30 second time out, if Read requires two network packets, then it has 30 seconds to read both network packets. If you call
-	/// Read again, it will have another 30 seconds to read any data that it requires."
-	/// The <see cref="ResetCommandTimeout"/> method is called by public ADO.NET API methods to reset the effective time remaining at the beginning of a new
-	/// method call.</remarks>
-	public static void ResetCommandTimeout(this ICancellableCommand command)
```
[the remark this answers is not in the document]
bgrainger: I need to study the change more closely, as I may have missed what it's attempting to do ("avoid executing complete ResetCommandTimeout method each time").

rusher:
yes, url was using "Default Command Timeout=0"

src/MySqlConnector/MySqlCommand.cs
rusher: ok, i see the point, wrong correction then

bgrainger:
OK, I see your insight that `ResetCommandTimeout` repeatedly does unnecessary work if `CommandTimeout == 0`. I've tried to accomplish the same thing with a simple `int?` to track the state in #1338.

What's your experimental setup where `Get1Rows` takes 36.50 µs? I'm running MySQL 8.0 under Docker on Windows, and I'm benchmarking that at 900+ µs.

bgrainger:
> What's your experimental setup

By running `mariadbd.exe` directly on Windows, I was able to get times of about 80 µs to select a single row, which is much closer to your experimental output.

rusher:
closing this one since https://github.com/mysql-net/MySqlConnector/pull/1338 handle this change better than this PR. 

About your benchmark, this is strange to have docker this slow. my observation depending on computer, is that single thread benchmark on simple command are generally between 10 and 50 µs using local server and connector.

PR #1394, opened 2023-11-15
Support empty array as restrictionValues

Thorium:
There is no reason why this driver should crash if empty array as restrictionValues is provided as argument.


bgrainger:
Thanks for the contribution! Do you certify the DCO terms (https://developercertificate.org/) for this contribution?

Thorium:
Yes.

Issue #1398, opened 2023-11-21
Read past end of buffer. in `ColumnCountPayload`


trejjam:
**Software versions**
MySqlConnector version: 2.3.1
Server type (MySQL, MariaDB, Aurora, etc.) and version: 10.6.5-MariaDB-1:10.6.5+maria~stretch
.NET version: 7.0
ORM: EntityFramework 7.0.14

**Describe the bug**
The server indicates the `MariaDbCacheMetadata` capability.

[`(MySqlConnector.Protocol.Payloads.ColumnCountPayload.Create()`](https://github.com/mysql-net/MySqlConnector/blob/master/src/MySqlConnector/Protocol/Payloads/ColumnCountPayload.cs) is called with the following arguments `new [1] { 2 }.AsSpan(), true`. The first parameter is sent as `new [1] { 1, 1 }.AsSpan()` which works fine.

```cs
var reader = new ByteArrayReader(span);
var columnCount = (int) reader.ReadLengthEncodedInteger();

// since `supportsOptionalMetadata is false` `reader.ReadByte()` is evaluated which causes the exception
var metadataFollows = !supportsOptionalMetadata || reader.ReadByte() == 1;
return new ColumnCountPayload(columnCount, metadataFollows);
```

**Exception**
System.InvalidOperationException
Read past end of buffer.

```
System.InvalidOperationException: Read past end of buffer.
  File "/_/src/MySqlConnector/Protocol/Serialization/ByteArrayReader.cs", line 158, col 4, in void ByteArrayReader.VerifyRead(int length)
  File "/_/src/MySqlConnector/Protocol/Serialization/ByteArrayReader.cs", line 22, col 3, in byte ByteArrayReader.ReadByte()
  File "/_/src/MySqlConnector/Protocol/Payloads/ColumnCountPayload.cs", line 24, col 3, in ColumnCountPayload ColumnCountPayload.Create(ReadOnlySpan<byte> span, bool supportsOptionalMetadata)
  File "/_/src/MySqlConnector/Core/ResultSet.cs", line 117, col 6, in async Task ResultSet.ReadResultSetHeaderAsync(IOBehavior ioBehavior)
MySqlConnector.MySqlException: Failed to read the result set.
  File "/_/src/MySqlConnector/MySqlDataReader.cs", line 133, col 4, in void MySqlDataReader.ActivateResultSet(CancellationToken cancellationToken)
[...]

PR #1399, opened 2023-11-21
Handle missing metadata

trejjam:
Solve #1398


bgrainger:
Thanks; I think this makes sense to accommodate a "truncated" packet sent from the server as though it doesn't actually support `MARIADB_CLIENT_CACHE_METADATA`. (A packet capture to show exactly what's going on would still be interesting.)

Do you agree that these contributions are made under the terms of https://developercertificate.org/? (I can set that check to pass.)

trejjam:
ToC approve

trejjam:
I hope here is the problematic select: The issue is with 2. column, 1. has metadata
[problematic-select.zip](https://github.com/mysql-net/MySqlConnector/files/13430729/problematic-select.zip)

trejjam:
[mysql-missing-metadata.zip](https://github.com/mysql-net/MySqlConnector/files/13430805/mysql-missing-metadata2.zip)

Issue #1410, opened 2023-12-06
Request to support config LoggerFactory without using MySqlDataSource


yang-xiaodong:
According to the document description at https://mysqlconnector.net/diagnostics/logging/, it seems that MySqlConnector currently does not support configuring LogFactory without using `MySqlDataSourceBuilder`. This makes it impossible to initialize loggerfactory in some scenarios where dataSourceBuilder cannot be used.

Request to add static objects to support config logfactory without using  datasource like pgsql.

See: 
https://www.npgsql.org/doc/diagnostics/logging.html#configuration-without-npgsqldatasource

bgrainger:
See "Global logging" here: https://mysqlconnector.net/diagnostics/logging/.

bgrainger:
Npgsql's API is documented as "legacy only": https://github.com/npgsql/npgsql/blob/074703eb3596eb2451f7858bf9ea04b6f3e2e720/src/Npgsql/NpgsqlLoggingConfiguration.cs#L48-L49

MySqlConnector already has a legacy interface; I don't know if there's really a need to add a new, legacy interface?

bgrainger:
There are instances where it could be nice to set a global `ILoggerFactory` without having to pull in the MySqlConnector.Logging.Microsoft.Extensions.Logging package; this could be done via a new `MySqlConnectorLogManager.InitializeLogging(ILoggerFactory)` method.

PR #1412, opened 2023-12-07
Support configuration with MySqlConnectorLoggingConfiguration without using MySqlDataSource

yang-xiaodong:
The visibility of `MySqlConnectorLoggingConfiguration` has been changed from `internal` to `public`, and an `InitializeLogging` static method has been added to allow users to globally initialize `LoggerFactory`. Other internal properties of the object have been changed from `public` to `internal`, making them invisible to users.

A `GlobalLoggerFactory` object property has been created, which allows in special scenarios the retrieval of the configured `GlobalLoggerFactory` via reflection, thereby enabling the addition of more logging providers.

Issue: #1410

### DIFF
--- a/src/MySqlConnector/Logging/MySqlConnectorLoggingConfiguration.cs
+++ b/src/MySqlConnector/Logging/MySqlConnectorLoggingConfiguration.cs
@@ -3,14 +3,28 @@ using Microsoft.Extensions.Logging.Abstractions;
 
 namespace MySqlConnector.Logging;
 
-internal sealed class MySqlConnectorLoggingConfiguration(ILoggerFactory loggerFactory)
+public sealed class MySqlConnectorLoggingConfiguration(ILoggerFactory loggerFactory)
 {
-	public ILogger DataSourceLogger { get; } = loggerFactory.CreateLogger("MySqlConnector.MySqlDataSource");
-	public ILogger ConnectionLogger { get; } = loggerFactory.CreateLogger("MySqlConnector.MySqlConnection");
-	public ILogger CommandLogger { get; } = loggerFactory.CreateLogger("MySqlConnector.MySqlCommand");
-	public ILogger PoolLogger { get; } = loggerFactory.CreateLogger("MySqlConnector.ConnectionPool");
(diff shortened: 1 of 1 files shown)


src/MySqlConnector/Logging/MySqlConnectorLoggingConfiguration.cs
```diff
@@ -3,14 +3,28 @@
...
-	public ILogger BulkCopyLogger { get; } = loggerFactory.CreateLogger("MySqlConnector.MySqlBulkCopy");
+	internal ILogger DataSourceLogger { get; } = loggerFactory.CreateLogger("MySqlConnector.MySqlDataSource");
+	internal ILogger ConnectionLogger { get; } = loggerFactory.CreateLogger("MySqlConnector.MySqlConnection");
+	internal ILogger CommandLogger { get; } = loggerFactory.CreateLogger("MySqlConnector.MySqlCommand");
+	internal ILogger PoolLogger { get; } = loggerFactory.CreateLogger("MySqlConnector.ConnectionPool");
+	internal ILogger BulkCopyLogger { get; } = loggerFactory.CreateLogger("MySqlConnector.MySqlBulkCopy");
 
-	public static MySqlConnectorLoggingConfiguration NullConfiguration { get; } = new MySqlConnectorLoggingConfiguration(NullLoggerFactory.Instance);
-	public static MySqlConnectorLoggingConfiguration GlobalConfiguration { get; set; } = NullConfiguration;
+	internal static ILoggerFactory GlobalLoggerFactory { get; set; } = NullLoggerFactory.Instance;
```
bgrainger, line 14: What is the purpose of this static property? I don't see where it's used (outside this class).

yang-xiaodong:
> Add multiple ILoggerProviders to an ILoggerFactory then set MySqlConnector to use that (singleton) ILoggerFactory.

Yes, But can add logger providers at runtime

bgrainger:
How does the `GlobalLoggerFactory` property support adding multiple loggers at runtime?

Additionally, this should not be the responsibility of MySqlConnector. Changing the logging configuration at runtime should be a function of the logging framework you're using.

yang-xiaodong:
Hi @bgrainger ,

Thanks for your reply. I'm closing this PR now!

Maybe adding logging provider in the runtime can be as a enhancement backlog, thanks.

PR #1418, opened 2023-12-14
[NEW]added Foreign keys, Indexes, IndexColumns to the schema provider

eaexo:
as per discussion in https://github.com/mysql-net/MySqlConnector/issues/375 and https://github.com/mysql-net/MySqlConnector/pull/1188#issuecomment-1852373890  I create a PR, the changes work for our project

### DIFF
--- a/docs/content/overview/schema-collections.md
+++ b/docs/content/overview/schema-collections.md
@@ -1,12 +1,11 @@
 ---
-lastmod: 2023-01-28
 date: 2021-04-24
-title: Schema Collections
-customtitle: "Supported Schema Collections"
-description: Information about the schemas supported by MySqlConnection.GetSchema.
+lastmod: 2023-12-14
 menu:
   main:
     parent: getting started
+title: Schema Collections
+customtitle: "Supported Schema Collections"
 weight: 80
 ---
 
@@ -45,3 +44,6 @@ weight: 80
 * `Triggers`
 * `UserPrivileges`
 * `Views`
+* `Foreign Keys`
+* `Indexes`
+* `IndexColumns`
--- a/docs/content/overview/schema/columns.md
+++ b/docs/content/overview/schema/columns.md
@@ -1,6 +1,6 @@
 ---
 date: 2022-07-10
-lastmod: 2022-07-11
+lastmod: 2023-12-14
 title: Columns Schema
 ---
 
--- a/docs/content/overview/schema/datatypes.md
+++ b/docs/content/overview/schema/datatypes.md
@@ -1,6 +1,6 @@
 ---
 date: 2022-07-10
-lastmod: 2022-07-11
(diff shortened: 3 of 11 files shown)


bgrainger:
Thanks for the contribution. 

There seems to be an error with the `Foreign Keys` schema: it's returning a large number of rows where the constraint name is `PRIMARY` and `REFERENCED_TABLE_SCHEMA`, `REFERENCED_TABLE_NAME`, `REFERENCED_COLUMN_NAME` are all `null`. Connector/NET doesn't do this, and I assume it's incorrect to return PKs in a schema for FKs.

This PR is also missing the `MATCH_OPTION`, `UPDATE_RULE`, `DELETE_RULE`, `REFERENCED_TABLE_CATALOG` columns that Connector/NET provides. I don't know how critical those are, but it's plausible that `GetSchema` clients are relying on those columns and so they should be returned. (This PR also adds `COLUMN_NAME` and `REFERENCED_COLUMN_NAME`; there doesn't seem to me to be any harm in adding extra columns to the output.)

bgrainger:
The schemas returned for `Indexes` and `IndexColumns` is also significantly different between this PR and Connector/NET.

If MySqlConnector doesn't return the same (or almost the same) schemas as MySQL Connector/NET then it's not clear to me that this is an improvement. For example, if NHibernate is looking for specific columns in the returned DataTables, does returning a different schema than what NHibernate expects (for the same name) solve the problem? (I guess maybe it does since you said "the changes work for our project"?).

eaexo:
> The schemas returned for `Indexes` and `IndexColumns` is also significantly different between this PR and Connector/NET.
> 
> If MySqlConnector doesn't return the same (or almost the same) schemas as MySQL Connector/NET then it's not clear to me that this is an improvement. For example, if NHibernate is looking for specific columns in the returned DataTables, does returning a different schema than what NHibernate expects (for the same name) solve the problem? (I guess maybe it does since you said "the changes work for our project"?).

Hello @bgrainger 

I focused on getting NHibernate and schema update running.

If you check NHiberate they use collection/schema name "Foreign Keys" via ForeignKeysSchemaName. For whatever reason it is "Foreign Keys" with space, where as for all other dialects it is "ForeignKeys" (without space), see MySQLDataBaseSchema.

If one further looks at MySQLForeignKeyMetadata/MySQLIndexMetadata only relevant columns are CONSTRAINT_NAME for foreign key and INDEX_NAME for the index.

With these changes it works in NHibernate 5.4

bgrainger:
Thanks for the contribution!

There's some unnecessary formatting changes in the YAML file in this PR. Since I can't push to your branch, I'm going to close this PR and replace it with https://github.com/mysql-net/MySqlConnector/pull/1419.

bgrainger:
Merged via #1419.

Issue #1384, opened 2023-11-03
MySqlParamater Value unboxing enum - cast  exception


BOBJohnson:
**Software versions**
MySqlConnector version: 2.2.7
Server type (MySQL, MariaDB, Aurora, etc.) and version: MySQL 5.7 & 8.0 AWS RDS
.NET version: 4.8

**Describe the bug**
Not sure if this is a bug, or a misunderstanding on my part. Recently migrated from Oracle's connector to your connector. Other than swapping namespaces, drop in was seamless.

Found an issue during testing from unboxing an enum value that is being passed into a MySqlParameter. In Oracle's connector, no unboxing / cast exception.

The column in question has a MySql type of smallint - so a short/int16 in .Net
The enum has always just been the default underlying type in .Net (int32).
The amount of values the enum has nowhere the limit of int16 - so not worried about an overflow when casting.

**Exception**
System.InvalidCastException: Specified cast is not valid.
   at MySqlConnector.MySqlParameter.AppendSqlString(ByteBufferWriter writer, StatementPreparerOptions options) in /_/src/MySqlConnector/MySqlParameter.cs:line 463
   at MySqlConnector.Core.StatementPreparer.ParameterSqlParser.DoAppendParameter(Int32 parameterIndex, Int32 textIndex, Int32 textLength) in /_/src/MySqlConnector/Core/StatementPreparer.cs:line 88
   at MySqlConnector.Core.StatementPreparer.ParameterSqlParser.OnNamedParameter(Int32 index, Int32 length) in /_/src/MySqlConnector/Core/StatementPreparer.cs:line 77
   at MySqlConnector.Core.SqlParser.Parse(String sql) in /_/src/MySqlConnector/Core/SqlParser.cs:line 197
   at MySqlConnector.Core.StatementPreparer.ParseAndBindParameters(ByteBufferWriter writer) in /_/src/MySqlConnector/Core/StatementPreparer.cs:line 39
   at MySqlConnector.Core.SingleCommandPayloadCreator.WriteCommand(IMySqlCommand command, ByteBufferWriter writer, Boolean appendSemicolon) in /_/src/MySqlConnector/Core/SingleCommandPayloadCreator.cs:line 260
[...]

PR #1421, opened 2023-12-17
Fixed packet corruption issue when enum was not 4 bytes

daebo01:
Fixed an issue where packets were corrupted when enum was not 4 bytes

This appears to be because the column type and the actual size of data used in the SingleCommandPayloadCreator.WriteBinaryParameters function are different.

Code and screenshots to reproduce the issue are provided below

Please check

```
CREATE TABLE `test1` (
	`Id` INT NOT NULL AUTO_INCREMENT,
	`A` TINYINT NOT NULL,
	`B` INT NOT NULL,
	`C` INT NOT NULL,
	PRIMARY KEY (`Id`)
);
```

```
internal class TestLogic
{
    private readonly MySqlConnection _connection;

    public TestLogic(string connStr)
    {
        _connection = new MySqlConnection(connStr);
    }

    public async Task Run()
    {
        
        await _connection.OpenAsync();

        await InsertDummyRow();
        var list = await SelectAllRow();

        foreach (var row in list)
        {
            Console.WriteLine(row);
        }
    }

    private async Task InsertDummyRow()
    {
        using var cmd = _connection.CreateCommand();
        cmd.CommandText = @"
            INSERT INTO test1 (A, B, C) 
                   VALUES (@A, @B, @C)";

        cmd.Parameters.AddWithValue("@A", TestEnum1.X);
        cmd.Parameters.AddWithValue("@B", 0x40302010);
        cmd.Parameters.AddWithValue("@C", 0);

        await cmd.PrepareAsync();

        await cmd.ExecuteNonQueryAsync();
    }

    private async Task<List<TestTable1Model>> SelectAllRow()
    {
        using var cmd = _connection.CreateCommand();
        cmd.CommandText = @"
            SELECT Id, A, B, C
              FROM test1 LIMIT 1000";

        await cmd.PrepareAsync();

        var list = new List<TestTable1Model>();
        var reader = await cmd.ExecuteReaderAsync();

        {
            var row = new TestTable1Model
            {
                Id = reader.GetInt32(0),
                A = (TestEnum1)reader.GetByte(1),
                B = reader.GetInt32(2),
                C = reader.GetInt32(3),
            };

            list.Add(row);
        }

        return list;
    }
}

internal class TestTable1Model
{
    public required int Id { get; set; }

    public required TestEnum1 A { get; set; }

    public required int B { get; set; }

    public required int C { get; set; }

    public override string ToString() => $"Id: {Id}, A: {A}, B: {B:X}, C: {C:X}";
}

internal enum TestEnum1 : sbyte
{
    None = 0,
    X = 1,
    Y = 2,
    Z = 3,
};
```

test

### DIFF
--- a/src/MySqlConnector/MySqlParameter.cs
+++ b/src/MySqlConnector/MySqlParameter.cs
@@ -835,7 +835,45 @@ public sealed class MySqlParameter : DbParameter, IDbDataParameter, ICloneable
 		}
 		else if (Value is Enum)
 		{
-			writer.Write(Convert.ToInt32(Value, CultureInfo.InvariantCulture));
+			var enumType = Value.GetType();
+			var enumUnderlyingType = enumType.GetEnumUnderlyingType();
+
+			if (enumUnderlyingType == typeof(byte))
+			{
+				writer.Write((byte) Value);
+			}
+			else if (enumUnderlyingType == typeof(sbyte))
+			{
+				writer.Write(unchecked((byte) (sbyte) Value));
+			}
+			else if (enumUnderlyingType == typeof(short))
+			{
+				writer.Write(unchecked((ushort) (short) Value));
+			}
+			else if (enumUnderlyingType == typeof(ushort))
+			{
+				writer.Write((ushort) Value);
+			}
+			else if (enumUnderlyingType == typeof(int))
+			{
+				writer.Write((int) Value);
+			}
+			else if (enumUnderlyingType == typeof(uint))
+			{
+				writer.Write((uint) Value);
+			}
+			else if (enumUnderlyingType == typeof(long))
+			{
+				writer.Write(unchecked((ulong) (long) Value));
+			}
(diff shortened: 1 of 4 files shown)


bgrainger:
> This appears to be because the column type and the actual size of data used in the SingleCommandPayloadCreator.WriteBinaryParameters function are different.

This is the bug. More concretely, the `parameter_type` value sent earlier in the `COM_STMT_EXECUTE` packet doesn't match the format of the data that is sent later.

This PR doesn't fix the underlying problem for the general case where the user uses an enum whose underlying type doesn't match the column definition. Thanks for bringing this to my attention; I have an idea of how this might be fixed.

(This may also have been made worse by #1384 which caused the `MySqlDbType` property to be ignored.)

bgrainger:
Thanks for the PR! Fixed in [2.3.2](https://github.com/mysql-net/MySqlConnector/releases/tag/2.3.2).

daebo01:
thanks for checking

PR #1446, opened 2024-02-05
[Docs] Update migration guide for Connector/NET

rbanks54:
I've updated the docs as there was a Connector/NET issue (70111) resolved in a recent update that wasn't reflected in the migration guide.

I also found it difficult looking through the list of bugs to see which were open/closed given all the strikethrough content. As such I reordered the list to make it easier to see; grouping the open and closed items whilst retaining the issue number ordering.

Finally, I updated the migration docs for a handful of other issues to indicate which C/NET version a problem was fixed in. Coming from an older C/NET version, I would have found it quite useful to know which bug fixes weren't relevant and which ones I needed to pay attention to.

### DIFF
--- a/docs/content/tutorials/migrating-from-connector-net.md
+++ b/docs/content/tutorials/migrating-from-connector-net.md
@@ -91,7 +91,7 @@ MySqlConnector has some different default connection string options:
     <td><code>LoadBalance</code></td>
     <td>Default is <code>RoundRobin</code></td>
     <td>(not configurable, effective default is <code>FailOver</code>)</td>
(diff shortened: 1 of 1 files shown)


bgrainger:
Thanks for the contribution to the docs!

bgrainger:
> found it difficult looking through the list of bugs to see which were open/closed given all the strikethrough content. As such I reordered the list to make it easier to see; grouping the open and closed items whilst retaining the issue number ordering.

Now that the fixed-first bugs are in their own section (great idea BTW), the strikethrough feels very unnecessary (and makes the text difficult to read) so I just removed it all: https://github.com/mysql-net/MySqlConnector/commit/dd518bf99fa385104111a1de7be9f28b2db39e9f

Thanks again.

PR #1460, opened 2024-03-14
Mention MariaDB prominently in the README

dlenski:
While several of the detailed documentation pages _do_ mention MariaDB, the `README` does not mention it explicitly.  

This might cause MariaDB users to overlook this package as a good solution for connecting to their databases from .NET applications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,8 @@
 [![NuGet](https://img.shields.io/nuget/vpre/MySqlConnector.svg)](https://www.nuget.org/packages/MySqlConnector/)
 
 This is an [ADO.NET](https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/) data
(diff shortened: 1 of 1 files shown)


dlenski:
I and my colleagues at AWS learned about this package thanks to @bgrainger's very useful input in the discussion over at 
https://github.com/mariadb-corporation/mariadb-connector-c/pull/227#issuecomment-1623623646

bgrainger:
Thanks; good idea.

This package is basically recommended [in the MariaDB KB](https://mariadb.com/kb/en/net-connector/), and now that Oracle has essentially [broken MariaDB compatibility](https://bugs.mysql.com/bug.php?id=109331) in MySql.Data, it feels pretty reasonable to promote MariaDB compatibility as a feature.

Issue #1236, opened 2022-11-01
Don't issue `SET NAMES` for MariaDB


bgrainger:
When MySqlConnector performs a connection reset, it also issues a `SET NAMES` statement (see https://github.com/mysql-net/MySqlConnector/issues/66) to work around https://bugs.mysql.com/bug.php?id=97633.

MariaDB doesn't have this bug: https://jira.mariadb.org/browse/MDEV-18281.

Although we currently have pipelining (https://github.com/mysql-net/MySqlConnector/pull/1089) which avoids multiple round trips, it may still improve efficiency slightly to detect MariaDB and skip the `SET NAMES` packet.

PR #1494, opened 2024-07-12
Avoids SET NAMES commands if not needed

rusher:
For MariaDB since 15.1 (https://jira.mariadb.org/browse/MDEV-31609) session variables tracked (session_track_system_variables) are sent in initial OK_Packet. This PR permits to avoids  initial query SET NAMES command, since charset is really know, defaulting to server corresponding default collation.

### DIFF
--- a/src/MySqlConnector/Core/Context.cs
+++ b/src/MySqlConnector/Core/Context.cs
@@ -1,0 +1,29 @@
+using MySqlConnector.Protocol;
+
+namespace MySqlConnector.Core;
+
+internal sealed class Context
+{
+	public Context(ProtocolCapabilities protocolCapabilities, string? database, int connectionId)
+	{
+		SupportsDeprecateEof = (protocolCapabilities & ProtocolCapabilities.DeprecateEof) != 0;
+		SupportsCachedPreparedMetadata = (protocolCapabilities & ProtocolCapabilities.MariaDbCacheMetadata) != 0;
+		SupportsQueryAttributes = (protocolCapabilities & ProtocolCapabilities.QueryAttributes) != 0;
+		SupportsSessionTrack = (protocolCapabilities & ProtocolCapabilities.SessionTrack) != 0;
+		ConnectionId = connectionId;
+		Database = database;
+		m_initialDatabase = database;
+	}
+
+	public bool SupportsDeprecateEof { get; }
+	public bool SupportsQueryAttributes { get; }
+	public bool SupportsSessionTrack { get; }
+	public bool SupportsCachedPreparedMetadata { get; }
+	public string? ClientCharset { get; set; }
+
+	public string? Database { get; set; }
+	private readonly string? m_initialDatabase;
+	public bool IsInitialDatabase() => string.Equals(m_initialDatabase, Database, StringComparison.Ordinal);
+
+	public int ConnectionId { get; set; }
+}
--- a/src/MySqlConnector/Core/ResultSet.cs
+++ b/src/MySqlConnector/Core/ResultSet.cs
@@ -38,7 +38,7 @@ internal sealed class ResultSet(MySqlDataReader dataReader)
 				var firstByte = payload.HeaderByte;
 				if (firstByte == OkPayload.Signature)
 				{
-					var ok = OkPayload.Create(payload.Span, Session.SupportsDeprecateEof, Session.SupportsSessionTrack);
+					var ok = OkPayload.Create(payload.Span, Session.Context);
 
 					// if we've read a result set header then this is a SELECT statement, so we shouldn't overwrite RecordsAffected
 					// (which should be -1 for SELECT) unless the server reports a non-zero count
@@ -48,8 +48,6 @@ internal sealed class ResultSet(MySqlDataReader dataReader)
 					if (ok.LastInsertId != 0)
 						Command?.SetLastInsertedId((long) ok.LastInsertId);
 					WarningCount = ok.WarningCount;
-					if (ok.NewSchema is not null)
-						Connection.Session.DatabaseOverride = ok.NewSchema;
 					m_columnDefinitions = default;
 					State = (ok.ServerStatus & ServerStatus.MoreResultsExist) == 0
 						? ResultSetState.NoMoreData
@@ -109,7 +107,7 @@ internal sealed class ResultSet(MySqlDataReader dataReader)
 				}
 				else
 				{
-					var columnCountPacket = ColumnCountPayload.Create(payload.Span, Session.SupportsCachedPreparedMetadata);
+					var columnCountPacket = ColumnCountPayload.Create(payload.Span, Session.Context.SupportsCachedPreparedMetadata);
 					var columnCount = columnCountPacket.ColumnCount;
 					if (!columnCountPacket.MetadataFollows)
 					{
@@ -132,7 +130,7 @@ internal sealed class ResultSet(MySqlDataReader dataReader)
 						m_columnDefinitions = m_columnDefinitionPayloadCache.AsMemory(0, columnCount);
 
 						// if the server supports metadata caching but has re-sent it, something has changed since last prepare/execution and we need to update the columns
-						var preparedColumns = Session.SupportsCachedPreparedMetadata ? DataReader.LastUsedPreparedStatement?.Columns : null;
+						var preparedColumns = Session.Context.SupportsCachedPreparedMetadata ? DataReader.LastUsedPreparedStatement?.Columns : null;
 
 						for (var column = 0; column < columnCount; column++)
 						{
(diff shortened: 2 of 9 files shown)


src/MySqlConnector/Protocol/Payloads/OkPayload.cs
```diff
@@ -1,6 +1,7 @@
+using System.Globalization;
 using System.Text;
+using MySqlConnector.Core;
 using MySqlConnector.Protocol.Serialization;
-using MySqlConnector.Utilities;
```
bgrainger, line 3: This `using` is required to bring in backwards-compatibility extension methods (for older frameworks).

bgrainger:
The `SET NAMES` is only there to work around a MySQL Server bug: https://github.com/mysql-net/MySqlConnector/issues/1236. It's possible that if we detect MariaDB we could just assume that its implementation of COM_RESET_CONNECTION actually works, and skip the `SET NAMES` without the unnecessary effort of parsing system variables just to confirm that it's actually doing the right thing?

rusher:
i think, this is a good default command. Yes, MariaDB doesn't have the com reset issue, but ensuring using utf8 exchanges in all connector exchange is a good thing (avoiding errors and for security)

src/MySqlConnector/Protocol/Payloads/OkPayload.cs
rusher: changed

bgrainger:
I misunderstood the point of this change; I now see that it handles system variables in the initial handshake, and thus avoids a `SET NAMES` query if the server already has the right client charset. 👍

I'll leave a little more PR feedback later, but hope to get this merged in soon.

src/MySqlConnector/Core/Context.cs
```diff
@@ -0,0 +1,29 @@
...
+		m_initialDatabase = database;
+	}
+
+	public bool SupportsDeprecateEof { get; }
+	public bool SupportsQueryAttributes { get; }
+	public bool SupportsSessionTrack { get; }
+	public bool SupportsCachedPreparedMetadata { get; }
+	public string? ClientCharset { get; set; }
+
+	public string? Database { get; set; }
```
bgrainger, line 24: I see the appeal in moving the connection-context-specific data into this new class, but I think it makes it harder to understand to have state spread across `ServerSession` and now `Context`. I'd rather this class be an immutable collection of data that's passed into methods that need to query the current capabilities.

I also think it's a little "unpredictable" to have these properties updated without any notification that it's happening. (The old way wasn't great, either, in that code had to read the `OkPayload.NewSchema` and act on it, and if it didn't, that was a bug.) I also don't like the OkPayload parsing code having the side-effect of updating this mutable state, instead of just being a property-bag/DTO of all the data that was read out of the OK packet.

Made it immutable here: https://github.com/mysql-net/MySqlConnector/pull/1497/commits/77ad1910007923da2a6f12244eaf01eb30c77dcd

src/MySqlConnector/Logging/Log.cs
```diff
@@ -47,7 +47,7 @@ internal static partial class Log
...
 
 	[LoggerMessage(EventIds.SessionMadeConnection, LogLevel.Debug, "Session {SessionId} made connection; server version {ServerVersion}; connection ID {ConnectionId}; supports: compression {SupportsCompression}, attributes {SupportsAttributes}, deprecate EOF {SupportsDeprecateEof}, cached metadata {SupportsCachedMetadata}, SSL {SupportsSsl}, session track {SupportsSessionTrack}, pipelining {SupportsPipelining}, query attributes {SupportsQueryAttributes}")]
-	public static partial void SessionMadeConnection(ILogger logger, string sessionId, string serverVersion, int connectionId, bool supportsCompression, bool supportsAttributes, bool supportsDeprecateEof, bool supportsCachedMetadata, bool supportsSsl, bool supportsSessionTrack, bool supportsPipelining, bool supportsQueryAttributes);
+	public static partial void SessionMadeConnection(ILogger logger, string sessionId, string serverVersion, long connectionId, bool supportsCompression, bool supportsAttributes, bool supportsDeprecateEof, bool supportsCachedMetadata, bool supportsSsl, bool supportsSessionTrack, bool supportsPipelining, bool supportsQueryAttributes);
```
bgrainger, line 50: Was there a plan to make this variable a `long` throughout the rest of the code? Does MariaDB use 64-bit connection IDs?

https://github.com/mysql-net/MySqlConnector/pull/1494/files#diff-abcf6fc970e193538504a795eb54ea931b52864f45ef4fa0e8040ee1273dac9aR96 still converts to an Int32.

Reverted to `int` here: https://github.com/mysql-net/MySqlConnector/pull/1497/commits/b19f1ee1eed604983c87e42c6232813fbb5ecc76.

src/MySqlConnector/Core/ResultSet.cs
```diff
@@ -38,7 +38,7 @@ public async Task ReadResultSetHeaderAsync(IOBehavior ioBehavior)
 				var firstByte = payload.HeaderByte;
 				if (firstByte == OkPayload.Signature)
 				{
-					var ok = OkPayload.Create(payload.Span, Session.SupportsDeprecateEof, Session.SupportsSessionTrack);
+					var ok = OkPayload.Create(payload.Span, Session.Context);
```
bgrainger, line 41: I like bundling these together, but it could be done using a new interface that `ServerSession` implements. That would avoid allocating a second object every time a `ServerSession` is created. Changed here: https://github.com/mysql-net/MySqlConnector/pull/1497/commits/8466783f5647063092fc54dc47a43e2a3e19e16a.

src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -454,30 +452,26 @@ public async Task DisposeAsync(IOBehavior ioBehavior, CancellationToken cancella
 			}
 
 			ServerVersion = new(initialHandshake.ServerVersion);
-			ConnectionId = initialHandshake.ConnectionId;
+			Context = new Context(initialHandshake.ProtocolCapabilities, cs.Database, initialHandshake.ConnectionId);
```
bgrainger, line 455: By implementing an interface on `ServerSession` we also avoid this second `Context` allocation every time a session is connected (so reduced from three back down to one object).

src/MySqlConnector/Protocol/Payloads/OkPayload.cs
```diff
@@ -79,7 +75,28 @@ internal sealed class OkPayload
...
 							case SessionTrackKind.Schema:
-								newSchema = Encoding.UTF8.GetString(reader.ReadLengthEncodedByteString());
+								context.Database = Encoding.UTF8.GetString(reader.ReadLengthEncodedByteString());
+								break;
+
+							case SessionTrackKind.SystemVariables:
+								var systemVariableOffset = reader.Offset + dataLength;
+								do
+								{
+									var variableSv = Encoding.ASCII.GetString(reader.ReadLengthEncodedByteString());
```
bgrainger, line 85: This is potentially a lot of string allocations for server variables we don't even read. We can perform a string comparison on the UTF-8 data by using `ReadOnlySpan<byte>` and `""u8` string literals: https://github.com/mysql-net/MySqlConnector/pull/1497/commits/3326a0ec06a8b32776271a7c4de103e8088b422a.

src/MySqlConnector/Protocol/Payloads/OkPayload.cs
```diff
@@ -79,7 +75,28 @@ internal sealed class OkPayload
...
+									var lenSv = reader.ReadLengthEncodedIntegerOrNull();
+									var valueSv = lenSv == -1
+										? null
+										: Encoding.ASCII.GetString(reader.ReadByteString(lenSv));
+									switch (variableSv)
+									{
+										case "character_set_client":
```
bgrainger, line 92: This is just one of the three possible connection-level character set variables, including `character_set_connection` and `character_set_results`. It seems prudent to verify that all three have the expected value (since the server will send them to us). Updated here: https://github.com/mysql-net/MySqlConnector/pull/1497/commits/3f50c14f734b23025d98f0d090416cec2a1940cf.

src/MySqlConnector/Protocol/Payloads/OkPayload.cs
```diff
@@ -79,7 +75,28 @@ internal sealed class OkPayload
...
+											context.ClientCharset = valueSv;
+											break;
+										case "connection_id":
+											context.ConnectionId = Convert.ToInt32(valueSv, CultureInfo.InvariantCulture);
```
bgrainger, line 96: We can also use `Utf8Parser` to convert UTF-8 bytes directly to an `int` without allocating a temporary string.

src/MySqlConnector/Logging/Log.cs
rusher: Sorry, my fault, i began changing that and revert it, but not completly. 
technically, mariadb use internally long, but using the integer truncated value on initial handshake works ok (see https://jira.mariadb.org/browse/MDEV-15089). To be honest, i've only heard about wikipedia having an issue with that, creating more than 2147483647 connections without restart is an achievement !

src/MySqlConnector/Protocol/Payloads/OkPayload.cs
rusher: good point

PR #1497, opened 2024-07-17
Avoid SET NAMES commands if not needed

bgrainger:
Based on #1494.


bgrainger:
Fixed in [2.4.0](https://github.com/mysql-net/MySqlConnector/releases/tag/2.4.0).

PR #1499, opened 2024-07-20
Changing Permit redirection

rusher:
Permit redirection, based on mariadb implementation https://jira.mariadb.org/browse/MDEV-15935: first OK_Packet can contain variable `redirect_url` using "mariadb/mysql://[{user}[:{password}]@]{host}[:{port}]/[{db}[?{opt1}={value1}[&{opt2}={value2}]]]']" format.

### DIFF
--- a/src/MySqlConnector/Logging/Log.cs
+++ b/src/MySqlConnector/Logging/Log.cs
@@ -402,26 +402,26 @@ internal static partial class Log
 	[LoggerMessage(EventIds.FoundSessionToCleanUp, LogLevel.Debug, "Pool {PoolId} found session {SessionId} to clean up")]
 	public static partial void FoundSessionToCleanUp(ILogger logger, int poolId, string sessionId);
 
-	[LoggerMessage(EventIds.HasServerRedirectionHeader, LogLevel.Trace, "Session {SessionId} has server redirection header {Header}")]
-	public static partial void HasServerRedirectionHeader(ILogger logger, string sessionId, string header);
+	[LoggerMessage(EventIds.HasServerRedirectionHeader, LogLevel.Trace, "{poolPrefix}Session {SessionId} has server redirection header {Header}")]
+	public static partial void HasServerRedirectionHeader(ILogger logger, string poolPrefix, string sessionId, string header);
 
-	[LoggerMessage(EventIds.ServerRedirectionIsDisabled, LogLevel.Trace, "Pool {PoolId} server redirection is disabled; ignoring redirection")]
-	public static partial void ServerRedirectionIsDisabled(ILogger logger, int poolId);
+	[LoggerMessage(EventIds.ServerRedirectionIsDisabled, LogLevel.Trace, "{poolPrefix}server redirection is disabled; ignoring redirection")]
+	public static partial void ServerRedirectionIsDisabled(ILogger logger, string poolPrefix);
 
-	[LoggerMessage(EventIds.OpeningNewConnection, LogLevel.Debug, "Pool {PoolId} opening new connection to {Host}:{Port} as {User}")]
-	public static partial void OpeningNewConnection(ILogger logger, int poolId, string host, int port, string user);
+	[LoggerMessage(EventIds.OpeningNewConnection, LogLevel.Debug, "{poolPrefix}opening new connection to {Host}:{Port} as {User}")]
+	public static partial void OpeningNewConnection(ILogger logger, string poolPrefix, string host, int port, string user);
 
-	[LoggerMessage(EventIds.FailedToConnectRedirectedSession, LogLevel.Information, "Pool {PoolId} failed to connect redirected session {SessionId}")]
-	public static partial void FailedToConnectRedirectedSession(ILogger logger, Exception ex, int poolId, string sessionId);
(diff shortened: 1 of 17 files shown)


rusher:
This is done for either pool and single connection, contrary to previous implementation. I don't know why there was a restriction like this. Logger messages from previous pool/single connection are keeped, but it might be a good idea to have a simplier implemenation

src/MySqlConnector/Logging/Log.cs
```diff
@@ -402,26 +402,26 @@ internal static partial class Log
 	[LoggerMessage(EventIds.FoundSessionToCleanUp, LogLevel.Debug, "Pool {PoolId} found session {SessionId} to clean up")]
 	public static partial void FoundSessionToCleanUp(ILogger logger, int poolId, string sessionId);
 
-	[LoggerMessage(EventIds.HasServerRedirectionHeader, LogLevel.Trace, "Session {SessionId} has server redirection header {Header}")]
-	public static partial void HasServerRedirectionHeader(ILogger logger, string sessionId, string header);
+	[LoggerMessage(EventIds.HasServerRedirectionHeader, LogLevel.Trace, "{poolPrefix}Session {SessionId} has server redirection header {Header}")]
+	public static partial void HasServerRedirectionHeader(ILogger logger, string poolPrefix, string sessionId, string header);
```
bgrainger, line 406: The log field names are significant, and are used for "structured logging" within Microsoft.Extensions.Logging; see https://learn.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-8.0#log-message-template.

We can't just change them for better code reuse (thinking of them like string concatenation).

bgrainger:
> This is done for either pool and single connection, contrary to previous implementation. I don't know why there was a restriction like this.

I don't remember why, either. Based on the commit message for https://github.com/mysql-net/MySqlConnector/commit/48c67547c756560d9c921c80adc0a2f20b6ca156 it looks like a bugfix introduced this change, and before that redirection was implemented for pooled and non-pooled.

tests/IntegrationTests/RedirectionTests.cs
```diff
@@ -0,0 +1,200 @@
+using System.Globalization;
+using System.Net;
+using System.Net.Sockets;
+
+namespace IntegrationTests;
+
+public class RedirectionTests : IClassFixture<DatabaseFixture>, IDisposable
```
bgrainger, line 7: Thanks for adding tests!

tests/IntegrationTests/RedirectionTests.cs
```diff
@@ -0,0 +1,200 @@
...
+		// wait for proxy to launch
+		Thread.Sleep(50);
+		var csb = AppConfig.CreateConnectionStringBuilder();
+		var initialServer = csb.Server;
+		var initialPort = csb.Port;
+		var permitRedirection = true;
+		try
+		{
+			m_database.Connection.Execute(
+				$"set @@global.redirect_url=\"mariadb://{initialServer}:{initialPort}\"");
```
bgrainger, line 35: Looks like this fails with MariaDB 11.4 since it does accept this global variable, but doesn't include it in the initial OK packet.

bgrainger: Usually I handle server-specific functionality by adding a new enum to `ServerFeatures` and skipping the test based on whether that's set in `config.json`.

bgrainger:
I'm getting errors in the `ConnectTimeout` test locally; I want to dig into that more first before reviewing this further or merging it.

src/MySqlConnector/MySqlConnection.cs
```diff
@@ -1062,22 +1064,10 @@ private async ValueTask<ServerSession> CreateSessionAsync(ConnectionPool? pool,
 				// only "fail over" and "random" load balancers supported without connection pooling
 				var loadBalancer = connectionSettings.LoadBalance == MySqlLoadBalance.Random && connectionSettings.HostNames!.Count > 1 ?
 					RandomLoadBalancer.Instance : FailOverLoadBalancer.Instance;
-
-				var session = new ServerSession(m_logger)
-				{
-					OwningConnection = new WeakReference<MySqlConnection>(this),
-				};
+				var session = await ServerSession.ConnectAndRedirectAsync(() => new ServerSession(m_logger), m_logger, null, connectionSettings, loadBalancer, this, null, startingTimestamp, null, actualIOBehavior, cancellationToken).ConfigureAwait(false);
```
bgrainger, line 1067:
```suggestion
				var session = await ServerSession.ConnectAndRedirectAsync(() => new ServerSession(m_logger), m_logger, null, connectionSettings, loadBalancer, this, null, startingTimestamp, null, actualIOBehavior, connectToken).ConfigureAwait(false);
```

See line 1073 below.

src/MySqlConnector/MySqlConnection.cs
```diff
@@ -628,6 +628,8 @@ public override string ConnectionString
 		}
 	}
 
+	public string? SessionConnectionString => m_session?.ConnectionString;
```
bgrainger, line 631: Is this just for tests? We shouldn't add a new public property just to make testing easier; that will imply a commitment to support this.

bgrainger (approved):

bgrainger:
Fixed in [2.4.0](https://github.com/mysql-net/MySqlConnector/releases/tag/2.4.0).

PR #1500, opened 2024-07-26
Using TLS without configuration

rusher:
Connector implementation of https://jira.mariadb.org/browse/MDEV-31855

Since MariaDB 11.4.1, TLS use has greatly been simplified.
Connector side doesn't require TLS configuration anymore, even with self-signed certificates.

Connectors can now validate ssl certificates using client password (using seed and server certificate SHA256 thumbprint).

limitations:
* only possible when using mysql_native_password/client_ed25519 authentication
* password is required

see https://mariadb.org/mission-impossible-zero-configuration-ssl/

### DIFF
--- a/src/MySqlConnector/Authentication/IAuthenticationPlugin.cs
+++ b/src/MySqlConnector/Authentication/IAuthenticationPlugin.cs
@@ -19,14 +19,20 @@ public interface IAuthenticationPlugin
 	/// Method Switch Request Packet</a>.</param>
 	/// <returns>The authentication response.</returns>
 	byte[] CreateResponse(string password, ReadOnlySpan<byte> authenticationData);
+}
 
+/// <summary>
+/// <see cref="IAuthenticationPlugin2"/> is an extension to <see cref="IAuthenticationPlugin"/> that returns a hash of the client's password.
+/// </summary>
+public interface IAuthenticationPlugin2 : IAuthenticationPlugin
+{
 	/// <summary>
-	/// create password hash for fingerprint verification
+	/// Hashes the client's password (e.g., for TLS certificate fingerprint verification).
 	/// </summary>
 	/// <param name="password">The client's password.</param>
 	/// <param name="authenticationData">The authentication data supplied by the server; this is the <code>auth method data</code>
 	/// from the <a href="https://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::AuthSwitchRequest">Authentication
 	/// Method Switch Request Packet</a>.</param>
-	/// <returns>password hash</returns>
+	/// <returns>The authentication-method-specific hash of the client's password.</returns>
 	byte[] CreatePasswordHash(string password, ReadOnlySpan<byte> authenticationData);
 }
--- a/src/MySqlConnector/Authentication/IAuthenticationPlugin.cs
+++ b/src/MySqlConnector/Authentication/IAuthenticationPlugin.cs
@@ -19,4 +19,14 @@ public interface IAuthenticationPlugin
 	/// Method Switch Request Packet</a>.</param>
 	/// <returns>The authentication response.</returns>
 	byte[] CreateResponse(string password, ReadOnlySpan<byte> authenticationData);
+
+	/// <summary>
+	/// create password hash for fingerprint verification
+	/// </summary>
+	/// <param name="password">The client's password.</param>
+	/// <param name="authenticationData">The authentication data supplied by the server; this is the <code>auth method data</code>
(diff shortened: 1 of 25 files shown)


tests/IntegrationTests/Attributes.cs
```diff
@@ -12,6 +12,11 @@ public SkippableFactAttribute(ServerFeatures serverFeatures)
 	{
 	}
 
+	public SkippableFactAttribute(ServerFeatures[] serverFeatureList)
```
bgrainger, line 15: `ServerFeatures` is a `[Flags]` enum, so we can use bitwise OR here to specify more than one feature.

src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -567,6 +609,57 @@ public async Task DisposeAsync(IOBehavior ioBehavior, CancellationToken cancella
...
+				passwordHashResult = ed25519Plugin!.CreatePasswordHash(password, challenge);
+				break;
+
+			default:
+				return false;
+		}
+
+		Span<byte> combined = stackalloc byte[32 + (challenge.Length - 1) + passwordHashResult.Length];
+		passwordHashResult.CopyTo(combined);
+		challenge.CopyTo(combined[passwordHashResult.Length..]);
```
bgrainger, line 645: Feels wrong that we're computing the length using `challenge.Length - 1` but copying the entirety of `challenge`. I'm assuming there's a NULL terminator byte here; but if so, we should strip it in the caller (and avoid doing math on the length here).

src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -567,6 +609,57 @@ public async Task DisposeAsync(IOBehavior ioBehavior, CancellationToken cancella
...
+#if NET5_0_OR_GREATER
+		hashBytes = SHA256.HashData(combined);
+#else
+		using (var sha256 = SHA256.Create())
+		{
+			hashBytes = sha256.ComputeHash(combined.ToArray());
+		}
+#endif
+
+		var clientGeneratedHash = hashBytes.Aggregate(string.Empty, (str, hashByte) => str + hashByte.ToString("X2", CultureInfo.InvariantCulture));
```
bgrainger, line 658: This is a very inefficient way to convert a byte array to a hex string. (I'd probably go the other way and convert the server hash to a byte array, then compare.)

src/MySqlConnector/Authentication/IAuthenticationPlugin.cs
```diff
@@ -19,4 +19,14 @@ public interface IAuthenticationPlugin
...
+
+	/// <summary>
+	/// create password hash for fingerprint verification
+	/// </summary>
+	/// <param name="password">The client's password.</param>
+	/// <param name="authenticationData">The authentication data supplied by the server; this is the <code>auth method data</code>
+	/// from the <a href="https://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::AuthSwitchRequest">Authentication
+	/// Method Switch Request Packet</a>.</param>
+	/// <returns>password hash</returns>
+	byte[] CreatePasswordHash(string password, ReadOnlySpan<byte> authenticationData);
```
bgrainger, line 31: Adding an interface member is a breaking change; we probably need to introduce a new interface to support this. 

(Not that it's very likely that anyone has ever implemented this interface...)

src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -528,6 +528,43 @@ public async Task DisposeAsync(IOBehavior ioBehavior, CancellationToken cancella
...
+						// now throwing SSL exception depending on m_rcbPolicyErrors
+						ShutdownSocket();
+						HostName = "";
+						lock (m_lock) m_state = State.Failed;
+						MySqlException ex;
+						switch (m_rcbPolicyErrors)
+						{
+							case SslPolicyErrors.RemoteCertificateNotAvailable:
+								// impossible
+								ex = new MySqlException(MySqlErrorCode.UnableToConnectToHost, "SSL not validated, no remote certificate available");
```
bgrainger, line 551: This probably changes the `InnerException` that's being caught and rethrown? I want to compare this to the current behaviour.

bgrainger: The simplest thing is probably to expose `m_rcbPolicyErrors` on an `AuthenticationException` used as the `.InnerException`.

bgrainger (approved):

bgrainger:
Fixed in [2.4.0](https://github.com/mysql-net/MySqlConnector/releases/tag/2.4.0); thanks for the contribution!

PR #1515, opened 2024-10-06
Add ConnectionOpenedCallback

### LINKED ISSUE #1508: Add (flexible) hook for initial connection opening event
**Is your feature request related to a problem? Please describe.**
MySQL allows for configuration of various behavior-changing settings on a per-connection/session basis, which is done through queries. This includes for example setting collation (`SET NAMES 'utf8mb4';`), setting timezone (`SET time_zone='Europe/Amsterdam';`) or locale support (`SET lc_time_names = 'nl_NL';`). Configuring these preferably should be done once at connection startup to ensure they are used throughout the entire lifetime of the connection and cause as little overhead as possible.

In my usecase I would like to set the (connection) timezone, however I think the problem can be generalized to any of such settings.

**Describe the solution you'd like**
Have an event hook that allows for safely and easily executing option-setting queries once after the first opening of a connection. This could be a generic event that allows arbitrary code execution, or could be some more subtle mechanism that only allows passing e.g. a list of statements to be executed once.

**Describe alternatives you've considered**
- In #487 there was some discussion about possibly adding this to `ConnectionString`; given the wide variety of possibly interesting options, I agree that adding this here is probably too much overhead; however it still can be considered.
- ~~I've tried to implement this using the `StateChange` event, however I run into troubles with some race conditions where the next query already seems to be trying to execute when the timezone setting query is still active, causing exceptions. Implementation roughly looks like this:~~
As a workaround, something similar to this can currently be used
```cs
connection.StateChange += (_, args) =>
    {
        // Only consider transition from Closed/Connecting to Open to denote the 'start' of a new connection
        if (args.CurrentState == ConnectionState.Open && (args.OriginalState is ConnectionState.Closed or ConnectionState.Connecting))
            {
                conn.Execute("SET time_zone = 'UTC'");
            }
    };
```
- Manually opening the connections at some point in the application logic, followed by setting the aforementioned settings with queries. I can't however manage to let this play nicely with passing the connection to `Dapper`.

**Additional context**
- There already seems to be some kind of implementation for this regarding the collation: https://github.com/mysql-net/MySqlConnector/blob/master/src/MySqlConnector/Core/ServerSession.cs#L580. Maybe this can be generalized somehow?

**Final note**
If there are any suggestions/alternative approaches on how to subtly set `time_zone` or other settings with existing code, please let me know!

bgrainger:
Fixes #1508.

This differs from the existing `StateChanged` event in that:

- it supports an async callback
- it's only invoked when a connection is opened
- it provides information about new vs existing and whether the connection was reset

Opening for API review. Needs tests before it can be merged.

### DIFF
--- a/src/MySqlConnector/Core/ServerSession.cs
+++ b/src/MySqlConnector/Core/ServerSession.cs
@@ -65,10 +65,12 @@ internal sealed partial class ServerSession : IServerCapabilities
 	public bool ProcAccessDenied { get; set; }
 	public ICollection<KeyValuePair<string, object?>> ActivityTags => m_activityTags;
 	public MySqlDataReader DataReader { get; set; }
+	public MySqlConnectionOpenedConditions Conditions { get; private set; }
 
 	public ValueTask ReturnToPoolAsync(IOBehavior ioBehavior, MySqlConnection? owningConnection)
 	{
 		Log.ReturningToPool(m_logger, Id, Pool?.Id ?? 0);
+		Conditions = MySqlConnectionOpenedConditions.None;
 		LastReturnedTimestamp = Stopwatch.GetTimestamp();
 		if (Pool is null)
 			return default;
@@ -414,6 +416,7 @@ internal sealed partial class ServerSession : IServerCapabilities
 				}
 			}
 
+			Conditions = MySqlConnectionOpenedConditions.New;
 			var connected = cs.ConnectionProtocol switch
 			{
 				MySqlConnectionProtocol.Sockets => await OpenTcpSocketAsync(cs, loadBalancer ?? throw new ArgumentNullException(nameof(loadBalancer)), activity, ioBehavior, cancellationToken).ConfigureAwait(false),
@@ -747,6 +750,7 @@ internal sealed partial class ServerSession : IServerCapabilities
 	public async Task<bool> TryResetConnectionAsync(ConnectionSettings cs, MySqlConnection connection, IOBehavior ioBehavior, CancellationToken cancellationToken)
 	{
 		VerifyState(State.Connected);
+		Conditions |= MySqlConnectionOpenedConditions.Reset;
 
 		try
 		{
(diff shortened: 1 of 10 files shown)


src/MySqlConnector/MySqlConnection.cs
```diff
@@ -917,6 +920,7 @@ internal void Cancel(ICancellableCommand command, int commandId, bool isCancel)
 
 			using var connection = CloneWith(csb.ConnectionString);
 			connection.m_connectionSettings = connectionSettings;
+			connection.ConnectionOpenedCallback = null; // clear the callback because the user doesn't need to execute any setup logic on this connection
```
ejball, line 923: How do we know that?

src/MySqlConnector/Core/ServerSession.cs
```diff
@@ -747,6 +750,7 @@ public static async ValueTask<ServerSession> ConnectAndRedirectAsync(ILogger con
 	public async Task<bool> TryResetConnectionAsync(ConnectionSettings cs, MySqlConnection connection, IOBehavior ioBehavior, CancellationToken cancellationToken)
 	{
 		VerifyState(State.Connected);
+		Conditions |= MySqlConnectionOpenedConditions.Reset;
```
ejball, line 753: This doesn't seem "mutually exclusive with New" but I don't understand the code flow very well.

src/MySqlConnector/MySqlConnection.cs
bgrainger: It's a non-pooled connection that will execute `KILL QUERY` and be closed immediately. This is more of an implementation detail than a user-visible `MySqlConnection`. I can reword the comment.

jnoordsij:
Thanks for this! However I'm struggling to see how this should be used when creating a `MySqlConnection`, given that `ConnectionOpenedCallback` is internal?

bgrainger:
Documentation isn't fully updated yet, but you need to use `MySqlDataSourceBuilder`: https://mysqlconnector.net/api/mysqlconnector/mysqldatasourcebuilder/useconnectionopenedcallback/

`DbDataSource` is the .NET 7+ way of configuring DbConnection objects, and will be the preferred API for MySqlConnector going forward.

Issue #1521, opened 2024-10-21
Unable to get access to table alias name in data reader


AlexKerman:
When used multiple joins to the same table there no way to find out which alias current column belongs to.

`SELECT * FROM Orders o LEFT JOIN Users u1 ON o.CreatedUserId = u1.Id LEFT JOIN Users u2 ON o.UpdatedUserId = u2.Id`

Both users has BaseTableName "Users" in DbColumn, but no fields have "u1" and "u2" aliases. 
ColumnDefinitionPayload class has PhysicalTableName and Table, but Table does not copy into MySqlDbColumn class.

bgrainger:
Fixed in [2.4.0](https://github.com/mysql-net/MySqlConnector/releases/tag/2.4.0); thanks for the contribution!

PR #1522, opened 2024-10-21
Make table alias accessible from MySqlDbColumn

AlexKerman:
This PR make table alias accessible in data reader.
Ref #1521

### DIFF
--- a/src/MySqlConnector/MySqlDbColumn.cs
+++ b/src/MySqlConnector/MySqlDbColumn.cs
@@ -22,6 +22,7 @@ public sealed class MySqlDbColumn : DbColumn
 		BaseColumnName = column.PhysicalName;
 		BaseSchemaName = column.SchemaName;
 		BaseTableName = column.PhysicalTable;
+		TableName = column.Table;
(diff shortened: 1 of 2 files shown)


bgrainger:
Thanks for the contribution! I'll take a look soon. (Mostly just want to see if there's precedent in other ADO.NET providers.)

It would be good to include some tests for this new behaviour.

bgrainger:
Npgsql adds a `TableOID` property: https://github.com/npgsql/npgsql/blob/32eb7097c81e3dc6e3cdc59c2f19592883f00c77/src/Npgsql/Schema/NpgsqlDbColumn.cs#L199-L202

SqlClient doesn't appear to have public `DbColumn`-derived type.

bgrainger:
Thanks for the contribution!

AlexKerman:
> Thanks for the contribution!

Thank you for your research. It very useful for me because I doing postgres support for my tool. Unfortunately I don't know how to test this properly. I had researched test for ProviderType field and didn't fount out how to insert test for my field corectly.

PR #1535, opened 2024-12-16
Add UseProcedureCache option and documentation. Fixes #1175

### LINKED ISSUE #1175: Support disabling stored procedure cache
Based on https://github.com/mysql-net/MySqlConnector/issues/924 and https://github.com/mysql-net/MySqlConnector/issues/1174, there are a small number of users who modify stored procedure definitions in the database at runtime; this can be a problem because MySqlConnector caches the procedure definitions and can invoke them with the wrong parameter types or order.

Connector/NET can disable its procedure cache by setting `Procedure Cache Size = 0` in the connection string. MySqlConnector doesn't support that option.

We could add `Procedure Cache Size` and only support zero vs non-zero (to disable or enable caching, respectively), or add a new `Use Procedure Cache = false` connection string option (default is `true`) to enable users to opt out.

This might have a significant impact on performance, so it could be worth testing whether it's even an option people would want to use. If it's too slow if caching is disabled, other workarounds (such as clearing the pool to get the side-effect of clearing the procedure cache) might be a better solution anyway.

tarasevichvlad:
@bgrainger  What do you think about this changes. This is implementation for second idea in #1175 issue.

Currently without tests.

### DIFF
--- a/docs/content/connection-options.md
+++ b/docs/content/connection-options.md
@@ -451,6 +451,11 @@ These are the other options that MySqlConnector supports. They are set to sensib
     distributed transactions, but may not be compatible with server replication; there are <a href="https://dev.mysql.com/doc/refman/8.0/en/xa-restrictions.html">other limitations</a>.
     When set to <code>false</code>, regular MySQL transactions are used, just like Connector/NET.</td>
   </tr>
+  <tr id="UseProcedureCache">
+    <td>Use Procedure Cache, UseProcedureCache</td>
+    <td>true</td>
+    <td>When <code>false</code> disables the procedure cache</td>
+  </tr>
 </table>
 
 ## Unsupported Options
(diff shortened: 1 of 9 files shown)


bgrainger:
The connection string changes look fine.

I think the change to `FillAsync` isn't appropriate, as we do actually want `CachedProcedure` information for `AlignParamsWithDb` later on. Instead, we might want to make `ConnectionPool.GetProcedureCache` return `null` if procedure caching is disabled. But even then we wouldn't want to cache the procedure (even on the MySqlConnection) beyond the lifetime of the `ExecuteXyz` call. (And we would also need to update logging so it doesn't log incorrect messages about what is happening.)

Sorry I don't have more specific guidance yet; I haven't looked into the details of how to solve this problem.

tarasevichvlad:
@bgrainger 
I took a closer look at the places where parameter caching is used for procedures and made some changes. 
If you have time, please look at additional edits to support this feature.

Issue #1524, opened 2024-11-22
Spammed by 'read-result-set-header' traces in application insights


Tobbeman:
**Is your feature request related to a problem? Please describe.**
Disclaimer; Still learning how to use OpenTelemetry

We recently added OpenTelemetry to our application as well as MySqlConnector as source
Sadly, we are spammed by traces with the message "read-result-set-header".

These seem to be created [here](https://github.com/mysql-net/MySqlConnector/blob/master/src/MySqlConnector/Core/ResultSet.cs#L170)

As a result our application insights are filled:
![image](https://github.com/user-attachments/assets/cd3b408f-7477-4078-8e99-09a8f3f7205d)

Note that all the "traces" are the message in question.

I have not found anyway to filter these, though I might simply be overlooking something

**Describe the solution you'd like**
These traces add little value as far as I can tell, I would like to be able to turn them off.
Perhaps we can add a option to do so?

I'm happy to take a crack at it.

Note: If it is possible to filter these, that would also help me. Currently our filter looks like this and I am unable to change the Events associated with the activity ( read only ):
```c#

internal class FilterProcessor : BaseProcessor<Activity>
{
    public override void OnStart(Activity activity)
    {
        FilterPomeloMySqlConnector(activity);
    }

    void FilterPomeloMySqlConnector(Activity activity) 
    {
        if (activity.Source.Name != "MySqlConnector") {
            return;
        }

        if (activity.DisplayName == "Open")
        {
            activity.IsAllDataRequested = false;
        }
    }
}
```

bgrainger:
How are you feeding the `Activity` events into Application Insights?

"Trace" is supposed to be used for logs; see the data model documentation: https://learn.microsoft.com/en-us/azure/azure-monitor/app/data-model-complete

MySqlConnector is calling `Activity.AddEvent` for this telemetry. That seems like it would be better modeled as an "Event" in Application Insights (or just dropped). I would look into how and why your code is converting events into traces.

vanbukin:
I see something similar.

MySqlConnector 2.4.0
.NET SDK 8.0.404
.NET Runtime 8.0.11
OpenTelemetry 1.10.0
Jaeger 2.0.0
![screencapture-localhost-16686-trace-af868dd767c81485f8ab3f43c20c9ff7-2024-11-25-02_31_56](https://github.com/user-attachments/assets/d4a71216-2d7e-4d59-b1b2-6fccda729a02)

[...]

PR #1538, opened 2025-01-02
Add config for enriching activity with read-result-set-header

Tobbeman:
Thread: https://github.com/mysql-net/MySqlConnector/issues/1524
Inspiration here: https://github.com/open-telemetry/opentelemetry-dotnet-contrib/pull/416/files

First PR, as far as I can tell the only configuration is done via the connection string to I kept using that pattern.
All feedback is appreciated.

### DIFF
--- a/src/MySqlConnector/Core/ConnectionSettings.cs
+++ b/src/MySqlConnector/Core/ConnectionSettings.cs
@@ -147,6 +147,7 @@ internal sealed class ConnectionSettings
 		UseAffectedRows = csb.UseAffectedRows;
 		UseCompression = csb.UseCompression;
 		UseXaTransactions = csb.UseXaTransactions;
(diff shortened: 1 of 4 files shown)


bgrainger:
Thanks for the PR!

> as far as I can tell the only configuration is done via the connection string

I noted on https://github.com/mysql-net/MySqlConnector/issues/1524#issuecomment-2506545847 that Npgsql configures this on NpgsqlDataSourceBuilder (but didn't provide a concrete specification for MySqlConnector). Rather than adding more connection string options, I'd like to go this way instead (probably with a `MySqlConnectorTracingOptions` that can be extended over time with more settings as necessary).

bgrainger:
Going to close this in favour of https://github.com/mysql-net/MySqlConnector/pull/1545.

Tobbeman:
Missed this one;

Thanks a ton for the follow up, looking forward to seeing the update in the wild!

PR #1545, opened 2025-02-02
Add options to configure tracing

### LINKED ISSUE #1524: Spammed by 'read-result-set-header' traces in application insights

bgrainger:
Fixes #1524

### DIFF
--- a/src/MySqlConnector/Core/ResultSet.cs
+++ b/src/MySqlConnector/Core/ResultSet.cs
@@ -166,7 +166,7 @@ internal sealed class ResultSet(MySqlDataReader dataReader)
 						ContainsCommandParameters = true;
 					WarningCount = 0;
 					State = ResultSetState.ReadResultSetHeader;
-					if (DataReader.Activity is { IsAllDataRequested: true })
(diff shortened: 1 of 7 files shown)


bgrainger:
As per the milestone assigned to this PR, the current plan is to add this new API in 2.5.0.

PR #1579, opened 2025-07-18
Implement COM_STMT_SEND_LONG_DATA

djeman:
Hi,

I work for a company that has very old software running on 32-bit systems, which stores documents in a MySQL database as BLOBs. In this context, which we are trying to evolve but it takes time, we are facing the memory limit available for an x86 process when storing large documents.

To address this issue with the latest unupdated software, we have tried your implementation of the COM_STMT_SEND_LONG_DATA command for using streams and avoiding memory overload. After several months of testing, we can confirm that it works as intended; memory is no longer impacted by the size of the documents.

Would it be possible to integrate this into your project?


src/MySqlConnector/MySqlCommand.cs
```diff
@@ -357,15 +359,52 @@ internal async ValueTask<MySqlDataReader> ExecuteReaderAsync(CommandBehavior beh
...
+			for (var i = 0; i < Parameters.Count; i++)
+			{
+				if (Parameters[i].Value is Stream stream)
+				{
+					var writer = new ByteBufferWriter();
+					writer.Write((byte) CommandKind.StatementSendLongData);
+					writer.Write(statements.Statements[0].StatementId);
+					writer.Write((ushort) i);
+
+					var buffer = ArrayPool<byte>.Shared.Rent(ProtocolUtility.MaxPacketSize);
```
bgrainger, line 378: This buffer should be returned.

(Looks like a bug copied from https://github.com/bgrainger/MySqlConnector/commit/b91617cac66a3006e16ac731b5768fa52cc5f6c3.)

src/MySqlConnector/MySqlCommand.cs
```diff
@@ -357,15 +359,52 @@ internal async ValueTask<MySqlDataReader> ExecuteReaderAsync(CommandBehavior beh
...
+					var dataLength = ProtocolUtility.MaxPacketSize - 7;
+
+					var bytesRead = stream.Read(buffer, 0, dataLength);
```
bgrainger, line 381: Could use `ReadAsync` here.

src/MySqlConnector/MySqlCommand.cs
```diff
@@ -357,15 +359,52 @@ internal async ValueTask<MySqlDataReader> ExecuteReaderAsync(CommandBehavior beh
...
+
+						do
+						{
+							bytesRead = stream.Read(buffer, 0, dataLength);
+
+							writer = new ByteBufferWriter();
+							writer.Write(buffer.AsSpan(0, bytesRead));
+							await Connection!.Session.SendReplyAsync(writer.ToPayloadData(), ioBehavior, cancellationToken).ConfigureAwait(false);
+						}
+						while (bytesRead == dataLength);
```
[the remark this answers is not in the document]
djeman: I tried all cases and it works perfect, it's important to send an empty packet if the size is exactly 16MBytes.

https://mariadb.com/docs/server/clients-and-utilities/server-client-software/client-libraries/clientserver-protocol/0-packet#packet-splitting

src/MySqlConnector/MySqlCommand.cs
djeman: Sry but I don't understand.

src/MySqlConnector/MySqlCommand.cs
djeman: Ok I will try.

src/MySqlConnector/MySqlParameter.cs
```diff
@@ -714,12 +714,9 @@ private void AppendBinary(ByteBufferWriter writer, object value, StatementPrepar
 			writer.WriteLengthEncodedInteger(unchecked((ulong) geometry.ValueSpan.Length));
 			writer.Write(geometry.ValueSpan);
 		}
-		else if (value is MemoryStream memoryStream)
+		else if (value is Stream)
```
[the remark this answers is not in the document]
djeman: Ok I will modify.

src/MySqlConnector/MySqlCommand.cs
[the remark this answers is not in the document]
bgrainger: But the bug here is if the `Stream` is >= 16MB but only returns, say, 8MB in the first read, which the `Stream.Read` method is allowed to do:

> An implementation is free to return fewer bytes than requested even if the end of the stream has not been reached.

https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.read

We would need to loop (or call `Stream.ReadAtLeast(Async)` until at least 16MiB has been read, in order to satisfy the MySQL protocol constraint of sending exactly 0xffffff bytes in the initial packet(s).

https://mariadb.com/docs/server/reference/clientserver-protocol/0-packet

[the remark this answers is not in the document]
bgrainger: On second thoughts, we probably don't even want to fill up to MaxPacketSize here (because SendAsync will already split large payloads across multiple packets).

src/MySqlConnector/MySqlCommand.cs
```diff
@@ -357,15 +359,68 @@ internal async ValueTask<MySqlDataReader> ExecuteReaderAsync(CommandBehavior beh
 		return await ExecuteReaderNoResetTimeoutAsync(behavior, ioBehavior, cancellationToken).ConfigureAwait(false);
 	}
 
-	internal ValueTask<MySqlDataReader> ExecuteReaderNoResetTimeoutAsync(CommandBehavior behavior, IOBehavior ioBehavior, CancellationToken cancellationToken)
+	internal async ValueTask<MySqlDataReader> ExecuteReaderNoResetTimeoutAsync(CommandBehavior behavior, IOBehavior ioBehavior, CancellationToken cancellationToken)
 	{
 		if (!IsValid(out var exception))
-			return ValueTaskExtensions.FromException<MySqlDataReader>(exception);
+			return await ValueTaskExtensions.FromException<MySqlDataReader>(exception).ConfigureAwait(false);
```
bgrainger, line 365: Just `throw` it.

PR #1580, opened 2025-07-24
Use `await using` in ASP.NET Core tutorial doc

jnoordsij:
When browsing through the docs I encountered this example, which can probably enhanced by using `await using` given that all respective classes from this package implement the `AsyncDisposable` method.


jnoordsij:
Further examples have been updated; this should be ready to review now.

bgrainger (approved):
The `await using MySqlCommand` is a bit overkill, as `DisposeAsync` is just a synchronous method that sets `m_isDisposed = true`. However, the compiler should handle a completed `ValueTask` efficiently, and it means client code doesn't have to be dependent on those internal details (which might change).

jnoordsij:
Thanks for merging!

> The `await using MySqlCommand` is a bit overkill, as `DisposeAsync` is just a synchronous method that sets `m_isDisposed = true`. However, the compiler should handle a completed `ValueTask` efficiently, and it means client code doesn't have to be dependent on those internal details (which might change).

I definitely did not look into the details that much and figured that "`await using` should always be preferred to `using`" is probably a good rule of thumb for client code, so just encouraging that in general is probably easier.

PR #1584, opened 2025-08-13
Issue#1435

Jatkingmodern:


bgrainger:
Using the `slnx` file format is intentional (see https://github.com/mysql-net/MySqlConnector/commit/5c0df946b6a51ead9a28dc5c1728c1cb5ab93161); don't add back a `.sln` file.

I don't understand the purpose of adding `internal static class DatabaseTelemetry` to `tools/SchemaGenerator/SchemaGenerator.cs`. Why was this code added?

Issue #1585, opened 2025-08-20
MySqlConnector.MySqlException : XAER_RMFAIL on TransactionScope.Dispose()


PatrickMNL:
**Software versions**
MySqlConnector version: 2.4.0
Server type (MySQL, MariaDB, Aurora, etc.) and version: MariaDB 11.8.3
.NET version: 8.0

**Describe the bug**
When running into a deadlock, in the following scenario;

T1 - XA START
T2 - XA START
T1 - XA END
T1 - XA PREPARE
T1 - XA COMMIT
T2 - Deadlock Exception (aborting the transaction, disposing the TransactionScope)
T2 - XA PREPARE

MariaDB 11.8.3 returns the following exception;

```
MySqlConnector.MySqlException : XAER_RMFAIL: The command cannot be executed when global transaction is in the  ROLLBACK ONLY state
   at MySqlConnector.Core.ServerSession.ReceiveReplyAsync(IOBehavior ioBehavior, CancellationToken cancellationToken) in /_/src/MySqlConnector/Core/ServerSession.cs:line 885
   at MySqlConnector.Core.ResultSet.ReadResultSetHeaderAsync(IOBehavior ioBehavior) in /_/src/MySqlConnector/Core/ResultSet.cs:line 37
   at MySqlConnector.MySqlDataReader.ActivateResultSet(CancellationToken cancellationToken) in /_/src/MySqlConnector/MySqlDataReader.cs:line 130
   at MySqlConnector.MySqlDataReader.InitAsync(CommandListPosition commandListPosition, ICommandPayloadCreator payloadCreator, IDictionary`2 cachedProcedures, IMySqlCommand command, CommandBehavior behavior, Activity activity, IOBehavior ioBehavior, CancellationToken cancellationToken) in /_/src/MySqlConnector/MySqlDataReader.cs:line 482
   at MySqlConnector.Core.CommandExecutor.ExecuteReaderAsync(CommandListPosition commandListPosition, ICommandPayloadCreator payloadCreator, CommandBehavior behavior, Activity activity, IOBehavior ioBehavior, CancellationToken cancellationToken) in /_/src/MySqlConnector/Core/CommandExecutor.cs:line 56
   at MySqlConnector.MySqlCommand.ExecuteNonQueryAsync(IOBehavior ioBehavior, CancellationToken cancellationToken) in /_/src/MySqlConnector/MySqlCommand.cs:line 309
[...]

PR #1586, opened 2025-08-20
Handle a MySqlException with error code XAERRemoveFail indicating the XA state is ROLLBACK ONLY.

PatrickMNL:
Potential fix for #1585


bgrainger:
Thanks for the contribution!

Issue #1317, opened 2023-05-04
MySqlConnector.MySqlException in TransactionScope.Dispose caused by a deadlock creates a memory leak


PatrickMNL:
**Software versions**
MySqlConnector version: 2.2.5
Server type (MySQL, MariaDB, Aurora, etc.) and version: 10.6.12-MariaDB-log
.NET version: 6.0

**Describe the bug**
When a deadlock is thrown by the SQL database, we catch and retry it as suggested by the exception. But before we retry, our using statement inside our try/catch block disposes of the then active TransactionScope. The dispose however is interrupted by another exception thrown by MySqlConnector, causing the TransactionScope to not be properly disposed and leaving behind a reference in the [s_transactionConnections ](https://github.com/mysql-net/MySqlConnector/blob/89e2d45b4c5f1432ee8a90656cb7172f9e4df1b9/src/MySqlConnector/MySqlConnection.cs#L1111) dictionary which will snowball into ever increasing memory leakage for each subsequently created transaction.

**Exception**
```
MySqlConnector.MySqlException (0x80004005): XA_RBDEADLOCK: Transaction branch was rolled back: deadlock was detected
   at MySqlConnector.Core.ResultSet.ReadResultSetHeaderAsync(IOBehavior ioBehavior) in /_/src/MySqlConnector/Core/ResultSet.cs:line 43
   at MySqlConnector.MySqlDataReader.ActivateResultSet(CancellationToken cancellationToken) in /_/src/MySqlConnector/MySqlDataReader.cs:line 130
   at MySqlConnector.MySqlDataReader.CreateAsync(CommandListPosition commandListPosition, ICommandPayloadCreator payloadCreator, IDictionary`2 cachedProcedures, IMySqlCommand command, CommandBehavior behavior, Activity activity, IOBehavior ioBehavior, CancellationToken cancellationToken) in /_/src/MySqlConnector/MySqlDataReader.cs:line 469
   at MySqlConnector.Core.CommandExecutor.ExecuteReaderAsync(IReadOnlyList`1 commands, ICommandPayloadCreator payloadCreator, CommandBehavior behavior, Activity activity, IOBehavior ioBehavior, CancellationToken cancellationToken) in /_/src/MySqlConnector/Core/CommandExecutor.cs:line 56
[...]

PR #1589, opened 2025-09-02
Handle XA_RBDEADLOCK on XA END while still calling XA ROLLBACK afterwards.

PatrickMNL:


PatrickMNL:
Unfortunately I have to make adjustments in this method once again (after applying fixes for MariaDB 10.6.12+ here #1317 and for MariaDB 11.8.3+ here #1585). 

We have migrated to AWS Aurora MySQL and since then we were suffering from connections that remained in an invalid state after having suffered a deadlock (while using XA). The connection that suffered from the deadlock subsequently kept throwing:
```
MySqlConnector.MySqlException : XA_RBDEADLOCK: Transaction branch was rolled back: deadlock was detected
```

When debugging I noticed that the issue we saw in #1317 was now occurring on the XA END call rather than the XA ROLLBACK call, causing the code to never call XA ROLLBACK, leaving the connection in an invalid state. 

To ensure compatibility it seems like we need to handle the case for both calls separately in the OnRollback scenario.

bgrainger:
Thanks for the contribution!

Is it possible to add tests to make this failure happen deterministically (e.g., with the steps given in https://github.com/mysql-net/MySqlConnector/issues/1585#issue-3338624938) to verify compatibility (and check for regressions) with the DBs used in the integration tests? (Unfortunately the integration tests don't run against Aurora, but do use MariaDB.)

PatrickMNL:
Thanks for the response! 

We are currently prepping for a production migration to AWS, so I won't be able to add that on the short term, but I am willing to have a go at it somewhere after that, that would hopefully be in 2 weeks or so.

PR #1637, opened 2026-05-04
[Experimental] Turn off CRL checks

bart-vmware:
Here's what I tried locally. It just hacks in CRL-off without a connection string parameter. Please discard the first commit; it doesn't help. I've added extra logging to analyze what happens, which should probably be left out. The code works for VerifyFull, but I haven't tested mutual-TLS yet. Note that it requires .NET 7.0 or higher, so it should probably log a warning when used with netstandard.


bart-vmware:
Superseded by #1657.

Issue #1644, opened 2026-05-15
Less overhead for MySqlBulkCopy


SoftStoneDevelop:
**Problem**: 

MySqlBulkCopy requires a DataRow or DataTable.

Even in you integration test:

```csharp

var bulkCopy = new MySqlBulkCopy(connection)
{
	DestinationTableName = "bulk_copy_column_mapping",
	ColumnMappings =
	{
		new(1, "@val", "intvalue = @val + 1"),
		new(3, "text"),
		new(4, "data"),
	},
};

var dataTable = new DataTable()
{
	Columns =
	{
		new DataColumn("c1", typeof(int)),
		new DataColumn("c2", typeof(int)),
		new DataColumn("c3", typeof(string)),
		new DataColumn("c4", typeof(string)),
		new DataColumn("c5", typeof(byte[])),
	},
	Rows =
	{
		new object[] { 1, 100, "a", "A", new byte[] { 0x33, 0x30 } },
		new object[] { 2, 200, "bb", "BB", new byte[] { 0x33, 0x31 } },
		new object[] { 3, 300, "ccc", "CCC", new byte[] { 0x33, 0x32 } },
	},
};

```

You need to specify a DataTable. This requires converting the values ​​for columns c1, c2, and basically all columns to `object` (which results in packing). Then, when sending, we unpack the value back in MySqlBulkCopy.WriteValue so we can write `Utf8Formatter.TryFormat(intValue, output, out bytesWritten);`

Therefore, if I have some data (sometimes in one collection, sometimes in several), I first need to convert it all to a DataTable/DataRow, which adds overhead, and then MySqlBulkCopy itself unpacks it back to write it.

**I propose an API similar to Npgsql (in terms of usage)**:

```csharp

var bulkCopy = new MySqlBulkCopy(connection)
{
	DestinationTableName = "bulk_copy_column_mapping",
	ColumnMappings =
	{
		new(1, "@val", "intvalue = @val + 1"),
		new(3, "text"),
		new(4, "data"),
	},
};

bulkCopy.StartRow();
bulkCopy.WriteInt(0, 1);
bulkCopy.WriteInt(1, 100);
bulkCopy.WriteString(2, "a");
bulkCopy.WriteString(3, "A");
bulkCopy.WriteBytes(4, new byte[] { 0x33, 0x30 });

bulkCopy.StartRow();
// ...

```

[...]

PR #1646, opened 2026-05-17
The method for converting byte values ​​is exposed

SoftStoneDevelop:
Now you can write data directly to a stream without any extra memory allocations, while relying on the library's conversion.

#1644

### DIFF
--- a/src/MySqlConnector/Helpers/ValueWriteHelper.cs
+++ b/src/MySqlConnector/Helpers/ValueWriteHelper.cs
@@ -1,0 +1,333 @@
+using System.Buffers.Text;
+using System.Globalization;
+using System.Numerics;
+using System.Text;
+using MySqlConnector.Utilities;
+
+namespace MySqlConnector.Helpers;
+
+internal class ValueWriteHelper
+{
+	private static readonly char[] s_specialCharacters = ['\t', '\\', '\n'];
+	private static readonly UTF8Encoding s_utf8Encoding = new(encoderShouldEmitUTF8Identifier: false, throwOnInvalidBytes: true);
+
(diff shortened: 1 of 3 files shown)


SoftStoneDevelop:
Of course, anyone can write their own ValueWriteHelper, but why bother if the library already has one? So, I moved it into a separate class and made it public.

MySqlBulkCopy.SendDataReaderAsync is good because it uses an internal array. Therefore, the trick with an external Stream that's filled elsewhere has a downside: it requires all data to be written to the Stream.

So it seems that without some high-level customization, MySqlBulkLoader.SourceStream, while it works, will require too much memory if the inserted collection is large, even if you use a RecyclableMemoryStream.

bgrainger:
I'm not completely sure what the point of this PR is.

The PR says "now you can write data directly to a stream" but also says "MySqlBulkLoader.SourceStream, while it works, will require too much memory".

Why are we trying to improve a method that will use too much memory?

Regarding #1644, it's not fully clear to me what the major problem is and why this is a good solution for it. If you have a lot of data you need to load quickly, have you tried putting it in CSV format (could be a file on disk, no memory requirement) and loading it with `MySqlBulkCopy`?

SoftStoneDevelop:
Give me a couple of days, and I'll finish a full example so it's clear what I'm trying to do and why. Right now, it's really not entirely clear (it's all just partly in my head) what I'm trying to do, why, and what benefit it will bring to the project.

I'll close this PR and come back with a full example and benchmark that clearly demonstrate what I'm trying to do and why.

bgrainger:
Benchmarks identifying the problem and showing how a particular approach (significantly) improves it would be really helpful here. (I suspect that in a lot of cases, client-side data value formatting is a trivial percentage of the overall elapsed time, which will be dominated by MySQL Server inserting the data.)

PR #1657, opened 2026-07-02
Add SkipCertificateRevocationCheck connection string option

### LINKED ISSUE #1615: Allow skipping Certificate Revocation List check from connection string
I'm trying to use `MySqlSslMode.VerifyFull` to ensure the server CA certificate is valid and trusted, which fails in a constrained cloud environment (no internet access) because the revocation check fails. The problem is the following code at https://github.com/mysql-net/MySqlConnector/blob/2.4.0/src/MySqlConnector/Core/ServerSession.cs#L1676:

```c#
var checkCertificateRevocation = cs.SslMode == MySqlSslMode.VerifyFull;
// ...
clientAuthenticationOptions.CertificateRevocationCheckMode = 
    checkCertificateRevocation ? X509RevocationMode.Online : X509RevocationMode.NoCheck
```

I would like to be able to configure MySqlConnector to use `X509RevocationMode.Offline` instead.

bart-vmware:
Contains the changes from #1638, rebased on the latest master branch.

This PR adds an additional commit that logs the full error chain at Trace level, which is pretty critical to fully see what `SslStream` returns.
Also addressed the concern at https://github.com/mysql-net/MySqlConnector/pull/1638#discussion_r3185718034, which I believe is valid.

Compared to #1637, this PR _doesn't_ include changes for client certificates (mutual TLS). Our cloud broker [doesn't support it](https://techdocs.broadcom.com/us/en/vmware-tanzu/platform/tanzu-mysql-tanzu-platform/10-4/mysql-tp/using-tls.html), so I'm unable to test that part.

Fixes #1615.

### DIFF
--- a/docs/content/connection-options.md
+++ b/docs/content/connection-options.md
@@ -176,6 +176,14 @@ These are the options that need to be used in order to configure a connection to
       <p>To provide a custom callback to validate the remote certificate, leave this option empty and set <code>SslMode</code> to <code>Required</code> (or <code>Preferred</code>), then set <a href="/api/mysqlconnector/mysqlconnection/remotecertificatevalidationcallback/"><code>MySqlConnection.RemoteCertificateValidationCallback</code></a> before calling <a href="/api/mysqlconnector/mysqlconnection/open/"><code>MySqlConnection.Open</code></a>. The property should be set to a delegate that will validate the remote certificate, as per <a href="https://docs.microsoft.com/en-us/dotnet/api/system.net.security.remotecertificatevalidationcallback" title="RemoteCertificateValidationCallback Delegate (MSDN)">the documentation</a>.</p>
     </td>
   </tr>
+  <tr id="AllowUnknownCertificateRevocation">
+    <td>Allow Unknown Certificate Revocation, AllowUnknownCertificateRevocation</td>
+    <td>false</td>
+    <td>
+      <p>Allows a connection using <code>SslMode=VerifyFull</code> to succeed even if revocation checking fails because certificate revocation status can't be determined. All other <code>VerifyFull</code> checks are still performed, including CA and hostname validation.</p>
+      <p>This option is only supported on .NET 7.0 or later, and it may only be used with <code>SslMode=VerifyFull</code>.</p>
+    </td>
+  </tr>
   <tr id="TlsVersion">
     <td>TLS Version, TlsVersion, Tls-Version</td>
     <td></td>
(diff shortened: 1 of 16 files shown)


bart-vmware:
I'm wondering if `SkipCertificateRevocationCheck` would better represent what this setting is supposed to provide?

bgrainger:
I think that with the change to:

```csharp
var checkCertificateRevocation = cs.SslMode == MySqlSslMode.VerifyFull && !cs.AllowUnknownCertificateRevocation;
```

then `SkipCertificateRevocationCheck` is a better name.

I can't push to your branch, so I've cherry-picked everything back to #1638.

Before I close this and merge that PR, there was no `Signed-off-by` line in your commit message. Can you attest that you're certifying all the statements in https://developercertificate.org/ for your contribution?

bart-vmware:
I've added your commits to this PR and added `Signed-off-by` to my commits. You can push to my branch now. I prefer to continue working here because it builds against the latest master, and we can both add commits now.

I've removed setting the `VerificationFlags`, which is now redundant since we already set `X509RevocationMode.NoCheck`. This removes the TFM limitation. `tests/SkipCrl/repro.ps1` still returns the same output, and I've verified that this works with our cloud broker. I've clarified the documentation based on Copilot feedback.

bgrainger (approved):
Thanks!

I had the same question (prompted by Copilot) about VerificationFlags so agree with removing that logic.

bart-vmware:
@bgrainger Can you please provide the ETA for release? Thanks

bgrainger:
I'm anticipating shipping v3.0 after November 10, 2026 (to coincide with .NET 11), so it'll release by then at the very latest. If a v2.6.2 becomes necessary, it will be included in that release.

bart-vmware:
Would you consider releasing this earlier (weeks) as a minor/patch release? I started looking into this based on an internal security audit for a product I'm actively working on with my team. We target large enterprises, some of which delay updates to newer .NET versions (even when out of support). I'm assuming that v3.0 will drop support for .NET 8/9?

bgrainger:
No plans to drop support for `net8.0`/`net9.0`. The next major version will very likely remove _explicit_ support for `net6.0` (https://github.com/mysql-net/MySqlConnector/commit/297f8b3721cf48664ab7f4d7affafe2f22102dfc) but even then the `netstandard2.0` TFM is being retained so it should still run on that framework.

The long gap between my asking for a review (May 4: https://github.com/mysql-net/MySqlConnector/issues/1615#issuecomment-4376120109) and one being provided (July 2: https://github.com/mysql-net/MySqlConnector/issues/1615#issuecomment-4866405811) made me assume this wasn't highly time-sensitive. (After the review was provided, I merged the resulting code the very next day, but unfortunately that was already after v2.6.1 had shipped.)

This is currently the only change on `master` since v2.6.1. There may be some outstanding PRs I can clean up and merge that would help merit a v2.6.2 in the next few weeks; I will take a look.

bart-vmware:
Thanks. You're right this wasn't time-sensitive back then. At that time, we were flooded with vulnerability reports in existing products, which got higher priority. This is for a new product, which is reaching completion and we'd like to release it soon.

bgrainger:
Shipped in [2.6.2](https://github.com/mysql-net/MySqlConnector/releases/tag/2.6.2).

bart-vmware:
Thanks!

PR #1659, opened 2026-07-15
Clarify EOL status for MySQL 8.0

jnoordsij:
See https://dev.mysql.com/doc/relnotes/mysql/8.0/en/

Note: 9.4 is listed as still tested, although it is not a LTS release nor receiving any kind of updates anymore. It might make sense to test just 9.7 from the 9.x series?


bgrainger:
Good feedback; I could probably thin the test suite to just 8.4 and 9.7 (LTS versions).

